PR #2, opened 2016-05-23
Migrate mobile_app scripts from testeng-ci into this repo

aleffert:
Add mobile app build scripts.
Setup checking python code in the .travis.yml file.

(note that this also includes the changes from aleffert/initial_setup. That will get resolved once that branch gets merged).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
 sudo: false
 language: groovy
 script:
-    # uncomment once we add a project
+    # not a python container + don't have sudo permissions so need
+    # to setup virtualenv ourselves
+    - pip install --user virtualenv
+    - $HOME/.local/bin/virtualenv venv
+    - source venv/bin/activate
+    - pip install -r shared/test-requirements.txt
+
+    # mobile app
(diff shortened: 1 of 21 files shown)


aleffert:
@benpatterson please review. This is broken out in two commits. The first is just moving stuff. The second one connects it to CI

.travis.yml
```diff
@@ -1,6 +1,17 @@
 sudo: false
 language: groovy
```
benpatterson: I think you can refactor this according to the below pattern. Have you tried that, perchance?
```
language:
  - python
  - groovy
```

aleffert: Oh, I didn't know you could do that - I thought they had different images per language. But I'll give it a shot.

aleffert: For those following along at home, that doesn't work. It just picks the first item.

benpatterson: As discussed offline: bummer. but thanks for looking.

benpatterson:
@aleffert I compared the files using `git diff --no-index` on the old repo, and only see the change from relative imports.

👍

PR #4, opened 2016-06-02
Estute/move testeng dsls to open repo

estute:
@benpatterson @jzoldak 
If you could check this out when you get a chance.


benpatterson:
It looks like this is failing on Travis. I'm guessing that Travis is defaulting to some other gradle installation that it ships with. Could you try uncommenting Akiva's comments on the .travis.yml file to see if that fixes it? If not, could you dig into it further?

estute:
@benpatterson @jzoldak .gitignore was blocking jars from being pushed (hence the gradle failures). Fixed now.

estute:
@jibsheet could you take a look at this?

benpatterson:
👍 from me.

PR #6, opened 2016-06-09
changed permissions for groups only

estute:
@benpatterson @aphanse This is a better solution than having to edit each job manually each time.


benpatterson:
Gotcha. LGTM.

aphanse:
👍

PR #7, opened 2016-06-09
Made changes to be compatible with plugins

aphanse:
@estute @jzoldak 
Hi! My plugin versions on my localhost didn't match build-jenkins, so I updated those and updated the scripts as a result. Also, made sure the shell commands were in one build step.

### DIFF
--- a/platform/jobs/edxPlatformAccessibilityMaster.groovy
+++ b/platform/jobs/edxPlatformAccessibilityMaster.groovy
@@ -115,7 +115,7 @@ secretMap.each { jobConfigs ->
             }
         }
         triggers { //trigger when change pushed to GitHub
-            gitHubPushTrigger()
+            githubPush()
         }
         wrappers { //abort when stuck after 75 minutes, use gnome-terminal coloring, have timestamps at Console
             timeout {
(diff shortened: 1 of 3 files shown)


platform/jobs/edxPlatformAccessibilityMaster.groovy
```diff
@@ -134,8 +134,8 @@ secretMap.each { jobConfigs ->
                    }
                }
            }
-           shell('cd edx-platform')
-           shell('RUN_PA11YCRAWLER=1 ./scripts/accessibility-tests.sh')
+           shell('cd edx-platform \n' +
+                 'RUN_PA11YCRAWLER=1 ./scripts/accessibility-tests.sh')
```
estute: Nit: Having multiple shell commands in the same line is creeping me out. Can you put them into separate lines?

estute:
Aside from nitpicking, LGTM 👍

platform/jobs/edxPlatformAccessibilityMaster.groovy
aphanse: Sure! I had them in the same line because otherwise it was 2 separate build steps, but I can revert it :D.

estute: Ah, no, sorry. I meant something more like this
`shell('command \n` +
          'other command\n' +
          '....'`
It was this part that struck me as weird 'RUN_PA11YCRAWLER=1 ./scripts/accessibility-tests.sh'

aphanse:
@jzoldak @benpatterson 
If you have time and can take a look at this that would be great :)

jzoldak:
👍

PR #8, opened 2016-06-09
DSL Script for Bok-Choy Job

aphanse:
@estute @jzoldak 
Hi whenever you have a chance can you go through this PR? Thank you!

### DIFF
--- a/platform/jobs/edxPlatformBokChoyMaster.groovy
+++ b/platform/jobs/edxPlatformBokChoyMaster.groovy
@@ -147,7 +147,7 @@ secretMap.each { jobConfigs ->
         triggers { //trigger when change pushed to GitHub
             githubPush()
         }
-        dslFile('edx-platform-bok-choy-master.groovy')//'testeng-ci/jenkins/flow/master/edx-platform-bok-choy-master.groovy')
+        dslFile('testeng-ci/jenkins/flow/master/edx-platform-bok-choy-master.groovy')
         publishers { //JUnit Test report, trigger GitHub-Build-Status, email, message hipchat
            archiveJunit(jUnitReports)
            downstreamParameterized {
--- a/platform/jobs/edxPlatformBokChoyMaster.groovy
+++ b/platform/jobs/edxPlatformBokChoyMaster.groovy
@@ -6,17 +6,20 @@ import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_PARSE_SECRET
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_HIPCHAT
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_BASE_URL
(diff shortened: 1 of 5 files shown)


platform/jobs/edxPlatformBokChoyMaster.groovy
```diff
@@ -0,0 +1,178 @@
...
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_HIPCHAT
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_BASE_URL
+
+/*
+Example secret YAML file used by this script
+publicJobConfig:
+    open : true/false
+    jobName : name-of-jenkins-job-to-be
+    url: github-url-segment
+    url-clone : github-url-segment-clone
```
estute: nit: perhaps better variable naming here. either numbered urls (url1, url2) or something like 'platformUrl'/'testengUrl', just to avoid confusion. Not super important

estute:
Aside from nit-picking, :+1

platform/jobs/edxPlatformBokChoyMaster.groovy
```diff
@@ -0,0 +1,178 @@
...
+predefinedPropsMap.put('CONTEXT', 'jenkins/bokchoy')
+
+String jUnitReports = 'edx-platform/**/nosetests.xml,edx-platform/reports/acceptance/*.xml,'
+jUnitReports += 'edx-platform/reports/quality.xml,edx-platform/reports/javascript/javascript_xunit.xml,'
+jUnitReports += 'edx-platform/reports/bok_choy/xunit.xml,edx-platform/reports/bok_choy/**/xunit.xml'
+
+/* stdout logger */
+/* use this instead of println, because you can pass it into closures or other scripts. */
+/* TODO: Move this into JenkinsPublicConstants, as it can be shared. */
+Map config = [:]
```
jzoldak: I think we might have at least 3 job definitions that now have this copy/pasted code. It probably would be a good time to refactor and pull this out to a shared location.

platform/jobs/edxPlatformBokChoyMaster.groovy
jzoldak: I'm also a bit confused as to what url-clone, credential-clone, and cloneReference are for.

aphanse: the url and credential clone are to check out the build flow scripts from the repo, the url-clone, credential-clone, and cloneReference refer to the platform. I will change the variables names, at the time I couldn't think of a better naming scheme.

platform/jobs/edxPlatformBokChoyMaster.groovy
aphanse: I will be taking the jUnitReports out, the map has values that change depending on the test running which I am hesitant to place in a shared location. However, that map is used in identical ways (through github-status) in several files which would be good to pull out.

aphanse:
Refactored code and added it to the JenkinsPublicConstants

aphanse:
If have time and can look over the changes that would be lovely! :D

estute:
👍 . Checked that the other files are just a weird side effect. They look the same as a previous merge. Git weirdness....

PR #11, opened 2016-06-21
Estute/dsl test poc

estute:
@jzoldak @aphanse poc for using the spock test framework for testing our dsl scripts.

I moved everything in shared/src down a level to src, just to make gradle easier to deal with.

important part here is src/test/groovy/SampleJobSpec.groovy

Couple of things I still want to iron out (later PR/jira):
* Add a cleanup method for each test (but need to figure out how to destroy the jenkins instance/remove created content)
* Fix build dependency for our dsls when we want to generate XML on the fly. Basically, you can say "./gradlew debugXml -D<somePath> and it will generate the XML config file so that you can eyeball it for a sanity check. However, due to us using secrets, there are some issues. However, this is not the main goal of the test framework.
* Possibly condense repetetive tasks into a closure for the tests


src/test/groovy/sample/SampleJobSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+        given:
+        JobManagement jm = new JenkinsJobManagement(System.out, [:], new File('.'))
+        DslScriptLoader loader = new DslScriptLoader(jm)
+        File dslScript = new File('sample/jobs/sampleJob.groovy')
+
+        when:
+        loader.runScript(dslScript.text)
+
+        then:
+        noExceptionThrown()
```
aphanse: Write a comment saying that these are the assertions, for anyone who doesn't know groovy it is really confusing.

src/test/groovy/sample/SampleJobSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+        File dslScript = new File('sample/jobs/sampleJob.groovy')
+
+        when:
+        loader.runScript(dslScript.text)
+        GPathResult project = new XmlSlurper().parseText(jm.getConfig('SampleJenkinsJob'))
+
+        then:
+        Node logRotatorBlock = project.childNodes().find { it.name == 'logRotator' }
+        logRotatorBlock.childNodes().any { it.name == 'daysToKeep' && it.text() == '10' }
+
```
aphanse: Add a comment for what is going on here

aphanse:
👍 , other than some comments to explain the weirdness of spock :)

README.md
```diff
@@ -6,19 +6,49 @@
...
+DSL jobs can be tested using the [Spock test framework](http://spockframework.github.io/spock/docs/1.0/index.html). 
+
+To run all of the test specs:
+`./gradlew test`
+To run all of the test specs within a given project:
+`./gradlew test --tests sample.*`
+To run all a single test spec:
+`./gradlew test --tests sample.SampleJobSpec`
+
+### Linting
```
benpatterson: nit: how about we call it static code analysis. that's more like what codenarc does, but without the baggage of sounding like a big nitter.

gradle.properties
```diff
@@ -0,0 +1,4 @@
+#jobDslVersion=1.45
```
benpatterson: remove the comments?

build.gradle
```diff
@@ -49,36 +55,84 @@ dependencies {
...
 }
 
+
+
 task libs(type: Copy) {
     into 'lib'
     from configurations.libs
 }
 
+task cleanLibs(type: Delete) {
```
benpatterson: I remember gradle having a default `clean` task. Is this also a default task? Perhaps it's a child, or it's the new `clean`? Basically I want to use conventional targets... I realize you didn't write this originally but it's sticking out to me at the moment.

build.gradle
```diff
@@ -49,36 +55,84 @@ dependencies {
...
-    configFile = file('config/codenarc/codenarcRules.groovy')
-    maxPriority2Violations = 49
-    maxPriority3Violations = 91
+task resolveTestPlugins(type: Copy) {
+    from configurations.testPlugins
+    into new File(sourceSets.test.output.resourcesDir, 'test-dependencies')
+    include '*.hpi'
+    include '*.jpi'
+
+    doLast {
```
benpatterson: could you comment to describe what this is doing?

src/main/groovy/org/edx/jenkins/dsl/debug/XmlOutputRunner.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+String pattern = System.getProperty('pattern')
+
+if (!pattern) {
+    println 'usage: -Dpattern=<pattern>'
+    System.exit 1
+}
+
+File outputDir = new File('./build/debug-xml')
+outputDir.deleteDir()
```
benpatterson: This confuses me....shouldn't that be picked up by clean as well? Or, if `clean` isn't clean enough...should this fall in some sort of `setUp` function?

src/main/groovy/org/edx/jenkins/dsl/rest/RestApiJobManagement.groovy
```diff
@@ -0,0 +1,121 @@
...
+    InputStream streamFileInWorkspace(String filePath) throws IOException {
+        new File(filePath).newInputStream()
+    }
+
+    @Override
+    String readFileInWorkspace(String filePath) throws IOException {
+        new File(filePath).text
+    }
+
+    private boolean create(String name, String xml, boolean isView) {
```
benpatterson: This is all good stuff but I admit I'm wondering if it's in a library that we could just import instead?

src/test/groovy/sample/SampleJobSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Basic POC tests for the sample dsl script sample/jobs/sampleJob.groovy
+ */
+class SampleJobSpec extends Specification {
+
+    @Shared
+    @ClassRule
+    JenkinsRule jenkinsRule = new JenkinsRule()
+
+    @Unroll
```
benpatterson: I don't know this annotation..what's it mean?

benpatterson:
Done with my pass. I was also wondering if your example could show us a spock data table? That's a useful pattern that folks new to the repo may want to know about.

src/test/groovy/sample/SampleJobSpec.groovy
estute: @benpatterson you can't really see the benefit of it here, but if you use something like a data table or iterate over instances within a single test, the @Unroll annotation will report them as separate tests. I was originally doing that and forgot to pull them out. I will add an example showing this off

build.gradle
estute: gradle clean won't affect the libs directory... it cleans out build artifacts. This is a custom task and is really just a convenience task for devs. We will NOT be using this in seed jobs. I have no qualms with removing it.

src/main/groovy/org/edx/jenkins/dsl/debug/XmlOutputRunner.groovy
estute: Clean is not a default task, it must be explicitly called. I can definitely change that.

estute:
@aphanse @benpatterson @jzoldak 
Temporarily removed functionality for debugXML, due to uncertainty about license and build dependency issues. Cleaned up the build.gradle file.

I also added more documentation to the sampleJobSpec to help people get groovy/spock. I created a generic Setup for the entire class and added a data table to show off some of the framework features.

build.gradle
benpatterson: ah ok. no need to change; carry on.

benpatterson:
I dig it. You should get a look from someone in devops, even if it's after you merge. They should see what we're up to and give feedback on whether or not it'll work for their use cases. We can iterate on it as necessary.

jzoldak:
Looks awesome! 👍

PR #13, opened 2016-06-22
Bok Choy PR DSL Script

aphanse:
@jzoldak @estute 
This is the Bok Choy Pr Job. Whenever you have time please check it out. Also the skip build phrase is set up in the global configuration which is why it is not present.

### DIFF
--- a/platform/jobs/edxPlatformBokChoyPr.groovy
+++ b/platform/jobs/edxPlatformBokChoyPr.groovy
@@ -1,12 +1,9 @@
-//package devops
+package devops
 
 import hudson.model.Build
-//import static org.edx.jenkins.dsl.
-import static JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
-//import static org.edx.jenkins.dsl.
-import static JenkinsPublicConstants.JENKINS_PUBLIC_PARSE_SECRET
-//import static org.edx.jenkins.dsl.
-import static JenkinsPublicConstants.JENKINS_PUBLIC_JUNIT_REPORTS
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_PARSE_SECRET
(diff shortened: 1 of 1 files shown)


platform/jobs/edxPlatformBokChoyPr.groovy
```diff
@@ -0,0 +1,147 @@
...
+            stringParam(stringParams.name, stringParams.default, stringParams.description)
+        }
+        logRotator JENKINS_PUBLIC_LOG_ROTATOR() //Discard build after a certain amount of time
+        concurrentBuild() //concurrent builds can happen
+        label('flow-worker-bokchoy') //restrict to jenkins-worker
+        checkoutRetryCount(5)
+        multiscm {
+            git { //using git on the branch and url, clean before checkout
+                remote {
+                        github(jobConfig['testengUrl'])
```
jzoldak: typo with the spacing

platform/jobs/edxPlatformBokChoyPr.groovy
```diff
@@ -0,0 +1,147 @@
...
+                    relativeTargetDirectory('edx-platform')
+                }
+            }
+        }
+        triggers { //trigger when pull request is created
+            pullRequest {
+                admins(jobConfig['admin'])
+                useGitHubHooks()
+                triggerPhrase('jenkins run bokchoy')
+                cron('H/5 * * * *')
```
jzoldak: I see in the current job configuration that it's set up with a Crontab line, but I'm not sure if/why it's needed. Are we sure that it is? If so please add a comment explaining why.

aphanse: It isn't needed. The global configuration also has a crontab set up.

platform/jobs/edxPlatformBokChoyPr.groovy
```diff
@@ -0,0 +1,146 @@
...
+                    cleanBeforeCheckout()
+                    relativeTargetDirectory('edx-platform')
+                }
+            }
+        }
+        triggers { //trigger when pull request is created
+            pullRequest {
+                admins(jobConfig['admin'])
+                useGitHubHooks()
+                triggerPhrase('jenkins run bokchoy')
```
estute: @aphanse while you are here, make the trigger phrase case insensitive

estute:
:+1:

PR #15, opened 2016-06-27
Script Changes After Deployment

aphanse:
@estute @jzoldak 

Changes made:
1) In Bok Choy PR, removed the parameterized variable that was never used. Still need to figure out why the job wasn't being triggered by 'jenkins run bokchoy'

2) In Python Unit Tests Master, removed the line about the relative html directory. The XML now reads:
<jenkins.plugins.shiningpanda.publishers.CoveragePublisher plugin="shiningpanda@0.21"/> which matches that of Build Jenkins: <jenkins.plugins.shiningpanda.publishers.CoveragePublisher plugin="shiningpanda@0.21"/>. The only difference is that the DSL scripted version has coverage.py as the first node in publishers while the version in Build Jenkins has it as a later node.

3) In Jenkins Public Constants added an * in the list of file names for the jUnit Archiver. This was to resolve the issues with JS Master failing because it was unable to publish the files. 

Whenever you get a chance please look this over :D Thanks!

### DIFF
--- a/platform/jobs/edxPlatformBokChoyPr.groovy
+++ b/platform/jobs/edxPlatformBokChoyPr.groovy
@@ -28,12 +28,6 @@ Binding bindings = getBinding()
 config.putAll(bindings.getVariables())
 PrintStream out = config['out']
 
-stringParams = [
-    name: 'sha1',
-    description: 'Sha1 hash of branch to build. Default branch : master',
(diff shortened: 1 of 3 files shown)


jzoldak:
👍

estute:
:+1

PR #16, opened 2016-06-29
BokChoy Pr Changes

aphanse:
These changes are for fixing Bok Choy Pr Job for deployment. They were done because the job dsl automatically generated some xml that was confusing the GHPRB on where to listen to, the testeng-ci repo or edx-platform. 
@estute @jzoldak  Whenever you have time please look it over :)

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
@@ -15,7 +15,7 @@ class JenkinsPublicConstants {
     }
   }
 
-  public static final String JENKINS_PUBLIC_GITHUB_BASEURL = "http://github.com/"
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
```diff
@@ -15,6 +15,8 @@ class JenkinsPublicConstants {
     }
   }
 
+  public static final String JENKINS_PUBLIC_GITHUB_BASEURL = "http://github.com/"
```
jzoldak: Should we be using http or https? What are the advantages/disadvantages to either of those choices?

aphanse: It should be https because that is what github uses and what is present in other non scripted pr jobs

jzoldak:
👍

estute:
:+1

PR #17, opened 2016-07-01
add tests for edxPlatformTestSubset dsl

estute:
@jzoldak @aphanse @benpatterson 
care to take a look when you get a chance?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,6 +30,7 @@ repositories {
 
 configurations {
     libs
+    testPlugins {}
 }
 
 dependencies {
@@ -52,10 +53,23 @@ dependencies {
     }
     // Needed for mocking in tests
     testCompile 'cglib:cglib:3.1' 
-    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}"
-    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}@jar"
+
+    // Jenkins test harness dependencies
     testCompile 'org.jenkins-ci.main:jenkins-test-harness:2.8'
     testCompile "org.jenkins-ci.main:jenkins-war:${jenkinsVersion}"
+    testCompile "org.jenkins-ci.main:jenkins-war:${jenkinsVersion}:war-for-test@jar"
+
+    // Job DSL plugin including plugin dependencies
+    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}"
+    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}@jar"
+    testCompile 'org.jenkins-ci.plugins:structs:1.1@jar'
+    testCompile 'org.jenkins-ci.plugins:cloudbees-folder:4.4@jar'
+
+    //testCompile "org.jenkins-ci.plugins:credential:1.24"
+    //testPlugins "org.jenkins-ci.plugins:structs:1.2"
+    //testPlugins "org.jenkins-ci.plugins:credentials:1.23"
+    //testPlugins "org.jenkins-ci.plugins:plain-credentials:1.0"
+    //testPlugins "org.jenkins-ci.plugins.workflow:workflow-step-api:2.2"
 }
 
 codenarc {
@@ -69,10 +83,22 @@ task libs(type: Copy) {
     into 'lib'
     from configurations.libs
 }
+task resolveTestPlugins(type: Copy) {
+    from configurations.testPlugins
+    into new File(sourceSets.test.output.resourcesDir, 'test-dependencies')
+    include '*.hpi'
+    include '*.jpi'
+
+    doLast {
+        def baseNames = source.collect { it.name[0..it.name.lastIndexOf('.')-1] }
+        new File(destinationDir, 'index').setText(baseNames.join('\n'), 'UTF-8')
+    }
+}
 
 tasks.getByName('test').inputs.files(sourceSets.jobs.groovy.srcDirs)
 
 test {
+    dependsOn tasks.resolveTestPlugins
     inputs.files sourceSets.jobs.groovy.srcDirs
     systemProperty 'buildDirectory', project.buildDir.absolutePath
 }
--- a/build.gradle
+++ b/build.gradle
@@ -6,18 +6,18 @@ defaultTasks 'clean', 'libs'
 sourceSets {
     jobs {
         groovy {
-            srcDir "platform/jobs"
-            srcDir "sample/jobs"
+            srcDir 'platform/jobs'
+            srcDir 'sample/jobs'
         }
     }
     src {
         groovy {
-            srcDir "src/main/groovy"
+            srcDir 'src/main/groovy'
         }
     }
     test {
         groovy {
-            srcDir "src/test/groovy"
+            srcDir 'src/test/groovy'
         }
     }
 }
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -52,10 +53,23 @@ dependencies {
...
     testCompile "org.jenkins-ci.main:jenkins-war:${jenkinsVersion}"
+    testCompile "org.jenkins-ci.main:jenkins-war:${jenkinsVersion}:war-for-test@jar"
+
+    // Job DSL plugin including plugin dependencies
+    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}"
+    testCompile "org.jenkins-ci.plugins:job-dsl:${jobDslVersion}@jar"
+    testCompile 'org.jenkins-ci.plugins:structs:1.1@jar'
+    testCompile 'org.jenkins-ci.plugins:cloudbees-folder:4.4@jar'
+
+    //testCompile "org.jenkins-ci.plugins:credential:1.24"
```
benpatterson: any reason these are commented out? do we want them? I'm particularly curious about the credentials plugins since we are heavily dependent on secrets.

build.gradle
```diff
@@ -69,10 +83,22 @@ task libs(type: Copy) {
     into 'lib'
     from configurations.libs
 }
+task resolveTestPlugins(type: Copy) {
+    from configurations.testPlugins
+    into new File(sourceSets.test.output.resourcesDir, 'test-dependencies')
+    include '*.hpi'
+    include '*.jpi'
+
+    doLast {
```
benpatterson: could you add a comment describing what this does?

build.gradle
```diff
@@ -69,10 +83,22 @@ task libs(type: Copy) {
     into 'lib'
     from configurations.libs
 }
+task resolveTestPlugins(type: Copy) {
```
benpatterson: nit: should have a different name, since this appears to actually create files.

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
```diff
@@ -0,0 +1,226 @@
...
+        setup:
+        String secretPath = baseSecretPath + "/non-existent-file.yml"
+        jm =  loadSecret(secretVar, secretPath)
+        loader = new DslScriptLoader(jm)
+
+        when:
+        GeneratedItems generatedItems = loader.runScript(dslScript.text)
+
+        then:
+        generatedItems.jobs.size() == 0
```
benpatterson: should we also encounter an error when the script is run?

benpatterson:
The testcases themselves look great. You cover some good negative cases. 

I had some comments about implementation.

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
```diff
@@ -0,0 +1,226 @@
...
+        }
+        else {
+            !properties.childNodes().any { it.name == 'hudson.security.AuthorizationMatrixProperty' }
+        }
+
+        where:
+
+        job                                 | open  | block | permissions 
+        "edx-platform-test-subset"          | true  | false | _
+        "edx-platform-test-subset_2"  | false | true  | pList
```
aphanse: Nit: can you fix the whitespace so that the table aligns. Pretty minor.

platform/jobs/edxPlatformTestSubset.groovy
```diff
@@ -64,10 +65,9 @@ def secretMap = [:]
...
     out.println('Parsing secret YAML file')
     /* Parse k:v pairs from the secret file referenced by secretFileVariable */
-    Thread thread = Thread.currentThread()
-    Build build = thread?.executable
-    Map envVarsMap = build.parent.builds[0].properties.get("envVars")
-    secretMap = JENKINS_PUBLIC_PARSE_SECRET.call(secretFileVariable, envVarsMap, out)
+    String contents = new File("${EDX_PLATFORM_TEST_SUBSET_SECRET}").text
+    Yaml yaml = new Yaml()
+    secretMap = yaml.load(contents)
     out.println('Successfully parsed secret YAML file')
```
aphanse: Is this specific for writing the tests, or is this something that needs to be changed about all the jobs?

aphanse:
Other than what Ben said, and a couple of questions/nits 👍

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
estute: Yeah, that looks terrible. Easy fix.

platform/jobs/edxPlatformTestSubset.groovy
estute: This will need to be changed in all of the jobs, in order to work with the tests. Here is why:
When you run the dsl scripts via a Jenkins seed job, you have access to the full Jenkins run time, and you can manipulate Executors/Build data. However, when you run the spock tests, we are using a minimal version of Jenkins (not the same as a full fledged installation), so we can't access, say, build history or executors. I have tested this on a sandbox installation, and the ${} notation works for secrets (it is the normal way of accessing env vars within the Build steps of a job.

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
estute: Technically, this would throw an error (File Not Found or IOException). However, we have encapsulated the Yaml secret parsing within a try block that returns a "1" if any exception is caught. The reason for this is that returning a 1 (or any int besides 0) will fail the Jenkins job, and this is desirable, so that the DSL engine doesnt continue running other scripts. Unfortunately, because we return a 1, the call to loader.runScript() doesn't funnel that back to the test. I don't like it, but until I can come up with a better way, it "seems" the only method to test for this type of failure is by verifying nothing was successful.

benpatterson: Ah yes I forgot about the exit 0 problem. Out of curiosity is there a way we could catch any exception and exit 1 generically (not for this PR)

estute: I would like to do this. Will continue working on it, but not in this PR

build.gradle
estute: This was a relic of me playing around with the Jenkins Test Harness. In order to test the XML content created by the plugin, other plugins are NOT necessary on the system.

build.gradle
estute: See above

build.gradle
estute: See above

estute:
I am going to delete this (testing github api)

build.gradle
```diff
@@ -6,18 +6,18 @@ defaultTasks 'clean', 'libs'
 sourceSets {
     jobs {
         groovy {
-            srcDir "platform/jobs"
-            srcDir "sample/jobs"
+            srcDir 'platform/jobs'
+            srcDir 'sample/jobs'
         }
```
aphanse: Any reason for switching to ' from "?

aphanse:
👍 Other than that question

build.gradle
estute: Trying to standardize... but unfortunately some of the strings are GStrings the repo as a whole to match codenarc style

PR #19, opened 2016-07-07
Modify BokChoy DSL

aphanse:
@benpatterson @estute @jzoldak 
This is to match the changes in the sharding script. 
There are two environment variables that are injected through the script whose values are in the yaml file and can to configured to be job specific.


benpatterson:
LGTM

estute:
:+1:

PR #21, opened 2016-07-12
Refactor Job DSLs

aphanse:
@benpatterson @jzoldak @estute 
Hi All! This is a large PR but the goal was to go through and refactor all the jobs to work well various repos. Also, a lot of the manual changing of the bok choy pr job and test subset job is reflected here. 
 
Changes to All Files:
1) References to edx-platform were changed to edx-platform* (in various archivers) or something specific to the directory the files were checked out to
2) Instead of always checking out to a directory called edx-platform, it checks out to one whose name matches the repo name
3) Added the new parsing method that works better with testing
4) In general, calling github() was replaced by url()

Changes to Buildflow Jobs (in particular):
1) Inject environment variables to specify which test-subset job to use, changes will be reflected in the sharding scripts as well

Changes to Test-Subset:
1) Use different urls to access the repo
2) Use ssh agent if the job is not open

### DIFF
--- a/src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
+++ b/src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
@@ -167,7 +167,7 @@ class edxPlatformTestSubsetSpec extends Specification {
         Node scm = project.childNodes().find { it.name == "scm" }
         Node urc = scm.childNodes().find { it.name == "userRemoteConfigs" }
         Node giturc = urc.childNodes().find { it.name == "hudson.plugins.git.UserRemoteConfig" }
-        giturc.childNodes().any { it.name == "url"  && it.text() == url }
+        giturc.childNodes().any { it.name == "url" && it.text() == url }
         if (!open) {
             giturc.childNodes().any { it.name == "credentialsId" && it.text() ==  cred }
         }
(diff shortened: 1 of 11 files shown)


estute:
:+1

estute:
:+1:

estute:
@aphanse what is the purpose of this:

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
```diff
@@ -167,7 +167,7 @@ class edxPlatformTestSubsetSpec extends Specification {
         Node scm = project.childNodes().find { it.name == "scm" }
         Node urc = scm.childNodes().find { it.name == "userRemoteConfigs" }
         Node giturc = urc.childNodes().find { it.name == "hudson.plugins.git.UserRemoteConfig" }
-        giturc.childNodes().any { it.name == "url" && it.text() == "https://github.com/${url}.git" }
+        giturc.childNodes().any { it.name == "url"  && it.text() == url }
```
jzoldak: is the extra space before the && needed?

jzoldak:
👍

src/test/groovy/platform/edxPlatformTestSubsetSpec.groovy
aphanse: no, i didn't notice it was there. fixing that now :)

PR #23, opened 2016-07-13
A11y PR DSL Script

aphanse:
@benpatterson @estute @jzoldak 
A11y PR DSL Script

Special things to note: 
1) The current shell script changes the directory to "edx-platform" before running the accessibility tests. It now works with whatever folder the repo is copied into. 
2) Makes use of the new parsing system

Please look it over whenever you have a chance!! Thanks!!

### DIFF
--- a/platform/jobs/edxPlatformAccessibilityPr.groovy
+++ b/platform/jobs/edxPlatformAccessibilityPr.groovy
@@ -1,0 +1,149 @@
+package devops
+
+import hudson.model.Build
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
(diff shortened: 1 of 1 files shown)


platform/jobs/edxPlatformAccessibilityPr.groovy
```diff
@@ -0,0 +1,149 @@
...
+                userWhitelist(jobConfig['userWhiteList'])
+                orgWhitelist(jobConfig['orgWhiteList'])
+                extensions {
+                    commitStatus {
+                        context('jenkins/a11y')
+                    }
+                }
+            }
+        }
+        wrappers { //abort when stuck after 65 minutes, use gnome-terminal coloring, have timestamps at Console
```
jzoldak: It would be better to not put the number in the comment. Actually the comment itself doesn't really add much value, I'd drop it. IIRC we intentionally over-commented the first one so that there would be a good example for people to follow, but it's not really necessary for all DSL scripts.

jzoldak:
👍

PR #30, opened 2016-07-19
JS PR DSL and A11y DSL Improvements

aphanse:
@estute @jzoldak 
There are 3 things here. 1) Creating the JS PR DSL script. 2) Fixing the A11y PR DSL script to correctly work with both repos. 3) Using the combined GHPRB secret instead of having it in each individual file.

Whenever you have a chance please look it over :)


estute:
:+1: thanks for keeping the codenarc violations down :)

PR #33, opened 2016-07-25
Refactor Scripts

aphanse:
These changes take care of the job being able to run on multiple repos (have to use a different protocol than before) and fix what went wrong during the deployment of quality pr. Please look it over whenever you have a chance :)


estute:
:+1:

PR #35, opened 2016-07-25
Typing Mistake

aphanse:
@estute 
I got a little too excited about getting python working. This is the proper way to handle handle the string


estute:
:+1:

PR #36, opened 2016-07-27
Inject Env Var for Python Unit Tests to work

aphanse:
In order to allow the .coveragerc file to be more flexible and work for the repos, we need to inject the name of the subset job as an environment variable. The name of the coverage job is already present as an environment variable.


estute:
Do you know if the rc file can deal with env varS?

aphanse:
It can! I tested it out and it works well. :)

estute:
:+1:

PR #37, opened 2016-07-27
Standardize secrets documentation

aphanse:
@estute @jzoldak 
The documentation on how to use the shared constants is added to this readme file as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,29 @@ To run codenarc on the shared code:
 `./gradlew codenarcSrc`
 etc...
 
+## Writing DSLs
+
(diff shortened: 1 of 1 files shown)


estute:
See my comment on https://github.com/edx-ops/edx-jenkins-job-dsl/pull/44. Basically the consumers of these scripts don't need to know how or where secrets exist on the system, but they need to know how to access them (via secret name and the constituent values within the secret)

estute:
:+1: Perhaps at a later date we can formalize all of this into a guide for those who don't use groovy/jenkins too often.

PR #38, opened 2016-08-01
Update to eslint

aphanse:
@jzoldak @estute @benpatterson 
Updating to include eslint in the list of HTML Reports and allowing files to be missing (depending on if eslint or jshint are present)


jzoldak:
👍

PR #42, opened 2016-08-08
Fix Python Master Job

aphanse:
@estute @jzoldak 
Ran on the master jobs on the private repo and found out that I missed a set of refactoring on the python master job. Needed to inject a coverage job so that it would use the correct version.


jzoldak:
👍

PR #43, opened 2016-08-09
Refactor Master Jobs

aphanse:
@estute @jzoldak 

Essentially, this PR is going through all the master jobs and making sure they report the correct results back to the appropriate repository.  This meant moving part of the predefined properties map into the area where is could be job specific.


jzoldak:
👍

PR #44, opened 2016-08-09
dsl script to backup jenkins instances

estute:
@aphanse @jzoldak still working out a failing test, but if you could check this out


aphanse:
Other than the failing test, LGTM :+1:

testeng/jobs/backupJenkins.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+        environmentVariables {
+            env('AWS_DEFAULT_REGION', jobConfig['region'])
+        }
+        
+        // This might seem overkill, but in case the pip requirements change, read them from
+        // the requirements file in the workspace
+        String script = ""
+        readFileFromWorkspace('testeng/resources/requirements.txt').split("\n").each { line ->
+            script += "pip install ${line}; "
```
jzoldak: We probably want to specify wipe for the exists-action option in the pip command, so that if we bump the version it will not get stuck at a user prompt.

Also, is there an advantage to concatenating with a semicolon onto one line for the script? I'm not that great with bash, but I am pretty sure that with a semicolon if the first command fails, then the second will still run. 
It'd be better (and more readable) to have them each on their own line anyhow.

testeng/jobs/backupJenkins.groovy
```diff
@@ -0,0 +1,95 @@
...
+        // Configure the Exclusive Execution plugin, to reduce the amount of things in memory during snapshotting
+        configure { project ->
+            project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
+                skipWaitOnRunningJobs false
+            }
+        }
+
+        wrappers {
+            timeout {
+                absolute(20)
```
jzoldak: curious if you have data on how long it takes to do the backup.

estute: First, about the data: I backed up Test Jenkins, which is about 9% the size of Build Jenkins, and it took about 5 minutes to back up. However, once the snapshot command has been issued, the snapshot process proceeds asynchronously, with regard to the Jenkins process. This timeout value will only come into play if the pip command is hanging. I suppose we should shorten it. 

However, this brings up a good question. What happens if the something goes wrong with aws. The script outputs some JSON data describing it's status. We can use Jenkins to verify something in the build process itself to determine if it should be considered a success. Then.. there is the issue of a snapshot starting, and then dying later on. I am not sure if it would make sense to make another call after some time to see if the snapshot no longer is marked as "pending"

testeng/jobs/backupJenkins.groovy
estute: No advantage to the semicolons. I am not a fan either, but I was a little confused about the virtualenv plugin in the DSL. I have since tested and line breaks are no issue. Will put it into next PR

estute: And adding the wipe-command, although I imagine 99.9% of the time we should be fine, b/c the worker should spin down and up between instances of running this job.

PR #46, opened 2016-08-11
Added script to run jenkins job against edx-e2e-tests repo

Muddasser:


benpatterson:
@estute could you review?

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+Map config = [:]
+Binding bindings = getBinding()
+config.putAll(bindings.getVariables())
+PrintStream out = config['out']
+
+
+/* Map to hold the k:v pairs parsed from the secret file */
+Map secretMap = [:]
+try {
+    out.println('Parsing secret YAML file')
```
benpatterson: I realize this is a copy from somewhere else, but I think you should put this `println` above the `try`.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+        /* For non-open jobs, enable project based security */
+        if (!jobConfig['open'].toBoolean()) {
+            authorization {
+                blocksInheritance(true)
+                permissionAll('edx')
+            }
+        }
+
+        properties {
+              githubProjectUrl(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['platformUrl'])
```
benpatterson: I think this should actually be the e2e tests github url?

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        logRotator JENKINS_PUBLIC_LOG_ROTATOR() //Discard build after 14 days
+
+        concurrentBuild() //concurrent builds can happen
```
benpatterson: I don't think this is true, in the case of this job. If you had the same tests run at the same time, you'd run into state problems as they're coded today.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+
+        label('master') //restrict to flow-worker-bokchoy
```
benpatterson: `'jenkins-worker'` unless @estute has another recommendation.  You should also change the comment to match whatever worker you choose.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+            env('COURSE_NUMBER', jobConfig['authPass'])
+            env('COURSE_RUN', jobConfig['courseRun'])
+            env('COURSE_DISPLAY_NAME', jobConfig['courseDisp'])
+        }
+
+        scm {
+
+           git { //using git on the branch and url, clone, clean before checkout
+                remote {
+                    url(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['platformUrl'] + '.git')
```
benpatterson: You want this to be the e2e tests url. The way it's written, it'll pull down edx-platform's repo.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+    authPass : password-of-auth-user-for-stage
+    loginEmail : email-for-stage-login
+    loginPass : pass-for-stage-login
+    courseOrg: organization-of-course-for-test
+    courseNumber: course-number-of-course-for-test
+    courseRun: course-run-number-of-course-for-test
+    courseDisp: course-display-name-of-course-for-test
+    repoName : repo-name
+    open : true/false
+    email: me@email.com
```
benpatterson: Is this supposed to be the user's login e-mail for the tests? Or is it the person that will be notified when the job fails?

benpatterson:
@Muddasser I took a pass; I think you'll need a couple of updates so it pulls down the right repo.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
Muddasser: It is the email of the person that will be notified.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
Muddasser: 'platformUrl' is coming from secret file. So you can match it to e2e repo.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
Muddasser: Oh gosh! Fair point.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
Muddasser: 'platformUrl' is coming from secret file. So you can match it to e2e repo.

Muddasser: 'platformUrl' is coming from secret file. So you can match it to e2e repo.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
estute: yes, 'jenkins-worker' is the appropriate worker type for this.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,143 @@
...
+
+        checkoutRetryCount(5)
+
+        environmentVariables {
+            env('BASIC_AUTH_USER', jobConfig['authUser'])
+            env('BASIC_AUTH_PASSWORD', jobConfig['authPass'])
+            env('USER_LOGIN_EMAIL', jobConfig['loginEmail'])
+            env('USER_LOGIN_PASSWORD', jobConfig['loginPass'])
+            env('COURSE_ORG', jobConfig['courseNumber'])
+            env('COURSE_NUMBER', jobConfig['authPass'])
```
estute: COURSE_ORG and COURSE_NUMBER are mixed up.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,141 @@
...
+    /* Test secret contains all necessary keys for this job */
+    /* TODO: Use/Build a more robust test framework for this */
+    assert jobConfig.containsKey('authUser')
+    assert jobConfig.containsKey('authPass')
+    assert jobConfig.containsKey('loginEmail')
+    assert jobConfig.containsKey('loginPass')
+    assert jobConfig.containsKey('courseOrg')
+    assert jobConfig.containsKey('courseNumber')
+    assert jobConfig.containsKey('courseRun')
+    assert jobConfig.containsKey('courseDisp')
```
estute: Looking at this again, I am not sure that you want to use a secret for the following values: courseOrg, courseNumber, courseRun, courseDisp. It seems that these will/could change for each run of this job. If that is the case,  these should be input into the job as String parameters. This way the user/script that triggers this job can control which values to use.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
estute: I think you should change the name of this variable. It will confuse future users of this script.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
estute: Again, change the variable name here for clarity.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,141 @@
...
+    assert jobConfig.containsKey('repoName')
```
estute: This is not used within the script.

Muddasser:
@estute Your comments have been addressed.

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,165 @@
...
+Example secret YAML file used by this script
+
+    authUser : name-of-auth-user-for-stage
+    authPass : password-of-auth-user-for-stage
+    loginEmail : email-for-stage-login
+    loginPass : password-for-stage-login
+    courseOrg: organization-of-course-for-test
+    courseNumber: course-number-of-course-for-test
+    courseRun: course-run-number-of-course-for-test
+    courseDisp: course-display-name-of-course-for-test
```
estute: Can you remove these lines now that they are not contained within a secret. It will make it easier for others to understand

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
estute: These values (i.e. courseDisp) are no longer contained within a secret, so no need to check for their values

platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy
```diff
@@ -0,0 +1,165 @@
...
+
+        environmentVariables {
+            env('BASIC_AUTH_USER', jobConfig['authUser'])
+            env('BASIC_AUTH_PASSWORD', jobConfig['authPass'])
+            env('USER_LOGIN_EMAIL', jobConfig['loginEmail'])
+            env('USER_LOGIN_PASSWORD', jobConfig['loginPass'])
+            env('COURSE_ORG', jobConfig['courseOrg'])
+            env('COURSE_NUMBER', jobConfig['courseNumber'])
+            env('COURSE_RUN', jobConfig['courseRun'])
+            env('COURSE_DISPLAY_NAME', jobConfig['courseDisp'])
```
estute: Again, since these values are not parsed out of secrets, you must access them differently

PR #47, opened 2016-08-12
TE-1621 Update Ecom Dsls

aphanse:
@jzoldak @estute 

I separated the end to end tests into 3 different files. They have default password parameters that cannot be seen in the config.xml or in the console. The password parameters can also be manually overwritten.


estute:
:+1: looks good to me. Only thing we might want to consider is that this currently masks every parameter, as opposed to the old version of these jobs which masks certain parameters. Not sure if ecom wants this. Let me know if we want this changed.

PR #48, opened 2016-08-18
E0d/rationalize plugins

e0d:
@jibsheet 
- sorts plugins.txt
- updates and rationalizes pinned versions
- synch'd with you plugin changes
- pins docker From to 1.x family


jibsheet:
👍

PR #50, opened 2016-08-18
Ansible libraries hardcode #!/usr/bin/python

jibsheet:
This fails when a library requires boto or some other package which has
not been installed in the system python.

@e0d this is horrible


e0d:
:+1: until we have something better.

PR #52, opened 2016-08-19
tweak script to use keys instead of roles

estute:
@JZ switched script to use aws keys following our conversation this morning. Also, set the jobs produced by this to run at different times. Also, I simply cat out the script output, since the virtualenv plugin is hardwired to use dash, not bash, and therefore no pipefail options.

### DIFF
--- a/testeng/jobs/backupJenkins.groovy
+++ b/testeng/jobs/backupJenkins.groovy
@@ -28,7 +28,7 @@ Config:
     jenkinsInstance : test
     volumeId : vol-123
     region : us-west.1
-    accessKey : 123
+    accessKeyId : 123
     secretAccessKey: 123
     hipchat : hipchat token
     email : email@address.com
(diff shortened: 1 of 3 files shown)


benpatterson:
I think you want @jzoldak

estute:
@benpatterson as well

benpatterson:
LGTM

testeng/jobs/backupJenkins.groovy
```diff
@@ -82,19 +92,22 @@ secretMap.each { jobConfigs ->
...
-        script += "sync\n"
+        String script = "sync\n"
         // This might seem overkill, but in case the pip requirements change, read them from
         // the requirements file in the workspace
         readFileFromWorkspace('testeng/resources/requirements.txt').split("\n").each { line ->
             script += "pip install --exists-action w ${line}\n"
         }
-        script += "aws ec2 create-snapshot --volume-id ${jobConfig['volumeId']} --description 'Automatic ${jobConfig['jenkinsInstance']} jenkins snapshot' |tee \${WORKSPACE}/snapshot-out"
+        script += "aws ec2 create-snapshot --volume-id ${jobConfig['volumeId']} --description 'Automatic ${jobConfig['jenkinsInstance']} jenkins snapshot' > \${WORKSPACE}/snapshot-out\n"
+        script += "cat \${WORKSPACE}/snapshot-out"
```
jzoldak: nit: this should have a file extension. I'd recommend .log.

PR #54, opened 2016-08-23
update schedule for backup jobs

estute:
@JZ changed schedule to avoid stepping on Lahore's toes

### DIFF
--- a/testeng/jobs/backupJenkins.groovy
+++ b/testeng/jobs/backupJenkins.groovy
@@ -70,7 +70,7 @@ secretMap.each { jobConfigs ->
         configure { project ->
             project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
                 // Do not wait for running jobs to complete before executing this job
-                // This is done because a locks when dealing with sub-jobs.
+                // This is done because of locks when dealing with sub-jobs.
                 skipWaitOnRunningJobs true
             }
         }
--- a/testeng/jobs/backupJenkins.groovy
+++ b/testeng/jobs/backupJenkins.groovy
@@ -69,14 +69,14 @@ secretMap.each { jobConfigs ->
         // during snapshotting
         configure { project ->
             project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
-                skipWaitOnRunningJobs false
+                skipWaitOnRunningJobs true
             }
         }
 
         // Run snapshotting script once a day, when there is usually no Jenkins activity
         triggers {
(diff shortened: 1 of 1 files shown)


estute:
@benpatterson > following our conversation this morning

benpatterson:
@jzoldak I argued that we can solve the "it's still writing to disk" problem separately. That was starting to balloon into properly stopping system activity and I'd rather get this up and working. We can address the part about stopping the system activity in a separate piece of work.

If we are doing backups once a day or once a week, I'd rather that be running, even if the last ~30 mins of data is not the most up to date, considering all the rest of the time we're backing up data.

Anyway: 👍

testeng/jobs/backupJenkins.groovy
```diff
@@ -69,18 +69,18 @@ secretMap.each { jobConfigs ->
         // during snapshotting
         configure { project ->
             project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
-                skipWaitOnRunningJobs false
+                skipWaitOnRunningJobs true
             }
         }
 
         // Run snapshotting script once a day, when there is usually no Jenkins activity
```
jzoldak: This is incorrect now that it's once a week, right? Probably best to remove this comment, and put a comment instead above the cron statements below (cron('0 1 * * 6')) that explain them in english (e.g. At 1am on Sundays).

testeng/jobs/backupJenkins.groovy
```diff
@@ -69,18 +69,18 @@ secretMap.each { jobConfigs ->
         // during snapshotting
         configure { project ->
             project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
-                skipWaitOnRunningJobs false
+                skipWaitOnRunningJobs true
```
jzoldak: nit: It'd be worth a comment here since this requires a bit of mental gymnastics. e.g. Don't wait for running jobs to finish.

estute:
fixed nits @jzoldak

jzoldak:
👍

testeng/jobs/backupJenkins.groovy
```diff
@@ -69,18 +69,22 @@ secretMap.each { jobConfigs ->
         // during snapshotting
         configure { project ->
             project / buildWrappers << 'hudson.plugins.execution.exclusive.ExclusiveBuildWrapper' {
-                skipWaitOnRunningJobs false
+                // Do not wait for running jobs to complete before executing this job
+                // This is done because a locks when dealing with sub-jobs.
```
jzoldak: s/because a locks/because of locks

PR #55, opened 2016-08-25
Estute/script android builder

estute:
@jzoldak @benpatterson 
* I am not adding the tests until I finish the job (hockey app, gradle tests, signing), because thats where testing is more important (xml generation through non-dsl closures)


jzoldak:
👍

PR #57, opened 2016-08-30
implement signing scripts in android build job

estute:
@benpatterson 
A couple of things:
* fixes to the hockey app calls
* TODO: implement jarsigner test (for app certificate)
* TODO: better spock test for this

### DIFF
--- a/mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
+++ b/mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
@@ -8,10 +8,16 @@ Example secret YAML file used by this script
 jobConfig:
     jobName: build-demo-android-app
     public: true
+    release: false
+    buildScript: build.sh
+    gradleProperties: GRADLE_PROPERTIES
+    keystore: APP_KEYSTORE
+    keystoreFile: APP_KEYSTORE_FILE
     gitRepo: https://github.com/org/project.git
     gitCredential: git_user
     appBaseDir: android-app-project
     appBaseName: edx-demo-app
+    generatedApkName: edx-demo-app.apk
     sshAgent: ssh_user
     hockeyAppApiToken: abc123
 */
@@ -35,6 +41,7 @@ catch (any) {
     out.println('Exiting with error code 1')
     return 1
 }
+
 out.println('Successfully parsed secret YAML file')
 
 // Iterate over the configurations in BUILD_ANDROID_SECRET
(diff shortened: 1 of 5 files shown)


mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
```diff
@@ -8,10 +8,16 @@ Example secret YAML file used by this script
 jobConfig:
     jobName: build-demo-android-app
     public: true
+    release: false
```
benpatterson: Could you comment on this to make it more clear what it means?

mobileApp/resources/check_release_build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# Verify that a APK intended to be released is both:
```
benpatterson: nit: s/a APK/an APK

mobileApp/resources/check_release_build.sh
```diff
@@ -0,0 +1,15 @@
...
+#   - TODO: Signed with a certificate
+#   - Not a debug apk (seems redundant, but the build script could be accidetnally changed)
```
benpatterson: nit: typo in 'accidentally'

mobileApp/resources/check_release_build.sh
```diff
@@ -0,0 +1,15 @@
...
+DEBUGGABLE=`$ANDROID_HOME/build-tools/23.0.3/aapt dump badging $WORKSPACE/artifacts/$APK_NAME |grep -q 'application-debuggable'`
+
+if [ $DEBUGGABLE -gt 0 ]; then
+    echo "This bug is debuggable, and should NOT be. Please fix and rebuild"
```
benpatterson: did you mean to say this _build_ is debuggable?

benpatterson:
Some wording/comments changes. 👍 once you sort those out.

mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
estute: @benpatterson We usually add the comments to the actual secret files. Do you think we should put them here too?

benpatterson: Ah. Well, yes that would be helpful. An open source person woudl more-easily understand what's going on. To a certain extent I'd rather see the comments in the open source repo rather than the secrets repo for that reason. And it would likely apply to not-just open source folks; it would apply to edX folks who don't have access to the secrets.

You don't have to solve that particular pattern here, but perhaps you can add the comment to this line?

PR #58, opened 2016-08-31
use credentials binding and fix resource scripts

estute:
@benpatterson additional fixes found when running the seed.

### DIFF
--- a/mobileApp/resources/copy_keys.sh
+++ b/mobileApp/resources/copy_keys.sh
@@ -1,6 +1,7 @@
-#!/usr/bin/bash env
+#!/usr/bin/env bash -e
 
 # Copy gradle.properties and the keystore file into the application directory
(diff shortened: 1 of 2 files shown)


mobileApp/resources/copy_keys.sh
```diff
@@ -1,6 +1,7 @@
-#!/usr/bin/bash env
+#!/usr/bin/env bash -e
```
benpatterson: Oh I didn't realize you could pass in the `-e` on the hashbang line. This works?

benpatterson:
👍

mobileApp/resources/copy_keys.sh
estute: no, just a typo

benpatterson:
from my perspective, merge when ready

PR #59, opened 2016-08-31
add documentation to android dsl

estute:
@jibsheet

### DIFF
--- a/mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
+++ b/mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
@@ -18,10 +18,12 @@ import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTA
  ==========================================================================================
  In order to run, the job requires the following be loaded into Jenkins as secrets files:
    - A secret yml file: This contains a number of jobConfigs, one for each desired Jenkins job
-   - A gradle.properties file: A key-value file used in the gradle build process **
+   - A gradle.properties file (release builds only): A key-value file used in the gradle build process **
        For more informataion, see https://github.com/edx/edx-app-android/blob/master/README.rst#building-for-release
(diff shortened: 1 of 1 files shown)


jibsheet:
@feanil can you be the second set of eyes who doesn't know DSL as well?

mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
```diff
@@ -3,24 +3,44 @@ package mobile
...
+   - Publish the apk to HockeyApp (a service for hosting & disrtibuting application files)
+   - TODO: Run tests on the generated apk file (linting, unit tests, screenshot tests)
+ Depending on the "release" flag mentioned below, the job will be considered either 'release' or 'debug'
+ Release builds will also use special gradle.properties and keystore files when compiled, to sign
+ the build.
+
+ ==========================================================================================
+ In order to run, the job requires the following be loaded into Jenkins as secrets files:
+   - A secret yml file: This contains a number of jobConfigs, one for each desired Jenkins job
+   - A gradle.properties file: A key-value file used in the gradle build process **
```
feanil: You mentioned above that there will be two different sets of (gradle.properties, keystore)  So in total 5 files will be uploaded is that right?  Perhaps we can enumerate both sets. eg.
` A debug gradle.properties file:...`
` A debug keystore file:...`
`A release gradle.properties file:...`
`A release keystore file:...`

feanil:
One issue otherwise generally makes sense to me.

mobileApp/jobs/buildWhiteLabelAndroidApps.groovy
jibsheet: I think we do need two gradle properties, but there's just no keystore for debug builds (there's a ** footnote below).

jibsheet:
I agree @estute clear up how we differentiate the gradle properties and I think you're good to go.

PR #61, opened 2016-09-19
Keep artifacts for jenkins subjobs for 5 days only

jzoldak:
@edx/testeng 

[TE-1694](https://openedx.atlassian.net/browse/TE-1694)


benpatterson:
LGTM

PR #63, opened 2016-09-20
fix var names and add xunit archival

estute:
@benpatterson @MuddasserHussain changed env var names to match test script. Also, if we use the JUnit archiver, it is much easier to navigate the results via Jenkins. However, I am not sure if we want this- is it crucial that the result xml is downloadable?


benpatterson:
@estute yes, xunit results were part of how we built it, and I wanted to see that happen particularly so we could use it in the Jenkins job. So it's critical that we use the xunit results as part of the job. Making it downloadable afterwards? I'd err on the side of 'yes' because it is not going to be a big file and we may want to run analytics on it at some point when we look for optimizations.

PR #64, opened 2016-09-22
change call to jenkins-report

estute:
@benpatterson @jzoldak

### DIFF
--- a/platform/jobs/edxPlatformUnitCoverage.groovy
+++ b/platform/jobs/edxPlatformUnitCoverage.groovy
@@ -161,6 +161,9 @@ secretMap.each { jobConfigs ->
                 sshAgent(jobConfig['credential'])
             }
             buildName('#\${BUILD_NUMBER}: \${GIT_REVISION,length=8}')
+            credentialsBinding {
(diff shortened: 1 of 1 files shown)


estute:
Now supporting both codecov & coveralls (until we are ready to pull the plug)

benpatterson:
👍

platform/jobs/edxPlatformUnitCoverage.groovy
```diff
@@ -176,7 +179,10 @@ secretMap.each { jobConfigs ->
                     }
                 }
             }
-            shell('./scripts/jenkins-report.sh ' + jobConfig['shellKey'] + ' $CI_BRANCH')
+            // Run jenkins-report.sh which will upload coverage results to
+            // both codecov and coveralls. This is temporary. We plan to move
+            // off of coveralls.
```
jzoldak: @estute I think I'm missing something here. With a refactor of jenkins-report.sh (over in the edx-platform repo), now that we are passing the CODE_COV token shouldn't we be able to control from that repo now whether the coverage results get uploaded to codecov, coveralls, both, or neither?

PR #67, opened 2016-09-29
scripted jenkins worker job

estute:
@jzoldak @benpatterson I cleaned up the build-packer-ami job. Everything is hidden/obfuscated now, and it should only readable, not editable by the org, but TE has full control


jzoldak:
👍

PR #73, opened 2016-10-13
Jenkins job for pull requests on e2e tests

Muddasser:


Muddasser:
This PR is to run e2e tests from edx/edx-e2e-tests on Jenkins. It uses two secret files. 

- For general configuration of script. It should be same as for [this job.](https://github.com/edx/jenkins-job-dsl/blob/master/platform/jobs/edxEndToEndRepoMasterWeeklyJob.groovy)
- For setting up github admin and whitelisting users/orgs for the pull request. 

Note: 
@benpatterson Please see what values you want to set for 'admin', 'userWhiteList' and 'orgWhiteList' in the second secret file. 

@benpatterson @estute Please review

estute:
@Muddasser concerning the second secret file: we have this file already uploaded and used by other jenkins jobs that are triggered by PRs. When we seed this job, we will just make sure that it has access to the secret file

platform/jobs/edxEndToEndTestsPullRequestJob.groovy
```diff
@@ -0,0 +1,183 @@
...
+                browser()
+            }
+        }
+
+        steps {
+            shell("jenkins/end_to_end_tests.sh")
+        }
+
+        triggers { //triggers on each git pull request
+          pullRequest {
```
estute: I think we should use useGitHubHooks() here. It will use web hooks instead of constantly polling github. For an example see  jenkins-job-dsl/platform/jobs/edxPlatformBokChoyPr.groovy.

We are using v1.45 of the DSL plugin (Sorry that that is not clear in the repo, we will change this). The Github Pull Request Builder is deprecated in more recent versions of the plugin, and therefore the API viewer (https://jenkinsci.github.io/job-dsl-plugin/#) must be set to the appropriate version (1.45) when looking up how to configure this in the DSL.

estute:
Aside from the change in triggers, :+1:

benpatterson:
This looks fine to me as well. Do we need to set up the webhook?

estute:
@benpatterson unless there is something to do on the repo itself, the Jenkins component of the web-hook is all set.

benpatterson:
It looked like we'll need to add a secret phrase to the webhook?

Well we can merge and then troubleshoot if it's not working.

sent via Android

PR #77, opened 2016-10-27
Archive the npm-debug.log file on jenkins jobs

jzoldak:
@edx/testeng

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
@@ -3,7 +3,7 @@ package org.edx.jenkins.dsl
 import org.yaml.snakeyaml.Yaml
 
 class JenkinsPublicConstants {
-  
+
   public static final String JENKINS_PUBLIC_BASE_URL = "https://build.testeng.edx.org/"
 
   public static final Closure JENKINS_PUBLIC_LOG_ROTATOR = {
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
```diff
@@ -22,7 +22,7 @@ class JenkinsPublicConstants {
   public static final Closure JENKINS_PUBLIC_ARCHIVE_ARTIFACTS = {
     return {
         pattern("reports/**/*,test_root/log/**/*.png,test_root/log/**/*.log, test_root/log/**/hars/*.har,**/nosetests.xml," +
-                "**/TEST-*.xml"
+                "**/TEST-*.xml,npm-debug.log"
```
benpatterson: should we just do `*.log`?

jzoldak: ah, that's a good idea. I'll amend this.

estute:
@jzoldak are you sure this log is captured on the $WORKSPACE level?

jzoldak:
@estute I was assuming that it would, as npm writes this file to the cwd and AFAIK the paver commands that do the install are run from the workspace folder. Do you have some info that I'm missing?

estute:
@jzoldak oh, ok. I had been poking around in a worker and did not see it, but it could have been that the job had not finished.

jzoldak:
yeah @estute it could also be that at the default logging level, the log is only written on error.

PR #78, opened 2016-11-09
This updates to newer versions of plugins

jibsheet:
THese are the versions of plugins run on the ops jenkins instance, which
is what many of the DSL scripts in here are running.  When it conflicts
with the buildeng plugins, we'll have to figure out a better fix :(

At the very least we need this thought because job-dsl in the ops
jenkins vs this docker container is incompatible.

@e0d @rlucioni


benpatterson:
CC @estute

PR #79, opened 2016-11-10
Jibsheet/create sandboxes

jibsheet:
This is the ansible-provision replacement we'll be using on the new jenkins.


jibsheet:
@e0d @feanil At this point, sandboxes work on tools-edx-jenkins thanks to Feanil's changes to the ssh arguments.  @cpennington and I recently added his timing tools to the output so that they'll go to splunk when we cut over.

I believe we should merge this, but I don't recall if @e0d had ever given it a read back when I wrote it (we were both on alternating vacation weeks and were out of sync).

PR #80, opened 2016-11-10
Cale/sandbox timer logs

cpennington:


cpennington:
@jibsheet: Not sure if you saw this or not.

PR #84, opened 2016-11-28
Build image with Python 3.5.2 installed

rlucioni:
Builds Python 3.5.2 from source. There's no Jessie PPA available.

@jibsheet

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -10,3 +10,13 @@ RUN echo "jenkins ALL=NOPASSWD: ALL" >> /etc/sudoers
 USER jenkins
 COPY plugins.txt /usr/share/jenkins/plugins.txt
 RUN /usr/local/bin/plugins.sh /usr/share/jenkins/plugins.txt
+
+USER root
+RUN apt-get update \
+    && apt-get install -y libssl-dev openssl \
+    && wget https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tgz \
+    && tar xzvf Python-3.5.2.tgz \
+    && cd Python-3.5.2 \
+    && ./configure --prefix=/usr/local \
+    && make \
+    && make install


jibsheet:
I agree that this gets you python3.5 for you and your team, but I honestly think I'd rather stab myself in the eye than merge this.

I'd like to at least spend an hour or so seeing if we can dockerfile from ubuntu instead, but that likely requires some conversations with @estute about their use (or non-use) of this docker on his team.

estute:
@jibsheet currently non-use. Planning on dockerizing our Jenkins set up (https://openedx.atlassian.net/browse/TE-1825)

jibsheet:
Thanks @estute - I think it may make sense to rename the dockerfile to be clearer and play with it a bit.  @rlucioni If I don't have time to experiment this week, we can chat next week.  I think it'd be nice if we had something closer to tools-edx-jenkins in the docker container (especially if plugins.txt didn't require manual syncing).

jibsheet:
@rlucioni  After discussing in retro last week, I'm going to spike today on what needs to happen short vs medium term to ease Job DSL development.  One of these is a better base image.  I'll write a ticket about it and do a little exploration. Please let me know if ECOM might have time to help with the Dockerfile work (basically getting it to run on 14.04 + tools-edx-jenkins playbook from configuration.git)

rlucioni:
Got it, thanks for the heads up @jibsheet. I'll leave this PR up in the meantime in case someone else needs access to an image with 3.5 on it. ECOM is in the middle of a sprint and I'm not sure what we'll be able to squeeze in, but please tag me on the ticket. Why aren't we aiming to have Jenkins running on 16.04?

jibsheet:
14.04 is just what that box was installed with and is running.
I'd love to see it on 16.04, but there's a bunch of manual work to do so, and 14.04 upgrades to 16.04 are lower priority than all the 12.04 upgrades we need to complete.

I suspect it will move to 16.04 when we get to the tasks to refactor ansible to allow generating the initial seed jobs, since that will allow a more automated migration.

rlucioni:
Understood, thanks.

jibsheet:
@jzoldak You're the one other person I thought of who might use docker-compose up in this repo to tests Jenkins (vs having a local Jenkins installed for testing).  Can you confirm whether or not you use this docker image?

jzoldak:
@jibsheet Thx for thinking of me and the heads up. I have not been using this dockerfile for development. I use test-jenkins instead.

jibsheet:
@rlucioni As @MichaelRoytman is getting close to landing a docker container which runs 14.04 and uses the ansible play to provide Jenkins, will you close this PR and delete the branch?  Thanks.

rlucioni:
@jibsheet *poof*. Good luck @MichaelRoytman!

PR #85, opened 2016-11-28
couple of fixes in run_tests

estute:
@benpatterson @jzoldak 
* using virtualenv for requirements
* increased gradle heap to speed up builds
* reorganize emulator target to give it a little more time to set up (querying it too soon causes everything to blow up... and it begins loading as a background job)


benpatterson:
LGTM

jzoldak:
👍

PR #86, opened 2016-11-29
Jenkins job for periodically refresh SailThru content library

umar-nawaz:
ECOM-6354
Currently, refreshing the Sailthru Content library is a manual step. This PR move this task into jenkins to be run periodically. 
@schenedx  you may have a look at this.
@estute can you help review this.

### DIFF
--- a/marketing/sailthruScriptJob.groovy
+++ b/marketing/sailthruScriptJob.groovy
@@ -81,7 +81,19 @@ secretMap.each { jobConfigs ->
                   " --content_api_url " + jobConfig['content_api_url'] + " --lms_url " + jobConfig['lms_url'] +
                   " --report_email " + jobConfig['report_email'] + " --fixups=" + jobConfig['fixups'] +
                   " --type " + jobConfig['type'] + " " + jobConfig['mode']
-
+        
+        wrappers common_wrappers
+        wrappers {
+            maskPasswordsBuildWrapper {
+                varPasswordPairs {
+                    varPasswordPair {
+                        var('SYNC_COMMAND')
+                        password(sync_content)
+                    }
+                }
+            }
+        }
+        
         properties {
               githubProjectUrl(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['githubRepoUrlSegment'])
         }
@@ -104,14 +116,13 @@ secretMap.each { jobConfigs ->
 
         steps {
             virtualenv {
-                pythonName('System-CPython-2.7')
+                pythonName('System-CPython-3.5')
                 nature('shell')
-                command( requirements )
-                command( sync_content )
+                command readFileFromWorkspace('resources/pull-translations.sh')
             }
         }
 
-        triggers { //triggers daily
+        triggers {
             scm('@daily')
         }
 
@@ -120,7 +131,6 @@ secretMap.each { jobConfigs ->
                absolute(75)
             }
             timestamps()
-            colorizeOutput('gnome-terminal')
         }
 
         publishers {
--- a/marketing/sailthruScriptJob.groovy
+++ b/marketing/sailthruScriptJob.groovy
@@ -1,8 +1,11 @@
-package devops
+package org.edx.jenkins.dsl
 
 import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.Constants.common_wrappers
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_GITHUB_BASEURL
+import static org.edx.jenkins.dsl.YAMLHelpers.parseYaml
+import org.yaml.snakeyaml.error.YAMLException
 
 
 /*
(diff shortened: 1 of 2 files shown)


marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,128 @@
...
+                remote {
+                    url(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['githubRepoUrlSegment'] + '.git')
+                }
+                branch('*/master')
+                browser()
+            }
+        }
+
+        steps {
+            shell( "sudo rm -rf /usr/local/lib/python2.7/dist-packages/requests*")
```
schenedx: Why do you need to remove requests library from the local version. I thought virtual env will help you shield this off from the conflict?

schenedx:
@clintonb Can you help review this PR since I believe only you have some experience in this area besides @rlucioni

marketing/sailthruScriptJob.groovy
clintonb: I second this question. This seems very much wrong.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,128 @@
...
+            shell( requirements )
+            shell( sync_content )
+        }
+
+        triggers { //triggers daily
+            scm('@daily')
```
clintonb: We refresh metadata 4 times per day. Is there a reason this script doesn't follow the same cadence?

marketing/sailthruScriptJob.groovy
umar-nawaz: Yeah, I was also not sure about this but it worked. 
Well, I was having this error:
`ImportError: cannot import name IncompleteRead` 
May be I was only getting an issue on my local instance. @schenedx I shared console log with you in email for this.
Looked like some updates were required.

marketing/sailthruScriptJob.groovy
umar-nawaz: hm, currently it is run weekly and I don't have any reason for this. 
@schenedx can better explain this.

marketing/sailthruScriptJob.groovy
schenedx: @umar-nawaz So what you are saying is that ShiningPanda is still not configured correctly on your local Jenkins instance? I think you should spend the time to look into how ShiningPanda should be configured, rather than having this hack here. I can try to help more tomorrow if you are still stuck.

marketing/sailthruScriptJob.groovy
clintonb: Fine with me.

marketing/sailthruScriptJob.groovy
clintonb: The Docker instance is supposed to have everything configured: https://github.com/edx/jenkins-job-dsl/tree/master/docker. If that is not the case, please take time to update it so that we can avoid this issue in the future.

jibsheet: Jenkins does not have sudo access.
This code isn't using a virtualenv, which is why it's failling.

please look at something like Renzo's recent work https://github.com/edx-ops/edx-jenkins-job-dsl/blob/master/jobs/tools-edx-jenkins.edx.org/createTranslationJobs.groovy to get a sense of how this sort of thing should be configured.  Think about whether this is an internal job or something that should be public.  Think about credential management.  Also, we don't use scm, we use multiscm.  That's just the most basic review from what I saw in here around sudo, I don't have cycles to look at the rest of this PR right now.

umar-nawaz:
@schenedx @clintonb Please take a look at updates.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+    assert jobConfig.containsKey('githubRepoUrlSegment')
+    assert jobConfig.containsKey('jobName')
+    assert jobConfig.containsKey('notificationEmail')
+
+    job(jobConfig['jobName']) {
+
+        requirements = "pip install -r " + jobConfig['requirements']
+        sync_content = "python " + jobConfig['scriptPath'] + " --oauth_host "+ jobConfig['oauth_host'] + 
+                  " --oauth_key " + jobConfig['oauth_key'] + " --oauth_secret " + jobConfig['oauth_secret'] + 
+                  " --sailthru_key " + jobConfig['sailthru_key'] + " --sailthru_secret " + jobConfig['sailthru_secret'] +
```
jibsheet: Did you want your oauth keys and sailthru keys in splunk?  Because that's what will happen with this design.
Console logs of jenkins jobs run on tools-edx-jenkins are available in splunk to any user with access.

I'll point you again to Renzo's recent work on translations and using credentials.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+                remote {
+                    url(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['githubRepoUrlSegment'] + '.git')
+                }
+                branch('*/master')
+            }
+        }
+
+        steps {
+            virtualenv {
+                pythonName('System-CPython-2.7')
```
jibsheet: 3.5 is installed on the system, see Renzo's recent code.

umar-nawaz: fair point. okay, I give this a hit.

marketing/sailthruScriptJob.groovy
umar-nawaz: Good question. Yeah I believe, we should move these to splunk.

marketing/sailthruScriptJob.groovy
rlucioni: Verifying this locally will require building a new image with Python 3.5 included. See https://github.com/edx/jenkins-job-dsl/pull/84.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+    return 1
+}
+
+/* Iterate over the job configurations */
+secretMap.each { jobConfigs ->
+
+    Map jobConfig = jobConfigs.getValue()
+
+    /* Test secret contains all necessary keys for this job */
+    assert jobConfig.containsKey('oauth_host')
```
rlucioni: Why bother with this? If something's missing, why not let the job blow up?

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+Map config = [:]
+Binding bindings = getBinding()
+config.putAll(bindings.getVariables())
+PrintStream out = config['out']
+
+/* Map to hold the k:v pairs parsed from the secret file */
+Map secretMap = [:]
+out.println('Parsing secret YAML file')
+try {
+    /* Parse k:v pairs from the secret file referenced by secretFileVariable */
```
rlucioni: Don't do this parsing yourself. Use the helper we already have: https://github.com/edx-ops/edx-jenkins-job-dsl/blob/master/jobs/tools-edx-jenkins.edx.org/createTranslationJobs.groovy#L39

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+    String contents = new File("${EDX_SAILTHRU_SYNC_CONTENT}").text
```
rlucioni: Pull these secrets from the ecom-secure repo instead of a file on disk. Follow the pattern established here: https://github.com/edx-ops/edx-jenkins-job-dsl/blob/master/jobs/tools-edx-jenkins.edx.org/createTranslationJobs.groovy#L40

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+                branch('*/master')
+            }
+        }
+
+        steps {
+            virtualenv {
+                pythonName('System-CPython-2.7')
+                nature('shell')
+                command( requirements )
+                command( sync_content )
```
rlucioni: Look at how we use a single separate bash file for these commands. If you actually need to spawn multiple child jobs with different configs, you can use environment variables to interpolate job-specific bits into the script.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        triggers { //triggers daily
```
rlucioni: This comment adds no value.

marketing/sailthruScriptJob.groovy
```diff
@@ -0,0 +1,130 @@
...
+            scm('@daily')
+        }
+
+        wrappers {
+            timeout {
+               absolute(75)
+            }
+            timestamps()
+            colorizeOutput('gnome-terminal')
```
rlucioni: What's the point of colorizing output?

marketing/sailthruScriptJob.groovy
rlucioni: No, you should not move anything to Splunk. You need to mask any secrets so they don't end up visible in Splunk. https://github.com/edx-ops/edx-jenkins-job-dsl/blob/master/jobs/tools-edx-jenkins.edx.org/createTranslationJobs.groovy#L56

umar-nawaz:
Moved this work to [private repo here #58](https://github.com/edx-ops/edx-jenkins-job-dsl/pull/58)

PR #88, opened 2016-12-01
parameterize worker in subset job

estute:
@benpatterson 
stage 1: subset job can take a worker label. if deployed this will not affect our current set up, bc the testeng-ci dsls don't yet know about this param and the default value will be chosen


benpatterson:
👍

PR #89, opened 2016-12-01
add worker parameter to all master + pr jobs

estute:
@benpatterson 
stage 3: a11y, bokchoy, lettuce, python and quality now have either a label parameter OR string param which will be passed to the testeng-ci scripts (and then passed to subset as a label param)


estute:
@cpennington  as well

PR #90, opened 2016-12-01
mask e2e env vars

estute:
@benpatterson @Muddasser 
going to reseed tomorrow, but this good practice due to the fact the splunk indexes everything


estute:
@cpennington  as well

PR #91, opened 2016-12-05
add closure for masking passwords

estute:
@benpatterson @jzoldak 
this will allow users to input an ami base id (rather than having to make the change in the packer scripts. masked the param, as per @jibsheet advice


benpatterson:
Looks like you have to resolve a conflict.

estute:
@ben, yeah saw that. dealing with it. in the meantime, I have added more.. if you have a moment

benpatterson:
LGTM so far

PR #92, opened 2016-12-06
fix python+lettuce master scripts

estute:
@benpatterson @jzoldak 
found an error when trying to seed these

### DIFF
--- a/platform/jobs/edxPlatformLettuceMaster.groovy
+++ b/platform/jobs/edxPlatformLettuceMaster.groovy
@@ -98,7 +98,9 @@ secretMap.each { jobConfigs ->
         }
 
         parameters {
-            stringParam(stringParams.name, stringParams.default, stringParams.description)
+            stringParams.each { param ->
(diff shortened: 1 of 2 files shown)


benpatterson:
👍

platform/jobs/edxPlatformLettuceMaster.groovy
```diff
@@ -98,7 +98,9 @@ secretMap.each { jobConfigs ->
         }
 
         parameters {
-            stringParam(stringParams.name, stringParams.default, stringParams.description)
+            stringParams.each { param ->
```
jzoldak: @estute do we know why this was working before and then it wasn't? E.g. have any versions of jenkins/plugins changed or something?

estute: @jz this was just a last minute fix to a commit I made earlier in the day

jzoldak: OIC. makes sense. carry on. :)

PR #93, opened 2016-12-06
add a wrapper job for running tests on special boxes

estute:
@benpatterson @jzoldak 
Use case: some one wants to test out a new box (xenial upgrade, optimized workers). However we are currently unable to specify this via github webhook commands. This job will allow us to pick a PR and worker type and run as if it were triggered via GH

### DIFF
--- a/platform/jobs/edxPlatformPrTestsSpecialWorker.groovy
+++ b/platform/jobs/edxPlatformPrTestsSpecialWorker.groovy
@@ -1,0 +1,49 @@
+package platform
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+buildFlowJob('edx-platform-pr-tests-special-worker') {
+    
+    description('a wrapper job used to run a platform PR test suite on a given worker type')
+    logRotator JENKINS_PUBLIC_LOG_ROTATOR
+    label('flow-worker')
+
+    List<String> prJobs = [ 'edx-platform-accessibility-pr',
+                            'edx-platform-bok-choy-pr',
+                            'edx-platform-js-pr',
+                            'edx-platform-lettuce-pr',
+                            'edx-platform-python-unittests-pr',
+                            'edx-platform-quality-pr' ]
+
+    parameters {
+
+        // Supply `ghprbActualCommit` to the jobs triggered by this wrapper job
+        // `ghprbActualCommit` is an environment variable normally supplied via
+        // the ghprb, but we can override it
+        stringParam('ghprbActualCommit', '*/master', 'The hash of git commit to run downstream jobs on')
+        // The label to run
+        // Note: this label will be used to specify the worker on the following jobs:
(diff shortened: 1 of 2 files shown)


estute:
[te-1378]

platform/jobs/edxPlatformPrTestsSpecialWorker.groovy
```diff
@@ -0,0 +1,48 @@
...
+        // test-subset jobs):
+        //  - edx-platform-(bok-choy|lettuce-python-unittests)-pr => this is a flow job and does not
+        //      actually run any test
+        stringParam('WORKER_LABEL', 'jenkins-worker', 'The Jenkins worker type should be used to run this job')
+        choiceParam('JOB_TO_RUN', prJobs, 'The platform job to run, given the hash & worker')
+    }
+
+    wrappers {
+        timeout {
+            absolute(60)
```
benpatterson: Sadly this value won't work for bok-choy right now. Better to bump it up to 80 for now.

platform/jobs/edxPlatformPrTestsSpecialWorker.groovy
```diff
@@ -0,0 +1,48 @@
+package platform
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+buildFlowJob('edx-platform-pr-tests-special-worker') {
```
benpatterson: Not a must, but could we use the build descriptor plugin <sic> to assign a description to the job so it's easy to look at past jobs and understand what they were run for?

platform/resources/runJobOnWorker.groovy
```diff
@@ -0,0 +1,7 @@
+import hudson.model.*
+
+JOB_TO_RUN = build.environment.get('JOB_TO_RUN')
+ghprbActualCommit = build.environment.get('ghprbActualCommit')
+WORKER_LABEL = build.environment.get('WORKER_LABEL')
+
+build(JOB_TO_RUN, ghprbActualCommit: ghprbActualCommit, WORKER_LABEL: WORKER_LABEL)
```
jzoldak: @estute - I'm a little confused here. Is this a hudson build() method call? Does this line work?
I don't see this pattern in other scripts, tho I could be missing it. Could you pls include a link to the doc or another example?

estute: @jzoldak  This isn't part of the job dsl, it is part of the build flow dsl (see https://wiki.jenkins-ci.org/display/JENKINS/Build+Flow+Plugin). We use it in our testeng-ci scripts (ex: https://github.com/edx/testeng-ci/blob/master/jenkins/flow/pr/edx-platform-bok-choy-pr.groovy)

jzoldak: Ah, thank you. Could you please add a comment above this line with that first link? That would be super helpful.

estute:
addressed nits

PR #96, opened 2016-12-12
Add a link to the discussion of Credentials management in the wiki

jibsheet:


jibsheet:
@estute At a minimum, I'd like to link to our documentation of credentials management.  Ideally, I'd like to see TestEng add their thoughts into that same wiki space and then just make the README link there.

PR #99, opened 2016-12-15
fix issue with git browser

estute:
@benpatterson @jzoldak 
this fixes the stacktrace jenkins throws when the github browser is used, and then you try to manually edit a job.


jzoldak:
👍

benpatterson:
👍

PR #101, opened 2016-12-20
adjust ghprb in platform jobs to handle named release prs

estute:
@jzoldak @benpatterson 
PTAL when you get a chance. This allows a job to only build via GHPRB if the target branch of PR is in the jobs white list (or not in the white list ;) ). What that means is:
* normal PRs trigger the normal "edx-platform...pr" jobs. In the normal jobs, we white list every branch EXCEPT those who contain the "open-release/tree.master" naming scheme
* PRs against named release branches can only trigger the Jenkins job for that named release. Named release jobs ONLY white lists the target branch with the "open-release/tree.master" naming scheme

### DIFF
--- a/platform/jobs/edxPlatformAccessibilityPr.groovy
+++ b/platform/jobs/edxPlatformAccessibilityPr.groovy
@@ -2,9 +2,9 @@ package devops
 
 import org.yaml.snakeyaml.Yaml
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
-import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_JUNIT_REPORTS
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_GITHUB_BASEURL
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.GHPRB_WHITELIST_BRANCH
 
 /*
 Example secret YAML file used by this script
@@ -16,6 +16,10 @@ publicJobConfig:
     repoName : name-of-github-edx-repo
     credential : n/a
     cloneReference : clone/.git
+    workerLabel: worker-label
+    whitelistBranchRegex: 'release/*'
+    context: jenkins/test
+    triggerPhrase: 'jenkins run test'
 */
 
 /* stdout logger */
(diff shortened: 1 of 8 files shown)


estute:
I would like your opinion on this. We don't technically "need" a different trigger phrase for each worker type due to the white listing. However, I have implemented release based trigger phrases (i.e. 'dogwood run js') to explicitly show what you are trying to do. It is very easy to revert. Let me know which you prefer

jzoldak:
Caught up in person on the trigger phrase.
LGTM. 👍

platform/jobs/edxPlatformAccessibilityPr.groovy
```diff
@@ -107,11 +113,14 @@ secretMap.each { jobConfigs ->
...
                 extensions {
                     commitStatus {
-                        context('jenkins/a11y')
+                        context(jobConfig['context'])
                     }
                 }
             }
         }
+
+        configure GHPRB_WHITELIST_BRANCH(jobConfig['whitelistBranchRegex'])
```
benpatterson: Is this a `configure` statement b/c this particular function is not available in the ghprb DSL?

src/test/resources/platform/secrets/edx-platform-quality-pr-secret.yml
```diff
@@ -6,7 +6,10 @@ publicJobConfig:
     platformCredential : n/a
     platformCloneReference : edx-platform-clone/.git
     protocol : "https://github.com/"
- 
+    workerLabel: worker-label
+    whitelistBranchRegex: 'master'
+    context: 'jenkins/quality'
```
benpatterson: Why didn't you need to implement this kind of change for the other contexts?

platform/jobs/edxPlatformAccessibilityPr.groovy
estute: Exactly.

src/test/resources/platform/secrets/edx-platform-quality-pr-secret.yml
estute: Because I need to write more tests (and therefore more dummy secrets)

platform/jobs/edxPlatformAccessibilityPr.groovy
benpatterson: and for my own clarification, is it not available in this *version* of the plugin, or is it not-available at all?

estute:
@benpatterson @jzoldak Just a heads up, I spoke with the OS team and they actually wanted the trigger phrases to contain the named release (i.e. "ficus run a11y"), so I modified the job.

PR #107, opened 2017-01-10
parameterize testeng-ci branch

estute:
@jzoldak @benpatterson 

this will fix broken eucalyptus test, because testeng-ci needs to be tagged/branched in relative lockstep with the edx-platform

### DIFF
--- a/platform/jobs/edxPlatformBokChoyMaster.groovy
+++ b/platform/jobs/edxPlatformBokChoyMaster.groovy
@@ -106,20 +106,6 @@ secretMap.each { jobConfigs ->
             env('REPO_NAME', jobConfig['repoName'])
         }
         multiscm {
-            git { //using git on the branch and url, clean before checkout
-                remote {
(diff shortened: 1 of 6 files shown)


jzoldak:
Oh, that's what it was. Good find!
👍

benpatterson:
:+1:

estute:
@benpatterson in case you were curious...
The failing master tests were a side effect of the multi-scm plugin upgrade yesterday. Prior to the upgrade, the plugin offered one environment variable (GIT_COMMIT, which we use in the testeng-ci flow scripts) for storing information about the repos used by a job. This was part of why polling was not functioning correctly. The last clone within the scm section would be stored in this variable. However, the upgrade reversed this. The first repo cloned gets stored in GIT_COMMIT, and all subsequent repos cloned in the job get stored in GIT_COMMIT_1, GIT_COMMIT_2, etc.

benpatterson:
Ok honestly this rings a bell. Sounds good to me.

PR #110, opened 2017-01-18
Adds ora2 base directory with jobs and resources subdirectories

jlajoie:
Adds in base and subdirectories for TNL to work off of at a common starting point for the upcoming work on converting ora2 Jenkins jobs to the DSL in the following stories:

* [TNL-6976](https://openedx.atlassian.net/browse/TNL-6276)
* [TNL-6977](https://openedx.atlassian.net/browse/TNL-6277)
* [TNL-6978](https://openedx.atlassian.net/browse/TNL-6278)


jlajoie:
Just opening this up for now @jibsheet not sure what your process is going to be for us getting code merged in.

FYI: @cahrens @staubina

jibsheet:
Once the first few jobs have landed here, we'll tie them together with a parent seed job like this:
https://github.com/edx/jenkins-job-dsl-internal/blob/master/jobs/tools-edx-jenkins.edx.org/createCatalogJobs.groovy
Although you may not need to deal with as much (or any) configuration parsing/passing and instead you may just be able to instantiate and call into your other jobs.  We'll need to pass along a folder.

jlajoie:
Closing this PR in favor of https://github.com/edx/jenkins-job-dsl/pull/112

PR #111, opened 2017-01-20
change base ami to be a string instead of masked

estute:
@jzoldak @benpatterson Doing this because feeding the job incorrect values has bitten us in the past. The job itself is private, so perhaps this is less of a concern than previously thought.


jzoldak:
👍

estute:
Note: Failure was not related to this change

PR #112, opened 2017-01-20
TNL-6334 Add ORA2 Add Course to Sandbox job

jlajoie:
### Description

Adds a job to the ora2 directory to create the add-course-to-sandbox jenkins job.

@jibsheet not sure if you're the one to review these, but this should be good to go depending on how the seeds are handled and if we're able to pass in yaml not in a file.

**FYI**
@staubina 
@cahrens 

- [ ] Rebase and squash commits

### DIFF
--- a/ora2/jobs/AddCourseToSandbox.groovy
+++ b/ora2/jobs/AddCourseToSandbox.groovy
@@ -1,0 +1,41 @@
+/*
+    Variables consumed from the EXTRA_VARS input to your seed job in addition
+    to those listed in the seed job.
+
+    * FOLDER_NAME: some-folder
+    * SSH_AGENT_KEY : your-ssh-credential-name
+    * SSH_USER : your-ssh-username
+    * NOTIFY_ON_FAILURE : e-mail-contact@example.com
+*/
+
+package ora2.jobs
+
+class AddCourseToSandbox {
+
+    public static def job = { dslFactory, extraVars ->
+        return dslFactory.job(extraVars.get('FOLDER_NAME', 'ora2') + '/add-course-to-sandbox') {
+            properties {
+                parameters {
+                    stringParam(
+                        'SANDBOX_BASE',
+                        'ora2',
+                        "For example, when the sandbox is foo.sandbox.edx.org, the sandbox_base is 'foo'"
+                    )
+                }
+            }
+            steps {
+                shell(
+                    "SSH_USER=${extraVars.get('SSH_USER', 'ubuntu')}\n" +
+                    dslFactory.readFileFromWorkspace('ora2/resources/ssh-install-test-course.sh')
+                )
+            }
+            publishers {
+                mailer(extraVars['NOTIFY_ON_FAILURE'], true, false)
+            }
+            wrappers {
+                sshAgent(extraVars['SSH_AGENT_KEY'])
+            }
+        }
+    }
+
+}
--- a/ora2/resources/ssh-install-test-course.sh
+++ b/ora2/resources/ssh-install-test-course.sh
@@ -1,0 +1,3 @@
+# Install a test course
+SANDBOX_HOST=${SANDBOX_BASE}.sandbox.edx.org
+ssh ${SSH_USER}@${SANDBOX_HOST} -o StrictHostKeyChecking=no "sudo su edxapp -s /bin/bash -c 'source /edx/app/edxapp/edxapp_env && /edx/app/edxapp/venvs/edxapp/src/edx-ora2/scripts/install-test-course.sh'"


cahrens:
@jibsheet are you able to review this PR tomorrow morning? It would be good to know if we are on the same page ahead of our iteration planning tomorrow afternoon.

jibsheet:
@cahrens As I mentioned to Jeff "Maybe".  Take a look at my calendar for today and tomorrow if you want a sense of why I say that.  I'm trying to land operational work prior to the end of the week myself.

cahrens:
@jibsheet OK, thanks for the update.

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+    * folder-name: some-folder
+    * credential : your-ssh-credential-name
+    * email : e-mail-contact@example.com
+*/
+
+package ora2.jobs
+
+import org.yaml.snakeyaml.Yaml
+import org.yaml.snakeyaml.error.YAMLException
```
jibsheet: These appear unused in this job

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,56 @@
+/*
+    Variables consumed from the EXTRA_VARS input to your seed job in addition
+    to those listed in the seed job.
+
+    * folder-name: some-folder
+    * credential : your-ssh-credential-name
+    * email : e-mail-contact@example.com
```
jibsheet: Can you use the style of var we've used in other jobs.  We tend towards ALL_CAPS.  Please also name credential something more specific (SSH_KEY).  Similarly for email - something like contact email?

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,56 @@
...
+                            throttleMatrixBuilds(false)
+                            throttleMatrixConfigurations(false)
+                        }
+                    }
+                }
+                steps {
+                    shell(dslFactory.readFileFromWorkspace('ora2/resources/ssh-install-test-course.sh'))
+                }
+                publishers {
+                    mailer(extraVars['email'], false, false)
```
jibsheet: Check with @cahrens about what extra vars we want here - 
https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.mailer
I suspect we may actually want true, false but I always forget exactly how the poorly named anti-boolean parameters work.

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,56 @@
...
+                    }
+                    parameters {
+                        stringParam(
+                            'SANDBOX_BASE',
+                            'ora2',
+                            "For example, when the sandbox is foo.sandbox.edx.org, the sandbox_base is 'foo'"
+                        )
+
+                    }
+                    throttleJobProperty {
```
jibsheet: Are any of these defaults, or are they all necessary overrides?  I wonder if some of these are just weird defaults from the original job.

ora2/resources/ssh-install-test-course.sh
```diff
@@ -0,0 +1,3 @@
+# Install a test course
+SANDBOX_HOST=${SANDBOX_BASE}.sandbox.edx.org
+ssh ubuntu@${SANDBOX_HOST} -o StrictHostKeyChecking=no "sudo su edxapp -s /bin/bash -c 'source /edx/app/edxapp/edxapp_env && /edx/app/edxapp/venvs/edxapp/src/edx-ora2/scripts/install-test-course.sh'"
```
jibsheet: Can you gist a console log of the job run?  Just want to get a sense of output.

ora2/jobs/AddCourseToSandbox.groovy
jlajoie: Will double check with @cahrens, but yes, it looks like "dontNotifyEveryUnstableBuild" should be true, because we don't want to do that, since that's been a major complaint about the ora2 jobs that we've had.

ora2/jobs/AddCourseToSandbox.groovy
jlajoie: I think these are all of the defaults, I'll try running without them enabled to see what XML gets generated.

ora2/jobs/AddCourseToSandbox.groovy
jlajoie: Sure thing 👍

ora2/jobs/AddCourseToSandbox.groovy
jlajoie: Left the package, removed the unneeded imports :+1:

jlajoie:
@jibsheet Made the requested updates. Output from a run can be found [here](https://gist.github.com/jlajoie/8474255f815d570b15140194aafcb6d0). Just noticed I don't have SMTP set up properly on my local to successfully send e-mails, so I can re-run that. Also added in a way to specify the ssh username as part of the seed job for the shell script.

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,40 @@
+/*
+    Variables consumed from the EXTRA_VARS input to your seed job in addition
+    to those listed in the seed job.
+
+    * JOB_FOLDER_NAME: some-folder
+    * SSH_CREDENTIAL_NAME : your-ssh-credential-name
+	* SSH_USERNAME : your-ssh-username
```
jibsheet: nit: formatting

ora2/jobs/AddCourseToSandbox.groovy
```diff
@@ -0,0 +1,40 @@
...
+                        'SANDBOX_BASE',
+                        'ora2',
+                        "For example, when the sandbox is foo.sandbox.edx.org, the sandbox_base is 'foo'"
+                    )
+                }
+            }
+            steps {
+                shell("SSH_USERNAME=${extraVars.get('SSH_USERNAME', 'ubuntu')}\n" + 
+					dslFactory.readFileFromWorkspace('ora2/resources/ssh-install-test-course.sh')
+				)
```
jibsheet: Can we get these to indent/lineup better, similar to publishers/wrappers below

ora2/jobs/AddCourseToSandbox.groovy
jlajoie: Ah right on, looks like hard tabs vs soft tabs here, can fix that and the formatting below.

PR #113, opened 2017-01-27
Ensure this plugin is on the docker image

jibsheet:
@e0d cc/@edx/devops


jibsheet:
edxops/jenkins:latest has been updated

PR #115, opened 2017-02-01
refactor e2e job for gocd pipeline testing

estute:
@jzoldak @benpatterson 

Notes:
* We had conflated the use cases for e2e tests in the past, i.e. PRs and merge testing for the tests and running the test as part of a CD pipeline. This job is only for the latter.
* Name will change once I get a thumb to 'edx-e2e-tests', so that no change is required from the pipeline code. I have this other name just for testing on build jenkins.
* I think we can eschew yaml-secrets in favor of credential binding for this, especially because we don't need to create variants on this job. However this exposes the env var names. Are you concerned or am I paranoid?
* I plan on putting in a PR to edx-e2e-tests to change the default screenshot dir, and will change this once I do. But this will get it working with the current state of the tests & pipeline.

### DIFF
--- a/platform/jobs/edxEndToEndTestsJob.groovy
+++ b/platform/jobs/edxEndToEndTestsJob.groovy
@@ -1,0 +1,96 @@
+package platform
+
+import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
+
+// The edx-e2e-tests job is run automatically on every deployment of the edx-platform
+// from the gocd pipeline.
+
+/* stdout logger */
+/* use this instead of println, because you can pass it into closures or other scripts. */
+Map config = [:]
+Binding bindings = getBinding()
+config.putAll(bindings.getVariables())
+PrintStream out = config['out']
+
+/* Map to hold the k:v pairs parsed from the secret file */
+Map mailingListMap = [:]
+try {
+    out.println('Parsing secret YAML file')
+    String mailingListSecretContents  = new File("${MAILING_LIST_SECRET}").text
+    Yaml yaml = new Yaml()
+    mailingListMap = yaml.load(mailingListSecretContents)
+    out.println('Successfully parsed secret YAML file')
+}
+catch (any) {
+    out.println('Jenkins DSL: Error parsing secret YAML file')
+    out.println('Exiting with error code 1')
+    return 1
+}
+
+assert mailingListMap.containsKey('e2e_test_mailing_list')
+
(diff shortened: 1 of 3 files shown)


platform/jobs/edxEndToEndTestsJob.groovy
```diff
@@ -0,0 +1,70 @@
...
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
+
+// The edx-e2e-tests job is run automatically on every deployment of the edx-platform
+// from the gocd pipeline.
+
+job('NEW-edx-e2e-test') {
+
+    description('Run end-to-end tests against the an instance of the edx-platform')
```
jzoldak: typo "the an"

platform/jobs/edxEndToEndTestsJob.groovy
```diff
@@ -0,0 +1,70 @@
+package platform
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
+
+// The edx-e2e-tests job is run automatically on every deployment of the edx-platform
+// from the gocd pipeline.
+
+job('NEW-edx-e2e-test') {
```
jzoldak: are you temporarily making this with a new job name so that we can reseed without destroying what's there now?

platform/jobs/edxEndToEndTestsJob.groovy
```diff
@@ -0,0 +1,70 @@
...
+        shell('jenkins/end_to_end_tests.sh')
+    }
+
+    publishers {
+        archiveJunit('reports/*.xml') {
+            allowEmptyResults(false)
+        }
+        archiveArtifacts {
+            pattern('reports/*.xml')
+            pattern('log/*.png')
```
jzoldak: couple things:
1. we also need to archive *.log files from the log dir (not just *.png). Probably makes sense to just slurp up anything that is put there.
2. please also archive anything under the 'screenshots' folder

estute:
@jzoldak done

platform/jobs/edxEndToEndTestsJob.groovy
```diff
@@ -0,0 +1,96 @@
...
+    out.println('Jenkins DSL: Error parsing secret YAML file')
+    out.println('Exiting with error code 1')
+    return 1
+}
+
+assert mailingListMap.containsKey('e2e_test_mailing_list')
+
+job('NEW-edx-e2e-test') {
+
+    description('Run end-to-end tests against the an instance of the edx-platform')
```
jzoldak: nit: typo "the an"

PR #119, opened 2017-02-06
adjust e2e job permissions and add e2e view

estute:
@jzoldak @benpatterson minor fix to the security of this so that bots can use the pipeline job. Also, added a view and codenarc fixes


platform/jobs/edxEndToEndTestsJob.groovy
```diff
@@ -30,14 +32,15 @@ catch (any) {
 }
 
 assert mailingListMap.containsKey('e2e_test_mailing_list')
+assert securityGroupMap.containsKey('e2eTestUsers')
```
benpatterson: is this supposed to be `e2eSecurityGroup`?

PR #120, opened 2017-02-13
Most RunAnsible jobs have run in a single AWS account per folder

jibsheet:
Extend this so we can create ones that use several different credentials in a single folder.  This lets me have edge jobs next to edx jobs, rather than needing to have folders per environment.

@edx/devops 

FYI @rlucioni since as your teams write more of these jobs, you may finally need something to run in Edge :)


rlucioni:
Thanks for the heads-up @jibsheet.

PR #123, opened 2017-02-16
In particular, forgetting SSH_AGENT_KEY leads to inscrutable errors

jibsheet:
This wil make it clearer what config you forgot to set up in EXTRA_VARS

@edx/devops

### DIFF
--- a/devops/jobs/RunAnsible.groovy
+++ b/devops/jobs/RunAnsible.groovy
@@ -37,11 +37,14 @@ import static org.edx.jenkins.dsl.DevopsConstants.common_read_permissions
 class RunAnsible {
     public static job(DslFactory dslFactory, String jobName, String environment, String deployment, Map extraVars) {
         return dslFactory.job(extraVars.get("FOLDER_NAME") + "/${environment}-${deployment}-${jobName}") {
+            // These 3 don't have defaults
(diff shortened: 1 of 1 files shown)


jibsheet:
@estute FYI - I started using assert() and I found the : very helpful (gives a better error message, a bare assert was generating a ginormous stack trace for me which is harder for folks to parse)

devops/jobs/RunAnsible.groovy
```diff
@@ -37,11 +37,14 @@ import static org.edx.jenkins.dsl.DevopsConstants.common_read_permissions
 class RunAnsible {
     public static job(DslFactory dslFactory, String jobName, String environment, String deployment, Map extraVars) {
         return dslFactory.job(extraVars.get("FOLDER_NAME") + "/${environment}-${deployment}-${jobName}") {
+            // These 3 don't have defaults
+            assert extraVars.containsKey('SSH_AGENT_KEY') : "Please defined SSH_AGENT_KEY"
```
estute: nit: typo

estute:
@jibsheet may/may not make sense to assert on all necessary variables. This can be helpful, especially if any of them are stored in secrets and therefore difficult to determine their values

jibsheet:
@estute Yeah, these are the 3 for this job.  The rest have sensible defaults in the .get('var','default').
I plan to sprinkle some more in the seed jobs for similar reasons, but started here.

PR #124, opened 2017-02-16
removing docker-in-docker functionality

MichaelRoytman:
Does it make sense to remove the docker-in-docker directory? Or is it worth the effort to work on finding a way to use Docker inside a Docker container without needing a separate Dockerfile?


jibsheet:
I think we should leave this - at some point soon we'll revisit wanting this

MichaelRoytman:
We did this here: https://github.com/edx/jenkins-job-dsl/pull/169. I'm closing this PR.

PR #127, opened 2017-03-02
removed programs e2e test job

estute:
@benpatterson 
re: TE-1998


estute:
@benpatterson I checked and there is nothing else to remove from the repo (wrt this job). In other words, there are no constants specific to this job.

benpatterson:
Ship it, I say.

PR #129, opened 2017-03-20
Add NR key from secrets to the packer job.

benpatterson:


jibsheet:
Heads up that NR is paging us on the jenkins workers using a lot of CPU
https://rpm.newrelic.com/accounts/88178/servers/50598828?tw%5Bdur%5D=last_30_minutes
Needs to be some config in that, not sure where (maybe we should make you folks a sub-account?)

benpatterson:
Tested on test-Jenkins as well. Looked good. Merging.

PR #132, opened 2017-03-28
add jenkins_worker_loadtest.json to the list of packer scripts

pwnage101:
I will use this to test with configuration branch `pwnage101/loadtest-worker`.

PERF-417


estute:
:+1:

PR #136, opened 2017-03-31
remove unused e2e dsl jobs

estute:
@jzoldak we are now using platform/jobs/edxEndToEndTestsJob.groovy for all of the e2e style jobs


jzoldak:
Just saw this PR now (like majorly late to the party!)
@estute I don't see those files there now so we must have removed them in another PR.

PR #141, opened 2017-04-05
clarify additional steps required for creating the Job Creator

pwnage101:
here are the updated steps I took to get a job seeded.


pwnage101:
FYI @e0d since you wrote most of these lines.

jibsheet:
@pwnage101  Worth revisiting - I just merged some doc updates and removed some things.  It seems to have caused conflicts.

PR #142, opened 2017-04-07
add microsites environment variables to e2e jobs

estute:
@benpatterson @Muddasser @jzoldak 
As we port white label tests into master, we need to migrate some of the environment variables to be shared between the two


benpatterson:
👍

PR #146, opened 2017-04-11
TE-2034: re-enable pa11ycrawler on master job.

benpatterson:
@estute @arizzitano @michaelyoungstrom 

As per feedback, here is the same change, but codified in the DSL. Note it's master only.


benpatterson:
Closing this since we'll always be running the crawler and therefore no switch is needed. See #147

PR #148, opened 2017-04-19
Add a new job: loadtest-driver

pwnage101:
The new jenkins job DSL testeng/jobs/loadtestDriver.groovy creates the
loadtest-driver job.  It is not yet capable of running distributed
loadtests, or allowing secrets to be specified.

PERF-393

### DIFF
--- a/testeng/jobs/loadtestDriver.groovy
+++ b/testeng/jobs/loadtestDriver.groovy
@@ -1,0 +1,125 @@
+package testeng
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_TEAM_SECURITY
+
+/* loadtestDriver.groovy
+ *
+ * Unlike most other jenkins jobs, loadtest-driver progressively refines its
+ * results by continuous sampling.  Currently there is no tooling around
+ * determining if a statistically significant amount data has been collected,
+ * (not impossible to build, but it would likely depend on many external
+ * factors).  Thus, this job relies on a human to manually abort it, or to set
+ * the MAX_RUN_TIME parameter.
+ *
+ * Instructions:
+ *
+ * 1. Ensure the "jenkins worker loadtest driver" AMI has been prepared (from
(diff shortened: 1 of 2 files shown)


testeng/jobs/loadtestDriver.groovy
```diff
@@ -0,0 +1,112 @@
...
+
+    /* This worker label corresponds to an AMI built using
+     * util/packer/jenkins_worker_loadtest.json from edx/configuration.
+     */
+    label('loadtest-driver-worker')
+
+    scm {
+        git {
+            remote {
+                url('https://github.com/edx/edx-load-tests')
```
jibsheet: Do we care about making this configurable?  Do we still need to run anything from the old repo?

testeng/jobs/loadtestDriver.groovy
```diff
@@ -0,0 +1,112 @@
+package testeng
+
+import hudson.util.Secret
```
jibsheet: I'm missing where this is used?

testeng/jobs/loadtestDriver.groovy
```diff
@@ -0,0 +1,112 @@
...
+import org.yaml.snakeyaml.Yaml
```
jibsheet: Same - missing where it's used

testeng/jobs/loadtestDriver.groovy
pwnage101: The old repo is missing the new helper scripts, and none of the settings file stuff has been backported.  I'd rather create work items to copy the old locust tests into the current repo, rather than backport any features.  If that makes sense to you, I'll create a ticket to track that separately.  I scanned a bit of the old test code, and I didn't come across any sensitive information.

jibsheet: Yep - moving to the new repo is definitely the goal, I was just checking that we weren't forgetting to make this configurable. 👍

PR #152, opened 2017-05-03
Create two new CI jobs

jibsheet:
Add a VERBOSE flag to the sandbox job, useful for people debugging and
also used by the CI jobs.

Switch to the more modern pattern for acls

@edx/devops

### DIFF
--- a/devops/jobs/CreateSandboxCI.groovy
+++ b/devops/jobs/CreateSandboxCI.groovy
@@ -1,0 +1,85 @@
+/*
+   This create a SandboxCIHourly and SandboxCIDaily job that run the CreateSandbox job,
+   targetting an int and int-nightly sandbox for testing.  It's assumed you will seed this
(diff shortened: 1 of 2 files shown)


devops/jobs/CreateSandboxCI.groovy
```diff
@@ -0,0 +1,85 @@
...
+        steps {
+          downstreamParameterized {
+            trigger(extraVars.get("FOLDER_NAME","Sandboxes") + "/CreateSandbox") {
+              block {
+                buildStepFailure('FAILURE')
+                failure('FAILURE')
+                unstable('UNSTABLE')
+              }
+              parameters {
+                booleanParam('VERBOSE',true)
```
feanil: These only need to happen for the 'hourly' job because the daily job runs a different playbook which should have all of these in it.

jibsheet: Moved around

PR #153, opened 2017-05-04
Don't run xqueue and xserver regularly.

feanil:
This code doesn't change very often so we can catch them with the nightly instead.

### DIFF
--- a/devops/jobs/CreateSandboxCI.groovy
+++ b/devops/jobs/CreateSandboxCI.groovy
@@ -61,8 +61,6 @@ class CreateSandboxCI {
                   booleanParam('reconfigure',true)
                   booleanParam('recreate',false)
                   predefinedProp('name_tag','continuous-integration')
-                  booleanParam('xqueue',true)
-                  booleanParam('xserver',true)
                   booleanParam('ecommerce_worker',true)
                   booleanParam('credentials',true)
                 }


devops/jobs/CreateSandboxCI.groovy
```diff
@@ -61,8 +61,6 @@ class CreateSandboxCI {
                   booleanParam('reconfigure',true)
                   booleanParam('recreate',false)
                   predefinedProp('name_tag','continuous-integration')
-                  booleanParam('xqueue',true)
-                  booleanParam('xserver',true)
```
jibsheet: You want to move these up into the Daily one.  They default to false on the sandbox job (so a move rather than a delete)

feanil: The daily one runs the from scratch which will definitely run these.

https://github.com/edx/configuration/blob/master/playbooks/edx-east/edx_continuous_integration.yml

jibsheet: Right, I keep forgetting that.  I should really have made it a standalone task that just runs a subset of ansible-provision.sh.  Some other day.

jibsheet:
I'm about to reseed, so I'll want for you to merge

jibsheet:
Seeded https://tools-edx-jenkins.edx.org/job/Seed%20Jobs/job/Sandbox%20Seed%20Jobs/41/console

PR #155, opened 2017-05-17
Added seed job and updated docker-compose.yml

clintonb:
A seed job has been added from https://github.com/sheehan/job-dsl-gradle-example. This allows developers to run a command against the Jenkins REST API to create the seed job rather than manually creating the job.

Additionally, the Docker Compose file has been updated to use the official Jenkins image. It is unclear what happened to the original Dockerfile.

LEARNER-509


clintonb:
@edx/testeng @edx/devops please review

PR #158, opened 2017-05-19
Added end-to-end test jobs for IDAs

clintonb:
LEARNER-509


clintonb:
@edx/testeng what's the procedure for adding a new credential to Jenkins?

clintonb:
Moving to internal repo to avoid issues with secrets.

PR #162, opened 2017-06-09
load tests use build flows to generate summaries

pwnage101:
Create a layer of indirection by changing the user-facing load test job
to a new build flow job (run-simple-loadtest) that builds downstream
jobs to run load tests (loadtest-driver) and generate summaries
(loadtest-summary) upon completion.

---

All artifacts, including the logfile and the summary report, become slurped up to the top level job.

This PR depends on https://github.com/edx/edx-load-tests/pull/106

An example jenkins test run containing two build artifacts (one logfile, one summary file): https://test-jenkins.testeng.edx.org/job/run-simple-loadtest/44/

### DIFF
--- a/testeng/jobs/loadtestDriver.groovy
+++ b/testeng/jobs/loadtestDriver.groovy
@@ -38,51 +38,127 @@ import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_TEAM_SEC
  *        ID: loadtest_basic_auth_settings.yml
  */
 
-job('loadtest-driver') {
+/* Define job parameters centrally so that they can be reused across two jobs
+ * (run-simple-loadtest and loadtest-driver).
+ */
+stringParams = [
+    [
+    name: 'TARGET_URL',
+    description: 'The loadtest target where requests are to be sent.',
+    default: 'https://courses-loadtest.edx.org'
+    ],
+    [
+    name: 'TEST_COMPONENT',
+    description: 'The desired component to loadtest. See complete list here: ' +
+                 'https://github.com/edx/edx-load-tests/tree/master/loadtests',
+    default: 'lms'
+    ],
+    [
+    name: 'REMOTE_BRANCH',
+    description: 'Branch of the edx-load-tests repo to use.',
+    default: 'master'
+    ],
+    [
+    name: 'NUM_CLIENTS',
+    description: 'This many locust clients (i.e. fake users) will be hatched.',
+    default: '500'
+    ],
+    [
+    name: 'HATCH_RATE',
+    description: 'Locust clients will be hatched at this rate ' +
+                 '(hatches/second).',
+    default: '30'
+    ],
+    [
+    name: 'MAX_RUN_TIME',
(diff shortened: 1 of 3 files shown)


testeng/resources/flow-run-simple-loadtest.groovy
```diff
@@ -0,0 +1,38 @@
...
+    driver_build.workspace,
+    'edx-load-tests/results/log.txt'
+)
+String loadtest_log = loadtest_logfile.readToString()
+
+/* Kick off a build to generate a load test summary */
+def summary_build = build(
+    'loadtest-summary',
+    REMOTE_BRANCH: params["REMOTE_BRANCH"],
+    LOG: loadtest_log
```
estute: Are there any size limitations you need to consider for text params? I'm not concerned about this log maxing out a java string, but I haven't inspected the text param definition

pwnage101: Skimming over the jenkins code, I can't see any obvious place where arbitrary limits are set on StringParameterValue, which TextParameterValue extends:

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/StringParameterValue.java

Seems like the limit is on the size String, which is related to system memory and jvm configuration.  These logfiles grow during the lifetime of the test; I estimate a 24h test produces a 200 MB log which is still an order of magnitude smaller than the String max size.

testeng/jobs/loadtestDriver.groovy
```diff
@@ -131,7 +207,81 @@ job('loadtest-driver') {
...
+
+    wrappers {
+        buildUserVars() /* gives us access to BUILD_USER_ID, among other things */
+        buildName('#${BUILD_NUMBER} by ${ENV,var="BUILD_USER_ID"}')
+        timestamps()
+        colorizeOutput('xterm')
+    }
+
+    steps {
+        copyArtifacts('run-simple-loadtest') {
```
pwnage101: added this `copyArtifacts` step.

testeng/jobs/loadtestDriver.groovy
```diff
@@ -131,7 +207,81 @@ job('loadtest-driver') {
...
+    description('Summarize a past load test.')
+
+    /* Abusing the team security feature to give all job control
+     * permissions to all edx employees.
+     */
+    authorization JENKINS_PUBLIC_TEAM_SECURITY.call(['edx'])
+
+    parameters {
+        stringParam('REMOTE_BRANCH', 'master',
+                    'Branch of the edx-load-tests repo to use.')
```
pwnage101: removed the LOG parameter because the copyArtifacts plugin knows where to find the log on its own.

pwnage101:
@estute github doesn't show changes since last force-push, so i added github comments to clarify differences.

estute:
:+1: Excellent! Do we need to install this plugin on build Jenkins?

pwnage101:
For future reference: 1.32.1 installed on build jenkins, 1.33 installed on test-jenkins.  The difference between the two, according to the changelog, is a non-issue. https://plugins.jenkins.io/copyartifact

PR #168, opened 2017-06-14
export ec2 cache for running ec2.py

hchen15:
add ec2-cache for run ansible so that caching errors do not occur when running ec2.py

### DIFF
--- a/devops/resources/run-ansible.sh
+++ b/devops/resources/run-ansible.sh
@@ -7,7 +7,7 @@ pip install -r requirements.txt
 pip install awscli
 
 env
-
+export EC2_CACHE_PATH="ec2-cache"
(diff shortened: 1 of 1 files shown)


devops/resources/run-ansible.sh
```diff
@@ -7,7 +7,7 @@ pip install -r requirements.txt
 pip install awscli
 
 env
-
+export EC2_CACHE_PATH="ec2-cache"
```
fredsmith: does this directory exist?  Does it need to?  '.' is a special unix directory meaning the current directory, so it's guaranteed to exist.  You can also use mktemp to generate a temp directory for each workspace, but '.' had a nice side-effect of being managed by the jenkins workspace for each assumed role/job.

PR #173, opened 2017-06-19
create app permissions job and shell script

hchen15:


feanil:
@hchen15 for this job we'll want to give access to other teams so you'll want to do something like [this](https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/CreateSandboxSSHAccess.groovy#L31-L38) and pass in a github team name so that a particular team has access to run the jobs manually if needed.  You can see [here](https://github.com/edx/edx-internal/blob/master/tools-edx-jenkins/enterprise.yml#L14-L15) what the team looks like in the config.  The team for this job will be `edx*jenkins-tools-it-jobs`

PR #182, opened 2017-06-29
Rename xss linter.

robrap:
The XSS Linter is being renamed, and this is a dependent PR that updates the Quality Report tabs appropriately.

Here is the main PR where the linter is being renamed:
https://github.com/edx/edx-platform/pull/15436

### DIFF
--- a/platform/jobs/edxPlatformQualityMaster.groovy
+++ b/platform/jobs/edxPlatformQualityMaster.groovy
@@ -33,7 +33,7 @@ archiveReports += 'edx-platform*/test_root/log/*.log, edx-platform*/test_root/lo
 archiveReports += 'edx-platform*/**/nosetests.xml,edx-platform*/**/TEST-*.xml'
 
 String htmlReports = 'pylint/*view*/, pep8/*view*/, jshint/*view*/, python_complexity/*view*/,'
-htmlReports += 'safecommit/*view*/, safelint/*view*/, eslint/*view*/'
+htmlReports += 'xsscommitlint/*view*/, xsslint/*view*/, eslint/*view*/'
 
 /* stdout logger */
 /* use this instead of println, because you can pass it into closures or other scripts. */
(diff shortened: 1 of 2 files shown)


platform/jobs/edxPlatformQualityMaster.groovy
```diff
@@ -33,7 +33,7 @@ archiveReports += 'edx-platform*/test_root/log/*.log, edx-platform*/test_root/lo
 archiveReports += 'edx-platform*/**/nosetests.xml,edx-platform*/**/TEST-*.xml'
 
 String htmlReports = 'pylint/*view*/, pep8/*view*/, jshint/*view*/, python_complexity/*view*/,'
-htmlReports += 'safecommit/*view*/, safelint/*view*/, eslint/*view*/'
+htmlReports += 'xsscommitlint/*view*/, xsslint/*view*/, eslint/*view*/'
```
estute: We should support both lint-file paths for a while (to not break existing PRs or require that everyone rebase). In a week or two, we can remove the old-style path

robrap: @estute I think it is important to not break builds, but it is not important to have these tabs during the transition.  Would you be ok with me removing the tabs in one PR and then in a week or two adding them with the new names in a new PR?  I want to keep the cost down on this change.

robrap: Actually, it doesn't break the build.  So I would vote we just go ahead and I'll communicate about the transition:

Here is a link to my PR that changes where the report goes:
https://github.com/edx/edx-platform/pull/15436
Here is the Jenkins build that passed, but just fails if you try to use that tab in the Quality Report:
https://build.testeng.edx.org/job/edx-platform-quality-pr/36506/Quality_Report/

robrap:
@estute I just merged the other PR.  We agree that this will be fine.  Can you approve?  Thanks.

PR #183, opened 2017-06-29
change job trigger to webhooks

hchen15:


jibsheet:
We're pulling out the parameters because this runs automatically?
Should we instead allow you to run manually against a branch to test things (would be helpful if you need to tests a config change related to the playbook)
It would mean adding a new variable, APP_PERMISSIONS_PROJECT_URL or some such.

hchen15:
@jibsheet I was defining the configuration repo/branch and the app-permissions repo/branch as parameters, but the webhooks didn't work because these params weren't defined until run time. So everything was null, and I had to change it to hardcode everything in

PR #186, opened 2017-07-05
Allow us to specify a github team

jibsheet:
Rotate logs after 7 days

Turn on our common wrappers
  Timestamps in console
  Mask Passwords
  Enable USER vars


jibsheet:
FYI @macdiesel  this is what will let me re-seed and grant permissions

PR #194, opened 2017-07-12
Fix edx.org theme issue on sandboxes

tuchfarber:
See WL-1104. Variable name was changed in configuration files and was
not mirrored in sandbox creation scripts. Also, theme directory was
incorrect.


jibsheet:
Can you update these docs, since the var was named wrong, and should be automatically set anyway.

https://openedx.atlassian.net/wiki/display/EdxOps/Sandboxes?focusedCommentId=161960215#comment-161960215

PR #195, opened 2017-07-12
Revert "Fix edx.org theme issue on sandboxes"

jibsheet:
Reverts edx/jenkins-job-dsl#194

This is being reverted because ansible-provision.sh was changed to expect the new variable name but it wasn't passing a list, it was passing a var.

This can either be fixed in the Jenkins DSL to pass a list or in ansible-provision to create a list when passing it down.

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -129,7 +129,7 @@ class CreateSandbox {
                 stringParam("configuration_secure_version","master","")
                 stringParam("configuration_internal_version","master","")
                 booleanParam("reconfigure",false,"Reconfigure and deploy, this will also run with --skip-tags deploy against all role <br />Leave this unchecked unless you know what you are doing")
-                choiceParam("edxapp_comprehensive_theme_dirs",["/edx/app/edxapp/edx-platform/themes","unset"],"")
+                choiceParam("edxapp_comprehensive_theme_dir",["/edx/app/edxapp/edx-platform/themes/edx.org","unset"],"")
                 booleanParam("testcourses",true,"")
                 booleanParam("performance_course",true,"")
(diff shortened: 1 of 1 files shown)


jibsheet:
FYI @tuchfarber

tuchfarber:
@jibsheet Could you give me a little more information on what would be needed to fix this? I'm not entirely sure how these interconnect, so I'm not sure what "passing it down" means in this context.

jibsheet:
Jenkins calls ansible-provision and changes the name of the variable passed from jenkins.
It did not change jenkins.  This means the variable we're using on sandboxes is null, causing warnings on literally every request on a sandbox right now unless you manually fix the theming config (go stand up a sandbox and read the logs).

Your change to DSL made the names match, but you passed a scalar to something expecting a list.  This meant sandboxes could not be provisioned at all.

Fix this code to pass a list, and change the variable name again.

but I stand by my comment that the two variables are not actually backwards compatible since one path contains edx.org and one contains a directory to look for edx.org in.  Your team designed the change and should be able to provide background, but at one time, theming on sandboxes worked out of the box (until this change was made).

tuchfarber:
I see, thanks for helpful explanation. I'll have to check with the team to see how they want to proceed.

@douglashall Could you lend and eye here?

douglashall:
@jibsheet Could you point out where/how we would pass a list as you have described above?

jibsheet:
Look at any of the other parameters here that pass a non-scalar?
The basic auth nginx users parameter comes to mind.

douglashall:
@tuchfarber It looks like you will want to convert the `edxapp_comprehensive_theme_dirs` param from a `choiceParam` to a `stringParam` and set the default value to `'["/edx/app/edxapp/edx-platform/themes","unset"]'`.

PR #200, opened 2017-07-21
jobDSL for status.edx.org job

hchen15:
@edx/devops

### DIFF
--- a/devops/jobs/StatusEdxOrg.groovy
+++ b/devops/jobs/StatusEdxOrg.groovy
@@ -38,7 +38,7 @@ class StatusEdxOrg {
                             'Git repo containing edX configuration.')
                 stringParam('CONFIGURATION_BRANCH', extraVars.get('CONFIGURATION_BRANCH', 'master'),
                         'e.g. tagname or origin/branchname')
-                stringParam('STATUS_REPO', extraVars.get('STATUS_REPO', 'git@github.com:edX/status.edx.org'),
+                stringParam('STATUS_REPO', extraVars.get('STATUS_REPO', 'git@github.com:edx/status.edx.org'),
                         'Git repo containing status.edx.org configuration.')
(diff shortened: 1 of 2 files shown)


devops/jobs/StatusEdxOrg.groovy
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            def gitCredentialId = extraVars.get('SECURE_GIT_CREDENTIALS','')
+
+            parameters{
+                stringParam('CONFIGURATION_REPO', extraVars.get('CONFIGURATION_REPO', 'https://github.com/edx/configuration.git'),
+                            'Git repo containing edX configuration.')
+                stringParam('CONFIGURATION_BRANCH', extraVars.get('CONFIGURATION_BRANCH', 'master'),
+                        'e.g. tagname or origin/branchname')
+                stringParam('STATUS_REPO', extraVars.get('STATUS_REPO', 'git@github.com:edX/status.edx.org'),
```
jibsheet: it'll worj because github fixes case, but can you s/edX/edx

devops/jobs/StatusEdxOrg.groovy
```diff
@@ -0,0 +1,97 @@
...
+            wrappers common_wrappers
+
+            wrappers{
+                credentialsBinding{
+                    file('AWS_CONFIG_FILE','tools-edx-jenkins-aws-credentials')
+                    string('ROLE_ARN', 'tools-jenkins-to-status-edx-org-role-arn')
+                }
+            }
+
+            def gitCredentialId = extraVars.get('SECURE_GIT_CREDENTIALS','')
```
jibsheet: Can we make a deploy key for this rather than using edx-secure?
https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys
You generate an ssh key, put it in lastpass, give github the public part for the edx.status.org repo (we can make sure you have access to do that) and then give Jenkins the private part as a folder level credential.

PR #202, opened 2017-07-26
Tuchfarber/fix sandbox theme error 2

tuchfarber:
https://openedx.atlassian.net/browse/WL-1104

Last fix didn't work because the types between this and ansible didn't
match. This should fix that. 

Previous conversations for reference:
Original: https://github.com/edx/jenkins-job-dsl/pull/194
Revert: https://github.com/edx/jenkins-job-dsl/pull/195

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -129,7 +129,6 @@ class CreateSandbox {
                 stringParam("configuration_secure_version","master","")
                 stringParam("configuration_internal_version","master","")
                 booleanParam("reconfigure",false,"Reconfigure and deploy, this will also run with --skip-tags deploy against all role <br />Leave this unchecked unless you know what you are doing")
-                choiceParam("edxapp_comprehensive_theme_dir",["/edx/app/edxapp/edx-platform/themes/edx.org","unset"],"")
                 booleanParam("testcourses",true,"")
                 booleanParam("performance_course",true,"")
                 booleanParam("demo_test_course",true,"")


devops/jobs/CreateSandbox.groovy
```diff
@@ -129,7 +129,7 @@ class CreateSandbox {
                 stringParam("configuration_secure_version","master","")
                 stringParam("configuration_internal_version","master","")
                 booleanParam("reconfigure",false,"Reconfigure and deploy, this will also run with --skip-tags deploy against all role <br />Leave this unchecked unless you know what you are doing")
-                choiceParam("edxapp_comprehensive_theme_dir",["/edx/app/edxapp/edx-platform/themes/edx.org","unset"],"")
+                stringParam("edxapp_comprehensive_theme_dirs",'[/edx/app/edxapp/edx-platform/themes/"]',"")
```
douglashall: Looks like you are missing a `"`.
```
stringParam("edxapp_comprehensive_theme_dirs",'["/edx/app/edxapp/edx-platform/themes/"]',"")
```

tuchfarber: I swear that was there before I pushed... Fixed now.

devops/jobs/CreateSandbox.groovy
```diff
@@ -129,7 +129,7 @@ class CreateSandbox {
                 stringParam("configuration_secure_version","master","")
                 stringParam("configuration_internal_version","master","")
                 booleanParam("reconfigure",false,"Reconfigure and deploy, this will also run with --skip-tags deploy against all role <br />Leave this unchecked unless you know what you are doing")
-                choiceParam("edxapp_comprehensive_theme_dir",["/edx/app/edxapp/edx-platform/themes/edx.org","unset"],"")
+                stringParam("edxapp_comprehensive_theme_dirs",'["/edx/app/edxapp/edx-platform/themes/"]',"")
```
andy-armstrong: Would it make sense to include edx-themes here by default too, so that folks can test with those themes?

douglashall: Sounds like a good idea to me!

tuchfarber:
@jibsheet Can you give this a lookover for me? I already broke sandboxes once, and don't particularly want to do it again.

andy-armstrong:
@tuchfarber what's the status of this work? I've had to build a number of sandboxes and manually configure themes on them so I've got fast at it, but it would be better if this could be fixed. In particular, I'd like to see this land with the usage of edx-themes that I suggested, so then every sandbox would automatically have all themes available for testing. Thanks!

jibsheet:
@tuchfarber If you can add in support for @andy-armstrong 's suggestion I can look at reviewing/merging/reseeding it

jibsheet:
So, the old behavior allowed you to choose between the edx theme or no theme (openedx theme).
This now hardcodes the list of directories in.  

Tell me if I'm wrong, but it feels like we should hardcode the list, but we should do it the way we do it for ecommerce (I'll send a pointer to sandbox later) but then let you put in the name of the theme you want instead.

@andy-armstrong Is your use case "make it go to edx.org out of the box", "let me choose any theme","let me add custom themes" or is it more mine "make the errors stop"

andy-armstrong:
@jibsheet great question. My personal needs are (a) to have edx.org be the default theme out of the box, and (b) to be able to test with all production themes by using ``/theming/admin`` to switch between them. I do both of these manually today, but this fix would allow me to do both on every sandbox without configuration. This is especially useful when I'm reviewing someone else's sandbox, as I don't want to have to go configure it for them.

It's an interesting point that there is no way to be able to see the default theme with this configuration. That seems like something that we should try to address.

andy-armstrong:
@jibsheet I made a PR for the change I think you were recommending against sandbox-internal: https://github.com/edx/jenkins-job-dsl/pull/202

jibsheet:
@andy-armstrong yes, that's where I was going.  Part of me wondered if we should expose EDXAPP_DEFAULT_SITE_THEME: "edx.org" in this job instead of exposing EDXAPP_COMPREHENSIVE_THEME_DIRS: ["/edx/var/edx-themes/edx-themes/edx-platform", "/edx/app/edxapp/edx-platform/themes"] which is something we're less likely to change, but I think the fix forward is:

- [ ] Merge  your PR (Feanil mentioned how to build a test sandbox it)
- [ ] Matt updates this PR to simply remove the invalid setting we were fixing
- [] Someone updates ansible-provision.sh to stop setting EDXAPP_COMPREHENSIVE_THEME_DIRS since it wasn't working anyway https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh#L190

tuchfarber:
Parameter has been removed on this PR

jibsheet:
@andy-armstrong Will you tweak the wiki instructions around theming?  I know they used to refer to some jenkins fields, one of which we just removed.  Having merged all the branches now, this resolves my major issue (errors on every request).

andy-armstrong:
I've updated the instructions here (https://openedx.atlassian.net/wiki/spaces/EdxOps/pages/13960183/Sandboxes#Sandboxes-Theming) and will send out an email letting folks now.

PR #203, opened 2017-07-27
jobDSL for tools-gp-users job

hchen15:
write the jobDSL for the tools-gp-users job.

update the existing app-permissions jobDSL to change the name of the file that the job is built in and to update docs


devops/jobs/ToolsGpUsers.groovy
```diff
@@ -0,0 +1,111 @@
...
+ Variables consumed for this job:
+    * SECURE_GIT_CREDENTIALS: secure-bot-user (required)
+    * NOTIFY_ON_FAILURE: alert@example.com
+    * FOLER_NAME: folder, default is User-Mananagement
+    * ACCESS_CONTROL: list of users to give access to
+        - user
+    * CONFIGURATION_SECURE_REPO: repository where ansible vars are located (required)
+ 
+ This job expects the following credentials to be defined on the folder
+    ubuntu_deployment_201407: ssh key
```
jibsheet: I know this is a pattern you're using elsewhere and I'm late, but can we rename this
It ties us to the particular deploy key
I'd rather just call is deploy_key since we might run this on a non-ubuntu host, or use a newer key

feanil: Can we pass in the expected name of this credentials via secure config as well.  At some point in the future we'll want to add the new credential and switch to it before deleting the old one and we'll want the name to reflect the content.

devops/jobs/ToolsGpUsers.groovy
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+
+            steps {
+                virtualenv {
+                    nature("shell")
+                    systemSitePackages(false)
+
+                    command(
+                        dslFactory.readFileFromWorkspace("devops/resources/tools-gp-users.sh")
```
jibsheet: Can this use https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/RunRemoteAnsiblePlaybook.groovy and save a bunch of new code, or is not quite right because of the single file we pass in?

hchen15: it's not exactly the same.. but very similar.

jibsheet: How bad would be to squeeze it into the other jobtype?  Maybe have to make one or two more things configurable?

PR #206, opened 2017-08-04
Estute/anonymous discover platform and e2e jobs

estute:
This will allow an unauthenticated user to follow a link from github, but will be redirected to a login page, hopefully reducing confusion.


estute:
Not very DRY, so I will be refactoring these soon

jibsheet:
Agree on the repetition

PR #207, opened 2017-08-07
Travis counts job

raeeschachar:


raeeschachar:
@jzoldak @estute @michaelyoungstrom 
First set of ticket, please review.

platform/jobs/edxTravisCounts.groovy
```diff
@@ -0,0 +1,57 @@
...
+// Travis Jobs
+
+// The travis-counts job
+Map prJob = [ name: 'travis-counts',
+              deprecated: false,
+              worker: 'coverage-worker',
+              branch: '*/master'
+              ]
+
+/* Iterate over the job configurations */
```
estute: I don't think we need to have any secrets (and therefore no assertions). This job doesn't contain any sensitive information.  You can write all the groovy code without any loop over secretMap/configs

platform/jobs/edxTravisCounts.groovy
```diff
@@ -0,0 +1,57 @@
...
+        logRotator {
+            daysToKeep(5)
+            numToKeep(201)
+        }
+        concurrentBuild(true)
+        label('coverage-worker')
+        environmentVariables {
+            env('SUBSET_JOB', jobConfig['subsetJob'])
+        }
+        scm {
```
estute: I was confused about why you had 'multijob', then I realized- the job on Jenkins is incorrect! There is only one git repo, so no need to use 'multijob/multiscm'. You can simply use git. See here for an example: https://github.com/edx/jenkins-job-dsl/blob/master/testeng/jobs/toggleSpigot.groovy#L65

platform/jobs/edxTravisCounts.groovy
```diff
@@ -0,0 +1,57 @@
...
+            multijob {
+                remote {
+                    url('https://github.com/edx/edx-e2e-tests.git')
```
estute: I think you mean testeng-ci here

platform/jobs/edxTravisCounts.groovy
```diff
@@ -5,53 +5,38 @@ Binding bindings = getBinding()
...
-                remote {
-                    url('https://github.com/edx/edx-e2e-tests.git')
-                    refspec(jobConfig.refspec)
-                    }
-                }
-                browser()
-                branch('\*/master')
-                    cleanBeforeCheckout()
-                }
+job(jobConfig['jobName']) {
```
estute: You can just put the name of the job here, since there are no secret values, or values we need to iterate over

platform/jobs/edxTravisCounts.groovy
```diff
@@ -5,53 +5,38 @@ Binding bindings = getBinding()
 config.putAll(bindings.getVariables())
 PrintStream out = config['out']
 
-// Travis Jobs
-
-// The travis-counts job
 Map prJob = [ name: 'travis-counts',
```
estute: We can delete this Map as well.

platform/jobs/edxTravisCounts.groovy
```diff
@@ -5,53 +5,38 @@ Binding bindings = getBinding()
...
-                }
+job(jobConfig['jobName']) {
+
+    logRotator {
+        daysToKeep(5)
+        numToKeep(201)
+    }
+    concurrentBuild(true)
+    label('coverage-worker')
+    environmentVariables {
```
estute: The "SUBSET_JOB' environment variable is only used when we run a jobs tasks on separate shads, as we do with python, lettuce and bok choy tests. We can get rid of this as well.

platform/jobs/edxTravisCounts.groovy
```diff
@@ -5,53 +5,38 @@ Binding bindings = getBinding()
...
         }
-        triggers {
-            periodic(10)
-        }
-
-        steps {
-           shell(readFileFromWorkspace('testeng/resources/travis-counts.sh'))
+    }
+    triggers {
+        periodic(10)
```
estute: I think you will want to use 'cron' for the trigger. See https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.triggers.TriggerContext.cron for more information

raeeschachar:
@estute 
Thank you for a detailed review. Done.

raeeschachar:
@estute 
Done and squashed for ease. Good to go.

estute:
^ just a note: we can do this in a later PR, but the code block concerning the printstream (lines 3-6) is unnecessary in this case (As mentioned in https://github.com/edx/jenkins-job-dsl/pull/212)

PR #209, opened 2017-08-07
This refactors the Docker CI jobs to use webhooks instead of polling.…

MichaelRoytman:
… It is not possible to correctly use webhooks with two repositories from a single job, so a change of structure was necessary. The changes are as follows:

1. DockerCI.groovy was removed.
2. ImageBuilder.groovy builds an application's Docker image, the name of which is passed in as a build parameter from an upstream job (e.g. app-watcher). It then pushes the Docker image up to Docker Hub.
3. AppWatcher.groovy watches a particular application's repository for pushes to master. Upon a push, this job triggers ImageBuilder, passing the name of the application as a parameter. Each application described in the job configuration file has an AppWatcher job. Those applications withoutapplication repositories serve only as interim layers between configuration-watcher and image-builder. That is, the configuration-watcher determines that application's configuration has been affected by a change and triggers the app-watcher for that application to trigger a downstream image-builder build and push.
4. ConfigurationWatcher.groovy watches the configuration repository for pushes to master. Upon a push, the changed files are piped into the parsefiles.py script, which resolves which Docker images need to be re-built as a result of the changes to the configuration repository. This list of images is then used to trigger the respective app-watcher job.


MichaelRoytman:
I'm going to create a graph that explains how these jobs work together, if it would be beneficial for the understanding of the structure.

Also, GitHub treated the AppWatcher job as a change to the DockerCI job, but for the sake of reviewing, I would just look at AppWatcher without the change log. I meant for DockerCI to be delete and AppWatcher to be a new job, but GitHub picked up similarities and treated it as a change.

MichaelRoytman:
@edx/devops Please review!

MichaelRoytman:
jenkins-job-dsl-internal PR: https://github.com/edx/jenkins-job-dsl-internal/pull/121

MichaelRoytman:
screen shot 2017-08-08 at 1 38 17 pm

Perhaps this is somewhat helpful in illustrating the job structure?

devops/jobs/AppWatcher.groovy
```diff
@@ -49,27 +57,22 @@ import static org.edx.jenkins.dsl.DevopsConstants.common_read_permissions
 import static org.edx.jenkins.dsl.Constants.common_logrotator
 import static org.edx.jenkins.dsl.Constants.common_wrappers
 
-class DockerCI {
+// is there a way not to create a job for which app_repo is undefined?
```
MichaelRoytman: Oops! I'll have to remove this.

devops/jobs/ConfigurationWatcher.groovy
```diff
@@ -0,0 +1,154 @@
...
+        dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/configuration-watcher") {
+
+            assert extraVars.containsKey('CONFIGURATION_REPO_URL') : "Please define CONFIGURATION_REPO_URL"
+            assert extraVars.containsKey('EDX_REPO_ROOT') : "Please define EDX_REPO_ROOT"
+            assert extraVars.containsKey('APPS_TO_CONFIG'): "Please define APPS_TO_CONFIG"
+
+            def config_branch = 'master'
+
+            // to space separated list as string
+            def apps = extraVars.get("APPS_TO_CONFIG").keySet().join(",").replaceAll(',', " ")
```
jibsheet: I assume join(" ") didn't work for some weird reason?

devops/jobs/AppWatcher.groovy
```diff
@@ -81,56 +84,55 @@ class DockerCI {
...
-                            relativeTargetDirectory('configuration')
-                            cleanAfterCheckout()
-                            // ignore notifications on commits to branch; this job will be triggered by
-                            // configuration-watcher job
-                            ignoreNotifyCommit()
-                        }
+                // for GitHub Webhooks
+                properties {
+                    if (app_repo) {
+                        githubProjectUrl(extraVars.get("EDX_REPO_ROOT") + app_repo + "/")
```
jibsheet: Ah, I think you're safe, because you're building from EDX_REPO_ROOT, ok, good

devops/jobs/AppWatcher.groovy
```diff
@@ -49,27 +57,22 @@ import static org.edx.jenkins.dsl.DevopsConstants.common_read_permissions
...
-                assert extraVars.containsKey('CONFIGURATION_REPO_URL') : "Please define CONFIGURATION_REPO_URL"
-                assert extraVars.containsKey('EDX_REPO_ROOT') : "Please define EDX_REPO_ROOT"
-                assert extraVars.containsKey('APPS_TO_CONFIG'): "Please define APPS_TO_CONFIG"
-
-                def app_repo = app_config.get('app_repo', '')
+                def app_repo = app_config.get("app_repo", '')
                 def app_repo_branch = app_config.get('app_repo_branch', 'master')
-                def config_branch = app_config.get('config_branch', 'master')
 
-                wrappers common_wrappers
```
jibsheet: I think we still want common_wrappers, right?

devops/jobs/ConfigurationWatcher.groovy
```diff
@@ -0,0 +1,154 @@
...
+            // for GitHub Webhooks
+            properties {
+                githubProjectUrl(extraVars.get("EDX_REPO_ROOT") +  "configuration/")
+            }
+
+            scm {
+                // check out edx/configuration repository from GitHub
+                git {
+                    remote {
+                        url(extraVars.get("CONFIGURATION_REPO_URL"))
```
jibsheet: It's a little strange we parameterize this, but hardcode "configuration/" above, but I'm not sure how better to do it

devops/jobs/ConfigurationWatcher.groovy
```diff
@@ -0,0 +1,154 @@
...
+            
+            // run the trigger-builds shell script in a virtual environment called venv
+            steps {
+                virtualenv {
+                    name('venv')
+                    nature('shell')
+                    command dslFactory.readFileFromWorkspace('devops/resources/trigger-builds.sh')
+                }
+
+                 // inject environment variables defined in the temp_props file (TO_BUILD)
```
jibsheet: Can you add more comments about where/how temp_props is generated?  I figured it out with grep, but would love a shortcut.

devops/jobs/ConfigurationWatcher.groovy
```diff
@@ -0,0 +1,154 @@
...
+                            stringsMatch('', '${TO_BUILD}', false)
+                        }
+                    }
+
+                    steps {
+
+                        // trigger the jobs defined in the TO_BUILD environment variable; this is set via the trigger-builds script
+                        // and injected into the environment from the temp_props file
+                        downstreamParameterized {
+                            trigger('${TO_BUILD}')
```
jibsheet: How does this work when more than 1 image file needs rebuilding? Does trigger parse a list that's showing up in temp_props file?

devops/jobs/ImageBuilder.groovy
```diff
@@ -0,0 +1,107 @@
+/*
+    This job serves to build an application's Docker image and to push it to Docker Hub. This job only serves as a downstream job to the 
+    various <app>-watcher jobs and does not run on its own. This job receives the name of an application as a build parameter. 
+    This application's Docker image is then built and pushed.
+
+    Upon a push to master, the job is triggered via a webhook, at which point it runs the parsefiles.py script with the files changed between
```
jibsheet: This doc is about the configuration job above, right?
This job is either called by ConfigurationWatcher or AppWatcher, right?

devops/jobs/ImageBuilder.groovy
```diff
@@ -0,0 +1,107 @@
...
+    Upon a push to master, the job is triggered via a webhook, at which point it runs the parsefiles.py script with the files changed between 
+    the old commit and the new commit. This resolves which Docker images need to be rebuilt as a result of the aforementioned changes to the 
+    configuration repository. For each Docker image that needs rebuilding, this job triggers the respective <app>-watcher job.
+
+    Variables consumed from the EXTRA_VARS input to your seed job in addition
+    to those listed in the seed job.
+    APPS_TO_CONFIG: a dictionary containing mappings from an edX IDA to a dictionary of various configuration values (REQUIRED). 
+        The structure of APPS_TO_CONFIG should be
+        
+        APPS_TO_CONFIG:
```
jibsheet: I think some of these docs are also more for AppWatcher, at least from reading the code

devops/jobs/ConfigurationWatcher.groovy
MichaelRoytman: Yes, essentially. The downstreamParameterized block takes a string of projects and triggers them. It's not super explicit about whether it can trigger multiple projects or how exactly it parses the string (whitespace?), but "projects" suggests that it does trigger multiple, and it does trigger all downstream jobs that are passed in through TO_BUILD.

Docs are [here](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.step.StepContext.downstreamParameterized), for reference.

devops/jobs/AppWatcher.groovy
MichaelRoytman: Oops. Yes, I'll add that back in.

devops/jobs/AppWatcher.groovy
MichaelRoytman: I will add a note about the https/slash restriction above the use of githubProjectUrl, as a heads up to those changing it in the future.

devops/jobs/ConfigurationWatcher.groovy
MichaelRoytman: That makes a lot more sense. I'm not entirely sure why I didn't try that, to be honest. I'll do that and try it out to make sure it works.

devops/jobs/ConfigurationWatcher.groovy
MichaelRoytman: Hm. We could remove the CONFIGURATION_REPO_URL variable and replace its use with extraVars.get("EDX_REPO_ROOT") +  "configuration.git"? Then it would be a bigger nuisance to modify it if the configuration repository moves in the future, though (although that seems unlikely to happen?). I could make CONFIGURATION_URL = "https://github.com/edx/configuration" and then append a "/" at line 88 and ".git" at line 95? Then we wouldn't need both variables. I don't know if either of those are ideal solutions, but I agree that it's strange to do it this way.

devops/jobs/ImageBuilder.groovy
MichaelRoytman: Yes. I'll fix up the docs.

devops/jobs/ConfigurationWatcher.groovy
jibsheet: Ah, yes, I remember this now.  I think this is probably ok in retrospect, I'm not sure how to make it clearer honestly.

devops/jobs/ConfigurationWatcher.groovy
jibsheet: You shouldn't need the .git on 95, github'll clone with or without just fine, but you would need the / at 88.

I think we should probably switch to using CONFIGURATION_REPO_URL in both places and documenting that it has to be https for the githubproject part use case.  That'll help anyone who uses this with an ssh repo later.

It's a little "pick your failure case" since either way can cause pain to change.  I'll leave final call to your judgement.

MichaelRoytman:
@edx/devops Please re-review. 

Related PRs: https://github.com/edx/edx-internal/pull/159, https://github.com/edx/jenkins-job-dsl-internal/pull/121

jibsheet:
@MichaelRoytman  Can we get the PR graphic you made into a wiki page somewhere?  I'm not sure if there's already a page under the docker space about how we build/push the images, but that seems like a logical place for it.

MichaelRoytman:
@jibsheet I don't see a page that seems completely appropriate for it on the wiki. There is [this](https://openedx.atlassian.net/wiki/spaces/EdxOps/pages/152797300/Docker) page; would that be a good place? Or should I make a new page under that heading?

devops/jobs/ConfigurationWatcher.groovy
```diff
@@ -0,0 +1,149 @@
...
+                        branch(config_branch)
+                    }
+                    extensions {
+                        relativeTargetDirectory('configuration')
+                    }
+                }
+                
+            }
+
+            triggers {
```
jibsheet: We do this twice, and will likely have the same trigger in the app permissions jobs
Does this want to be like common_wrappers and be merge_to_master_trigger defined in one of our jobdsl libraries?
Changing it here or adding it as an AC to our next OPS ticket seem fine to me.

MichaelRoytman: I can add that to [here](https://github.com/edx/jenkins-job-dsl-internal/blob/master/src/main/groovy/org/edx/jenkins/dsl/DevopsTasks.groovy), if that seems like the best place. I'm assuming that if I add the triggers { ... } context to one of the libraries, the config_branch variable would need to be defined in the job using the library, right? I figure I should probably change the name to be something a bit more generalizable than config_branch. I'm not positive how most of the existing jobs are structured, but do they generally pre-define the branch they check out of a git repository? I just want to make sure it's not confusing to use because of the variable I've got there.

jibsheet: Yeah, I would just pass an arg to the function that returns the closure, and naming is hard which is why I left it to you.
I can hunt around, I'm probably returning a closure somewhere else.

If you want, @hchen15  is going to fix App Permissions now, so we can just make her do the refactor (or you can work with her on the proper fix)

PR #214, opened 2017-08-14
dsl script for oep2 job

estute:


testeng/jobs/oep2Report.groovy
```diff
@@ -0,0 +1,56 @@
...
+            name('oep-venv')
+            nature('shell')
+            clear(true)
+            command(readFileFromWorkspace('testeng/resources/create-oep-report.sh'))
+        }
+    }
+    publishers {
+        archiveJunit('oep2-report.xml') {
+            retainLongStdout(false)
+            healthScaleFactor(1.0)
```
michaelyoungstrom: Trying to seed this gave me errors on test-jenkins as well as a local instance created from our jenkins-configuration repo. It looks like `healthScaleFactor()` becomes available in version 1.52 of the job-dsl plugin, and we are currently on 1.45

PR #217, opened 2017-08-15
add a snitch for retire instances job

hchen15:
switch from using email to snitches to alert for job failure

### DIFF
--- a/devops/jobs/MinosLifecycle.groovy
+++ b/devops/jobs/MinosLifecycle.groovy
@@ -10,7 +10,8 @@
     * DEPLOYMENTS: (required)
         deployment:
           environments:
-            - environment (required)
+            environment (required)
+                snitch: snitch url
     * SSH_ACCESS_CREDENTIALS: ssh access credentials, should be defined on the folder (required)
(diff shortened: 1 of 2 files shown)


devops/jobs/MinosLifecycle.groovy
```diff
@@ -87,6 +88,7 @@ class MinosLifecycle {
                         env('ENVIRONMENT', environment)
                         env('DEPLOYMENT', deployment)
                         env('AWS_REGION',extraVars.get('AWS_REGION','us-east-1'))
+                        env('SNITCH', inner_config.get('snitch', ''))
```
feanil: If this is empty, curl will fail in the bash script.  It seems like we should add an assert to ensure the snitch url is set?

jibsheet: I've normally done if (snitch) { shell('curl $snitch') } which achieves similar goals of not blowing up

jibsheet: (rabbit for example can run without the snitch)

feanil: I'm fine with making this an optional setting but then the bash script should be updated.

PR #219, opened 2017-08-15
rename application watcher jobs, update build names, and create a bui…

MichaelRoytman:
…lder job for each application instead of relying on a single builder job to improve clarity

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ build/*
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+*.DS_Store
--- a/devops/jobs/AppWatcher.groovy
+++ b/devops/jobs/AppWatcher.groovy
@@ -2,16 +2,15 @@
     This job handles changes to an application.
 
     For applications that have source code repositories, this job watches for changes pushed to master. 
-    Upon a push to master, the job is triggered via a webhook. The job triggers the downstream image-builder job, passing the name of the
-    application as a build parameter. The image-builder then builds and pushes the application's Docker image to DockerHub.
+    Upon a push to master, the job is triggered via a webhook. The job triggers the matching downstream image-builder job at
+    DockerCI/image-builders/<app>-image-builder. The <app>-image-builder then builds and pushes the application's Docker image to DockerHub.
 
-    For applications that do not have source code repositories, this job serves simply as an interim layer between the configuration-watcher
-    and image-builder, because they do not have an application code repository to watch. The configuration-watcher job triggers this job, 
-    and this job immediately triggers the image-builder downstream job, passing the name of the application as a build parameter. 
-    The image-builder then builds and pushes the application's Docker image to DockerHub.
+    For applications that do not have source code repositories, there are no watcher jobs, because there are no repositories to watch. If a new
+    image for these applications is needed, the associated <app>-image-builder job will need to be run manually.
 
     Variables consumed from the EXTRA_VARS input to your seed job in addition
     to those listed in the seed job.
+    
     APPS_TO_CONFIG: a dictionary containing mappings from an edX IDA to a dictionary of various configuration values (REQUIRED). 
         The structure of APPS_TO_CONFIG should be
         
@@ -61,39 +60,45 @@ class AppWatcher {
 
         // for each application
         extraVars.get("APPS_TO_CONFIG").each { app_name, app_config ->
-            dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/" + app_name) {
 
-                def app_repo = app_config.get("app_repo", '')
-                def app_repo_branch = app_config.get('app_repo_branch', 'master')
+            def app_repo = app_config.get("app_repo", '')
 
-                logRotator common_logrotator
+            // we don't need a watcher job if there is no repository to watch
+            if (app_repo) {
+                dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/application-watchers/" + app_name + "-watcher") {
 
-                wrappers common_wrappers
+                    // if an application has no repository, the app_repo_branch should be blank so as not to set up webhooks
+                    def app_repo_branch = !app_repo ? '' : app_config.get('app_repo_branch', 'master')
 
-                def access_control = extraVars.get('ACCESS_CONTROL',[])
-                access_control.each { acl ->
-                    common_read_permissions.each { perm ->
-                        authorization {
-                            permission(perm,acl)
+                    description('\rThis job watches the ' + app_repo_branch + ' branch of the ' + app_repo + ' repository for changes via a webhook. ' + 
+                        'Upon a change to the branch, this job will run and trigger the downstream ' + app_name + '-image-builder job, which will ' +
+                        'build and push the image to DockerHub. Applications without application code repositories do not have an associated watcher job.')
+
+                    logRotator common_logrotator
+
+                    wrappers common_wrappers
+
+                    def access_control = extraVars.get('ACCESS_CONTROL',[])
+                    access_control.each { acl ->
+                        common_read_permissions.each { perm ->
+                            authorization {
+                                permission(perm,acl)
+                            }
                         }
                     }
-                }
 
-                def APP_REPO_URL = extraVars.get("EDX_REPO_ROOT") + app_repo + "/"
-
-                // for GitHub Webhooks; note that you need to use an https URL and include
-                // the trailing slash
-                properties {
-                    if (app_repo) {
-                       githubProjectUrl(APP_REPO_URL)
+                    def app_repo_url = extraVars.get("EDX_REPO_ROOT") + app_repo + "/"
+                    
+                    // for GitHub Webhooks; note that you need to use an https URL and include
+                    // the trailing slash
+                    properties {
+                       githubProjectUrl(app_repo_url)   
                     }
-                }
 
-                scm {
-                    if (app_repo) {
+                    scm {
                         git {
                             remote {
-                                url(APP_REPO_URL)
+                                url(app_repo_url)
                                 branch(app_repo_branch)
                             }
                             extensions {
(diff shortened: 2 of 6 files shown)


devops/jobs/AppWatcher.groovy
```diff
@@ -61,10 +61,11 @@ class AppWatcher {
 
         // for each application
         extraVars.get("APPS_TO_CONFIG").each { app_name, app_config ->
-            dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/" + app_name) {
+            dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/application-watchers/" + app_name + "-watcher") {
```
MichaelRoytman: This will require an update to the internal documentation to include creation of this folder: https://github.com/edx/jenkins-job-dsl-internal/pull/126.

devops/resources/trigger-builds.sh
```diff
@@ -25,4 +25,4 @@ export TRAVIS_BUILD_DIR=${WORKSPACE}/configuration
 # as necessary to be built via the parsefiles.py script and redirect it into temp_props file to be read in as an
 # environment variable TO_BUILD
 cd configuration
-echo "TO_BUILD=$(comm -12 <(echo ${APPS} | tr " " "\n" | sort) <(git diff --name-only ${GIT_PREVIOUS_COMMIT}...${GIT_COMMIT} | python util/parsefiles.py | tr " " "\n" | sort) | while read play; do echo -n "$play, "; done)" > ../temp_props
+echo "TO_BUILD=$(comm -12 <(echo ${APPS} | tr " " "\n" | sort) <(git diff --name-only ${GIT_PREVIOUS_COMMIT}...${GIT_COMMIT} | python util/parsefiles.py | tr " " "\n" | sort) | while read play; do echo -n "image-builders/$play-image-builder, "; done)" > ../temp_props
```
MichaelRoytman: I'm just formatting each image name to correspond with a path to the corresponding image builder job in Jenkins.

devops/jobs/ImageBuilder.groovy
```diff
@@ -26,54 +26,61 @@ import static org.edx.jenkins.dsl.Constants.common_wrappers
...
     public static def job = { dslFactory, extraVars ->
-        dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/image-builder") {
+        //assertions come earlier to ensure APPS_TO_CONFIG can be read
+        assert extraVars.containsKey('APPS_TO_CONFIG'): "Please define APPS_TO_CONFIG"
+        assert extraVars.containsKey('CONFIGURATION_REPO_URL') : "Please define CONFIGURATION_REPO_URL"
 
-            assert extraVars.containsKey('CONFIGURATION_REPO_URL') : "Please define CONFIGURATION_REPO_URL"
+        // for each application
+        extraVars.get("APPS_TO_CONFIG").each { app_name, app_config ->
+            dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/image-builders/" + app_name + "-image-builder") {
```
MichaelRoytman: This will require an update to the internal documentation to include creation of this folder: edx/jenkins-job-dsl-internal#126.

MichaelRoytman:
@edx/devops Please review yet another change to the CI jobs!

devops/jobs/ImageBuilder.groovy
```diff
@@ -26,54 +26,61 @@ import static org.edx.jenkins.dsl.Constants.common_wrappers
...
-                credentialsBinding {
-                    file("CONFIG_JSON_FILE", extraVars.get("CONFIG_JSON_FILE_CREDENTIAL_ID", "docker-config-json"))
+                wrappers common_wrappers
+                // add credentials to log in to DockerHub; ID refers to credential ID as set up in Jenkins
+                wrappers {
+                    credentialsBinding {
+                        file("CONFIG_JSON_FILE", extraVars.get("CONFIG_JSON_FILE_CREDENTIAL_ID", "docker-config-json"))
+                    }
+                    // add master commits of application repository and configuration repository that are used to build the image
+                    buildName('#${BUILD_NUMBER} with app. version = ${ENV:0:7, var="APP_VERSION"} and config. version = ${GIT_REVISION, length=8}')
```
MichaelRoytman: I know this looks funny. length is a parameter of the GIT_REVISION macro but not of the ENV macro, so I had to use the substring equivalent. I could change GIT_REVISION to match if it's confusing this way.

MichaelRoytman: Also, I know this might not be as useful since each application will have their own builder job now, but I thought it might still be helpful to know which commits were checked out when building the image.

devops/jobs/AppWatcher.groovy
```diff
@@ -61,10 +61,11 @@ class AppWatcher {
...
 
                 def app_repo = app_config.get("app_repo", '')
-                def app_repo_branch = app_config.get('app_repo_branch', 'master')
+                // if an application has no repository, the app_repo_branch should be blank so as not to set up webhooks
+                def app_repo_branch = !app_repo ? '' : app_config.get('app_repo_branch', 'master')
```
jibsheet: Do we have anyone with an empty repo?  If so, we probably need to support polling, otherwise, manual and documented Works For Me

MichaelRoytman: @jibsheet Yeah, I realized polling would be pointless if there is no repository to poll... I don't think there are empty repos, so I'll remove that and just support webhooks or manual.

.gitignore
```diff
@@ -16,3 +16,4 @@ build/*
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+*.DS_Store
```
MichaelRoytman: This was annoying me, so I plopped this in here. Let me know if this should be in a separate PR.

devops/jobs/AppWatcher.groovy
```diff
@@ -61,56 +59,63 @@ class AppWatcher {
...
         extraVars.get("APPS_TO_CONFIG").each { app_name, app_config ->
-            dslFactory.job(extraVars.get("FOLDER_NAME", "DockerCI") + "/" + app_name) {
 
-                def app_repo = app_config.get("app_repo", '')
-                def app_repo_branch = app_config.get('app_repo_branch', 'master')
+            def app_repo = app_config.get("app_repo", '')
 
-                logRotator common_logrotator
+            // we don't need a watcher job if there is no repository to watch
+            if (app_repo) {
```
MichaelRoytman: I don't want to create watcher jobs for applications without application code repositories (e.g. firefox, chrome, rabbitmq, xenial-common).

devops/jobs/AppWatcher.groovy
```diff
@@ -61,56 +59,63 @@ class AppWatcher {
...
-                   downstreamParameterized {
-                        trigger('image-builder') {
-                            parameters {
-                                predefinedProp('APP_NAME', app_name)
+                    steps {
+                        // trigger image-builder job, passing name of application as a parameter
+                       downstreamParameterized {
+                            trigger('../image-builders/' + app_name + '-image-builder') {
+                                parameters {
+                                    predefinedProp('APP_VERSION', '${GIT_COMMIT}')
```
MichaelRoytman: Since each application has their own builder job, it's no longer necessary to pass through the application name, since the builder job has access to that. Now we pass in the version of the application checked out by the watcher job to the builder job so it can be used in the build name.

devops/jobs/ImageBuilder.groovy
```diff
@@ -26,54 +26,76 @@ import static org.edx.jenkins.dsl.Constants.common_wrappers
...
-            }
+                wrappers common_wrappers
+                // add credentials to log in to DockerHub; ID refers to credential ID as set up in Jenkins
+                wrappers {
+                    credentialsBinding {
+                        file("CONFIG_JSON_FILE", extraVars.get("CONFIG_JSON_FILE_CREDENTIAL_ID", "docker-config-json"))
+                    }
+                    // if the application has a repository, the build name will display the version of the application used; otherwise, display
+                    // only the configuration version
+                    if (app_config.get("app_repo")) {
```
MichaelRoytman: It doesn't make sense to have "app. version = " for applications without application code repositories, so we have different buildName structures depending upon whether the application has a code repository or not.

devops/jobs/AppWatcher.groovy
```diff
@@ -61,56 +59,67 @@ class AppWatcher {
...
 
-                def access_control = extraVars.get('ACCESS_CONTROL',[])
-                access_control.each { acl ->
-                    common_read_permissions.each { perm ->
-                        authorization {
-                            permission(perm,acl)
+                    // if an application has no repository, the app_repo_branch should be blank so as not to set up webhooks
+                    def app_repo_branch = !app_repo ? '' : app_config.get('app_repo_branch', 'master')
+
+                    description('\rThis job watches the ' + app_repo_branch + ' branch of the ' + app_repo + ' repository for changes via a webhook. ' +
```
MichaelRoytman: I added a carriage return because it looked funky in Jenkins otherwise.

devops/jobs/ImageBuilder.groovy
```diff
@@ -26,54 +26,87 @@ import static org.edx.jenkins.dsl.Constants.common_wrappers
...
+                }
+                logRotator common_logrotator
+
+                // add APP_VERSION as parameter; will be passed in from upstream job as version of application being built (for build name)
+                // applications without code repositories do not have this parameter to avoid confusion when performing a manual build
+                // the default value is a string of 8 spaces; this avoids the issue of someone triggering this build manually without supplying 
+                // a value for APP_VERSION, because if the value is less than 8 spaces, the buildName is not displayed because of the substring
+                if (app_config.get("app_repo")) {
+                    parameters {
+                        stringParam('APP_VERSION', '        ',
```
MichaelRoytman: The default for APP_VERSION is eight spaces, because if a user builds the job and either leaves the APP_VERSION parameter empty or enters a string with fewer than eight characters (they shouldn't), Jenkins is unable to display the buildName, because it wants to display the APP_VERSION with eight characters and is unable to do so. I've added this default to attempt to avoid that happening as much as possible.

src/main/groovy/org/edx/jenkins/dsl/DevopsConstants.groovy
```diff
@@ -93,11 +93,6 @@ class DevopsConstants {
             if (branchName == 'master') {
                 githubPush()
             }
-            // for all other jobs in this style, poll github for new commits on
-            // the 'defaultBranch' every 10 minutes
-            else {
-                scm('H/10 * * * *')
-            }
```
MichaelRoytman: We don't want to support polling. Jobs that can have webhooks (i.e. applications with application code repositories) have webhooks. Others (e.g. firefox, chrome, rabbitmq, xenial-common) will either be triggered by configuration-watcher or manually.

MichaelRoytman:
@edx/devops This is ready for re-review. I tried to point out the changes I've made since the last review with comments. Please and thank you!

devops/jobs/ImageBuilder.groovy
jibsheet: Now that we have ecommerce-image-builder is naming the run as important?  You can just click in and see the versions on the build page.
If that helps remove a weird workaround, I'm for removing the part that names the run

PR #221, opened 2017-08-22
upgrade gradle to output codenarc violations in travis logs

estute:
An html report of codenarc violations is not very useful on travis, so this will output the failures to stdout. It requires a substantial bump to the version of gradle used, but we are already using this in https://github.com/edx/jenkins-configuration/blob/master/build.gradle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,5 @@ script:
     - pep8 mobile_app/resources
     - nosetests mobile_app/resources -v --with-coverage --cover-package=mobile_app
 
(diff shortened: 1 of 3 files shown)


estute:
@jibsheet no. I reviewed the change logs and it would definitely change a lot if we were using this to build a java/groovy project. But Tools is only using gradle for some dependency management and calling codenarc. I assume that is the extent of what you guys are doing as well

jibsheet:
@estute  Your characterization of our usage is correct.  We use it to run tests and to manage deps.

PR #226, opened 2017-09-08
fix variable name for seed job

syed-awais-ali:
@edx/devops Review please. variable name changed.


feanil:
@syed-awais-ali the error is unrelated to your changes, I think you're safe to merge this.  If that changes master, we'll take a closer look at it here.

PR #233, opened 2017-09-13
install required package for job

syed-awais-ali:
@edx/devops Review please. Added code to install requirments


devops/resources/cloudflare-hit-rate.sh
```diff
@@ -2,6 +2,7 @@
 set -exuo pipefail
 
 cd $WORKSPACE/configuration
+pip install -r requirements-cloudflare.txt
```
feanil: update based or requested change in configuration PR: Requirement file should be under `util/jenkins`

PR #234, opened 2017-09-20
added var for terminating sandbox

syed-awais-ali:
@edx/devops Review please. Ticket [OPS-824](https://openedx.atlassian.net/browse/OPS-824).

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -202,6 +202,8 @@ class CreateSandbox {
 
                 stringParam("security_group","sandbox-vpc","")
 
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -202,6 +202,8 @@ class CreateSandbox {
 
                 stringParam("security_group","sandbox-vpc","")
 
+                stringParam("instance_initiated_shutdown_behavior","terminate","")
```
jibsheet: Do we even want this to be a user parameter?  I can't think of a reason why @edx/devops can you come up with one?
I'd either just code it into ansible-provision or send it as an Environment from here unless someone thinks of a reason to let folks override the behavior.

PR #248, opened 2017-10-23
refactor python pr job

estute:
The platform jenkins jobs stored a lot of configuration as secrets, which required modifying secrets in order to change the job. Since we have moved to a programmatic Jenkins process, this has become unmanageable for basic job changes. Therefore I am refactoring the python job (planning on doing the others afterwards).

### DIFF
--- a/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
+++ b/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
@@ -1,30 +1,10 @@
-package devops
+package platform
 
 import org.yaml.snakeyaml.Yaml
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
-import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_GITHUB_BASEURL
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_JUNIT_REPORTS
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.GHPRB_WHITELIST_BRANCH
-
-/*
-Example secret YAML file used by this script
-publicJobConfig:
-    open : true/false
-    jobName : name-of-jenkins-job-to-be
-    subsetJob : name-of-test-subset-job
-    repoName : name-of-github-edx-repo
-    coverageJob : name-of-unit-coverage-job
-    testengUrl: testeng-github-url-segment
-    platformUrl : platform-github-url-segment
-    testengCredential : n/a
-    platformCredential : n/a
-    platformCloneReference : clone/.git
-    workerLabel: worker-label
-    whitelistBranchRegex: 'release/*'
-    context: jenkins/test
-    triggerPhrase: 'jenkins run test'
-    defaultTestengBranch: 'master'
-*/
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.GENERAL_PRIVATE_JOB_SECURITY
 
 /* stdout logger */
 Map config = [:]
@@ -33,14 +13,11 @@ config.putAll(bindings.getVariables())
 PrintStream out = config['out']
 
 /* Map to hold the k:v pairs parsed from the secret file */
-Map secretMap = [:]
 Map ghprbMap = [:]
 try {
     out.println('Parsing secret YAML file')
-    String secretFileContents = new File("${EDX_PLATFORM_TEST_PYTHON_PR_SECRET}").text
     String ghprbConfigContents = new File("${GHPRB_SECRET}").text
     Yaml yaml = new Yaml()
-    secretMap = yaml.load(secretFileContents)
     ghprbMap = yaml.load(ghprbConfigContents)
     out.println('Successfully parsed secret YAML file')
 }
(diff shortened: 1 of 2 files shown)


platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -49,63 +26,130 @@ catch (any) {
...
-    buildFlowJob(jobConfig['jobName']) {
-        if (!jobConfig['open'].toBoolean()) {
-            authorization {
-                blocksInheritance(true)
-                permissionAll('edx')
-                permission('hudson.model.Item.Discover', 'anonymous')
-            }
+
+// This script generates a lot of jobs. Here is the breakdown of the configuration options:
+// Map exampleConfig = [ open: true/false if this job should be 'open' (use the default security scheme or not
```
jzoldak: typo: missing )

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -49,63 +26,130 @@ catch (any) {
...
+//                       jobName: name of the job
+//                       subsetjob: name of subset job run by this job (shard jobs)
+//                       repoName: name of the github repo containing the edx-platform you want to test
+//                       coveragejob: name of the coverage job to run after the unit tests
+//                       workerlabel: label of the worker to run the subset jobs on
+//                       whitelistbranchregex: regex of git branches to build
```
jzoldak: I'm actually not sure myself how this works. Maybe we can improve this help text. Is it that PRs with this branch as the merge target (for the PR jobs) and merges to this branch (for the master jobs) will trigger the job?

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -49,63 +26,130 @@ catch (any) {
...
         parameters {
-            stringParam('WORKER_LABEL', jobConfig['workerLabel'], 'Jenkins worker for running the test subset jobs')
+            stringParam('WORKER_LABEL', jobConfig.workerLabel, 'Jenkins worker for running the test subset jobs')
         }
         multiscm {
             git {
                 remote {
-                    url(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['testengUrl'] + '.git')
-                    if (!jobConfig['open'].toBoolean()) {
-                        credentials(jobConfig['testengCredential'])
```
jzoldak: doublechecking my understanding of this change - is it that jenkins workers don't need credentials for cloning this repo because it's open?

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -114,18 +158,18 @@ secretMap.each { jobConfigs ->
...
                     }
                 }
                 branch('\${ghprbActualCommit}')
                 browser()
                 extensions {
-                    relativeTargetDirectory(jobConfig['repoName'])
+                    relativeTargetDirectory(jobConfig.repoName)
                     cloneOptions {
-                        reference("\$HOME/" + jobConfig['platformCloneReference'])
+                        reference("\$HOME/${jobConfig.repoName}/.git")
```
jzoldak: Is the "/" needed before .git?

src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
```diff
@@ -139,7 +139,14 @@ class JenkinsPublicConstants {
             permission('hudson.model.Item.Cancel', 'edx')
         }
 
+    }
 
+    public static final Closure GENERAL_PRIVATE_JOB_SECURITY = {
```
jzoldak: Is this so that other job definitions can use this snippet also? Nice.

estute: Right. Trying to make these more DRY as I go

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
estute: In the past, we had considered creating a private version of testeng-ci, but the reasons to do so never fully materialized

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
estute: tbh not sure, but I am just configuring it the same as it is on the current jobs

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
estute: how is this?

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
jzoldak: For background, the purpose of the shallow clone is so that the checkout of the edx-platform repo at the SHA to be tested will not start with an initial clone, but instead start with objects at a point from which the changeset will be smaller and faster.

* I just ssh'd into a jenkins_worker and looked at /home/jenkins
  * /home/jenkins/edx-platform-clone and it looks like a clone from Jul 26 (which I'm guessing is when the current AMI that we are using was created.). 
* Seems like the right way to fix this line is to use $HOME/edx-platform-clone/.git. 
  * Note that this should also be true for the private job case, UNLIKE HOW IT IS CONFIGURED NOW.
  * Searching in the jenkins secrets in the secure repo, $HOME/edx-platform-private-clone/.git is being used by the private jobs now. This doesn't make sense, as that directory does not exist on the jenkins worker, and so using that directory as a reference won't save anybody anything.
* FWIW I'm not sure why it's $HOME/edx-platform-clone/.git rather than $HOME/edx-platform-clone. From my testing, they work equivalently. However the latter seems more straightforward, and is supported by the example in the git-clone documentation. :woman_shrugging:

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -114,18 +159,20 @@ secretMap.each { jobConfigs ->
             }
            git {
                 remote {
-                    url(JENKINS_PUBLIC_GITHUB_BASEURL + jobConfig['platformUrl'] + '.git')
+                    url("https://github.com/edx/${jobConfig.repoName}.git")
                     refspec('+refs/pull/*:refs/remotes/origin/pr/*')
                     if (!jobConfig['open'].toBoolean()) {
-                        credentials(jobConfig['platformCredential'])
+                        credentials("${GITHUB_BOT_CREDENTIALS}")
```
jzoldak: Let's double check the name of this variable "GITHUB_BOT_CREDENTIALS"

platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -1,30 +1,10 @@
-package devops
+package platform
```
jzoldak: oh nice catch!

PR #251, opened 2017-10-26
add configs for django upgrade python test jobs

estute:
This will create jobs for running the python unittest against django 1.11. They will not be triggered when a user creates or updates a PR, only when the user comments 'jenkins run django upgrade python'. The actual tests will be run on 'django-upgrade-workers' which are identical to jenkins workers but tagged differently, for cost-reporting.

I have a couple of questions:
* is the testing phrase ok? Unfortunately we can't do something like 'jenkins run python 1.11' because it will also match 'jenkins run python'.
* do we want to also create jobs for ginkgo? 
* do we need to consider anything in how coverage is run? currently a separate job runs 'paver coverage'. I tend to think we don't need to do anything here, but perhaps there is something I am overlooking?

For more information on how this all fits together see:
* https://github.com/edx/jenkins-job-dsl/pull/248
* https://github.com/edx/testeng-ci/pull/110
* https://github.com/edx/edx-platform/pull/16308

### DIFF
--- a/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
+++ b/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
@@ -54,6 +54,20 @@ Map publicJobConfig = [ open: true,
                         defaultTestengBranch: 'master'
                         ]
 
+Map djangoUpgradeJobConfig = [ open: true,
+                               jobName: 'edx-platform-django-upgrade-unittests-pr',
+                               subsetJob: 'edx-platform-test-subset',
+                               repoName: 'edx-platform',
+                               coverageJob: 'edx-platform-unit-coverage',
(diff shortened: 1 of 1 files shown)


estute:
@doctoryes I have spun up a test version of our production jenkins and used it to seed the DSL scripts from this PR. Then, I can collect the webhooks that jenkins sends when I add comments to a pull request, and curl them at my test jenins. Right now, that is the closest we can get to imitating a real scenario without interfering with production

estute:
For examples from testing, see:
https://test-jenkins.testeng.edx.org/job/edx-platform-python-unittests-pr/6/
https://test-jenkins.testeng.edx.org/job/edx-platform-django-upgrade-unittests-pr/2/

PR #255, opened 2017-11-08
Fix authentication issue in mongo prune script adding db name in connection string

nadeemshahzad:


nadeemshahzad:
@edx/devops Please review

PR #256, opened 2017-11-09
Estute/platform pr jobs need less secrets

estute:
Following the pattern in https://github.com/edx/jenkins-job-dsl/pull/248, this continues to refactor the platform pr jobs to contain most of their configurations in the DSL itself. It makes changing and reseeding jobs easier, as we won't need to reconfigure jenkins credentials as often.

Also, I have added a generic spock test for all of the pr-based jobs. It simply verifies the number of jobs created for each dsl and that no exceptions were thrown during the seeding process


estute:
@michaelyoungstrom good call. Runnning https://github.com/edx/edx-platform/pull/16544 over on https://test-jenkins.testeng.edx.org now.

PR #261, opened 2017-11-20
Update spigot hipchat notification

michaelyoungstrom:
Let's move the spigot notifier to Test Engineering and explain what it means so we can be more transparent with the org.

### DIFF
--- a/testeng/jobs/toggleSpigot.groovy
+++ b/testeng/jobs/toggleSpigot.groovy
@@ -12,7 +12,7 @@ PrintStream out = config['out']
 
 try {
     out.println('Parsing secret YAML file')
-    String constantsConfig = new File("${EDX_TOGGLE_SPIGOT_SECRET}").text
+    String constantsConfig = new File("${TOGGLE_SPIGOT_SECRET}").text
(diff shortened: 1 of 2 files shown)


testeng/jobs/toggleSpigot.groovy
```diff
@@ -109,9 +110,9 @@ secretMap.each { jobConfigs ->
             mailer(jobConfig['email'])
             hipChat {
                 token(jobConfig['hipchat'])
-                rooms('Tools Super Team')
+                rooms('Test Engineering')
```
estute: I think that you will need to create a new hipchat token. If I recall correctly, the token it is using is only for 'tools super team'. Not sure

testeng/jobs/toggleSpigot.groovy
```diff
@@ -95,6 +95,7 @@ secretMap.each { jobConfigs ->
 
         environmentVariables {
             env('AWS_DEFAULT_REGION', jobConfig['region'])
+            groovy(readFileFromWorkspace('testeng/resources/toggle-spigot-message.groovy'))
```
estute: Nice

PR #266, opened 2017-11-28
Notify @here when spigot state changes

michaelyoungstrom:


jzoldak:
Polling for comments in the TestEngineering chat room itself. Let's see what the people say.

nasthagiri:
+1 to the `@here`

PR #276, opened 2017-12-12
change names of django jobs

estute:
This will prevent the django jobs from overwriting eachother, which wasn't detected by the tests


estute:
I have decided to keep the django-1.11 jobs named 'edx-platform-django-upgrade-unittests-pr' as they have already been seeded and run on Jenkins. The other django jobs (i.e 1.9 and 1.0) are just temporary jobs

PR #277, opened 2017-12-13
Keep trigger phrase for django 1.11 consistent

michaelyoungstrom:
Let's not confuse the org by switching up the existing job name/trigger phrase


estute:
Let's just make sure that we communicate this to the platform team (as they are the one who will be using this trigger phrase and the 'jenkins run django 19/110' phrases)

michaelyoungstrom:
FYI @edx/platform-team

PR #279, opened 2017-12-14
Harden the curl statement used to update github status

jzoldak:


platform/resources/github_status_shell.sh
```diff
@@ -4,27 +4,10 @@ set -e
 echo "POSTing the request for GitHub to create a new status."
 echo "Response from GitHub:"
 curl \
--s -S --retry 3 \
+--connect-timeout 3.05 \
```
jzoldak: Note to reviewers:
The 3.05 follows [the suggestion in the Requests library](http://docs.python-requests.org/en/master/user/advanced/#timeouts): "It's a good practice to set connect timeouts to slightly larger than a multiple of 3, which is the default TCP packet retransmission window."

PR #283, opened 2017-12-29
use db cache creds in subet job

estute:
Related to https://github.com/edx/edx-platform/pull/17062. This will allow the jenkins job that runs the tests to access the s3 bucket where the cache files are stored


estute:
cc @jzoldak @michaelyoungstrom

PR #289, opened 2018-01-16
disable coverage for django upgrade jobs

estute:
[TE-2385]
This pull request is part of a multi-pr change to disable both running unit tests through coverage and the coverage job for the django 1.11 upgrade work.

This pull request adds a value for whether or not coverage should be run, and passes it to the build flow dsl script. There is another pull request for that work here: https://github.com/edx/testeng-ci/pull/129

### DIFF
--- a/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
+++ b/platform/jobs/edxPlatformPythonUnitTestsPr.groovy
@@ -32,6 +32,8 @@ catch (any) {
 //                       jobName: name of the job
 //                       subsetjob: name of subset job run by this job (shard jobs)
 //                       repoName: name of the github repo containing the edx-platform you want to test
+//                       runCoverage: whether or not the shards should run unit tests through coverage, and then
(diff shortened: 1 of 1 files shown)


platform/jobs/edxPlatformPythonUnitTestsPr.groovy
```diff
@@ -274,17 +286,20 @@ jobConfigs.each { jobConfig ->
...
                }
-           }
-           configure { node ->
-               node / publishers << 'jenkins.plugins.shiningpanda.publishers.CoveragePublisher' {
+               configure { node ->
+                   node / publishers << 'jenkins.plugins.shiningpanda.publishers.CoveragePublisher' {
+                   }
                }
-           }
+            }
```
jzoldak: nit: does this have an extra space in there now?

estute: indeed. Cleaned it up

PR #292, opened 2018-01-19
Refactor the quality job into a sharded flow TE-2396

jzoldak:


platform/jobs/edxPlatformQualityPr.groovy
```diff
@@ -166,43 +139,40 @@ jobConfigs.each { jobConfig ->
...
-
-        configure GHPRB_WHITELIST_BRANCH(jobConfig.whitelistBranchRegex)
-
         wrappers {
             timeout {
                 absolute(90)
             }
             timestamps()
             colorizeOutput()
-            buildName('#\${BUILD_NUMBER}: Quality Tests')
```
estute: Until we do https://openedx.atlassian.net/browse/TE-2333, I think we will need to keep the sshAgent part for private jobs

platform/jobs/edxPlatformQualityPr.groovy
```diff
@@ -30,121 +29,95 @@ catch (any) {
...
+    context: 'jenkins/quality',
+    triggerPhrase: 'jenkins run quality',
+    defaultTestengBranch: 'master',
+    diffJob: 'edx-platform-quality-diff'
+]
+
+Map privateJobConfig = [
+    open: false,
+    jobName: 'edx-platform-quality-flow-pr_private',
+    repoName: 'edx-platform-private',
```
estute: I think you will need to specify a subset worker for the private job

PR #293, opened 2018-01-22
Add django upgrade testing to master python unit testing

jzoldak:
Also refactor into the new layout.


jzoldak:
Needs this change in testeng-ci for the django/coverage options to be wired through. https://github.com/edx/testeng-ci/pull/131

PR #294, opened 2018-01-23
turn on django 1.11 unit tests for all prs

estute:
All pull requests to the edx-platform will now run both unittests and django upgrade (1.11) unittests. This pr does not touch the existing django 1.9 and 1.10 jobs. I will remove them in a separate pull request.

This job should be reseeded on Jenkins once the platform team has confirmed that all unit tests are passing for both normal and django 1.11 unit test jobs


estute:
@jzoldak FYI we are unable to run GHPRB jobs w/o posting a context back to Github. The plugin does not support this, unfortunately

PR #295, opened 2018-01-23
remove django 1.9 and 1.10 unit test jobs

estute:
If we are ready to run the django 1.11 jobs on all pull requests, I assume we no longer need the intermediate django 1.9 and 1.10 jobs. Can someone confirm?


jmbowman:
I'm still running these to make sure we've captured and fixed all the relevant deprecation warnings along the way to the 1.11 release.  I'm wrapping up that work in the next day or two, but it would be nice to retain the ability to run these until we've actually deployed the 1.11 upgrade.

estute:
Sure thing. Feel free to update this pr when you are ready for us to remove them

jmbowman:
@estute These were already removed, right?

estute:
Yes, these are gone. Closing.

PR #302, opened 2018-02-01
Create a job to toggle cron jobs

michaelyoungstrom:
This could be a convenient way to shut off all cron jobs, either when we're using a test/stage jenkins or if we're having a planned outage and dont want any jobs to run.

It uses the same secret as toggle-spigot, since it really only needs an email and the tools team.

### DIFF
--- a/testeng/jobs/toggleCronJobs.groovy
+++ b/testeng/jobs/toggleCronJobs.groovy
@@ -1,0 +1,72 @@
+package testeng
+
+import hudson.util.Secret
+import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_MASKED_PASSWORD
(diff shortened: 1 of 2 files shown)


testeng/jobs/toggleCronJobs.groovy
```diff
@@ -0,0 +1,72 @@
...
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+Map config = [:]
+Binding bindings = getBinding()
+config.putAll(bindings.getVariables())
+PrintStream out = config['out']
+
+try {
+    out.println('Parsing secret YAML file')
+    String constantsConfig = new File("${TOGGLE_SPIGOT_SECRET}").text
```
jzoldak: nit: I totally get what you're saying about it using the same secret as toggle-spigot, but still a bit afraid that this might cause confusion down the line if we ever remove or rename the toggle-spigot job. 
We could instead refactor the name TOGGLE_SPIGOT_SECRET. Admittedly this is kind of a pain as it touches a couple repos.

PR #303, opened 2018-02-01
Refactor platform master jobs to remove unneccessary secrets

michaelyoungstrom:
We've done this for the pr jobs and python master... let's finish it off.


platform/jobs/edxPlatformPythonUnitTestsMaster.groovy
```diff
@@ -16,21 +16,6 @@ Binding bindings = getBinding()
 config.putAll(bindings.getVariables())
 PrintStream out = config['out']
 
-/* Map to hold the k:v pairs parsed from the secret file */
-Map ghprbMap = [:]
```
jzoldak: good catch!

PR #306, opened 2018-02-06
Add in TEST_SUITE env name in a11y dsl

michaelyoungstrom:
To make paver timing log names unique we are now going to use the TEST_SUITE env var. a11y jobs are the only type without this variable (since they call a specific script rather than generic-ci), but let's add it in for this use case.


michaelyoungstrom:
@estute Great point, I agree. I'll add a ticket in the backlog.

PR #307, opened 2018-02-07
Add a test notifier job

michaelyoungstrom:
Adds a job that comments on edx-platform PR's to inform developers that their tests are complete. This will help prevent developers from staring at their PR until they see a green check, and will avoid any lost time between when their tests actually finish and when they realize they are done.


michaelyoungstrom:
Will wait to merge this until we are ready to run the "experiment."

PR #310, opened 2018-02-09
fail quality fast

estute:
This is a temporary fix while we roll out the new sharded quality pr job. To make sure that people with existing pull requests that do not yet contain the fix in edx/edx-platform#17252 do not pass quality without actually running anything, this will fail fast on the quality flow job, alerting them that they need to rebase


estute:
Tested this by:
* Pr that predates the fix in question (should fail) -> https://test-jenkins.testeng.edx.org/job/edx-platform-quality-flow-pr/41/
* Pr for the fix in question (should pass) -> https://test-jenkins.testeng.edx.org/job/edx-platform-quality-flow-pr/40/

PR #325, opened 2018-03-20
DSL: pr jobs shouldnt clone edx-platform

michaelyoungstrom:
Reverts edx/jenkins-job-dsl#324

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
@@ -81,10 +81,10 @@ class JenkinsPublicConstants {
         return secretMap
     }
 
-    public static final String JENKINS_PUBLIC_JUNIT_REPORTS = 'edx-platform*/**/nosetests.xml,edx-platform*/reports/acceptance/*.xml,' +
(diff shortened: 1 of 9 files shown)


src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
```diff
@@ -81,10 +81,10 @@ class JenkinsPublicConstants {
         return secretMap
     }
 
-    public static final String JENKINS_PUBLIC_JUNIT_REPORTS = 'edx-platform*/**/nosetests.xml,edx-platform*/reports/acceptance/*.xml,' +
-                                                              'edx-platform*/reports/quality.xml,edx-platform*/reports/javascript/' +
+    public static final String JENKINS_PUBLIC_JUNIT_REPORTS = '**/nosetests.xml,**/reports/acceptance/*.xml,' +
```
estute: Just want to double check- will the glob pattern work for both master and pr jobs? (because the filepaths will be different)

michaelyoungstrom: yep, i kicked off both pr/merge jobs and everything seems to be consistent.

PR #330, opened 2018-03-29
Make whitelabel options the default for sandbox configuration

ibrahimahmed443:
This PR makes the Whitelabel sandbox settings the default settings on Sandbox creation page. Specifically, it checks ecommerce, ecommerce_worker and set_whitelabel flags.

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -142,7 +142,7 @@ class CreateSandbox {
                 booleanParam("forum",true,"")
                 stringParam("forum_version","master","")
 
-                booleanParam("ecommerce",false,"")
+                booleanParam("ecommerce",true,"")
                 stringParam("ecommerce_version","master","")
 
                 booleanParam("notifier",false,"")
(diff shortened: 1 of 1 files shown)


ibrahimahmed443:
@bill-filler FYI

ibrahimahmed443:
@fredsmith can you please deploy this PR on sandbox page?

fredsmith:
@ibrahimahmed443 FYI, there's some defaults missing that is causing this change to break sandbox builds when included by default. I reverted this change and re-seeded the job. Please provide functional defaults and re-submit.

ibrahimahmed443:
@fredsmith @bill-filler The reason this failed was that the automated build does not have recreate checked. And in that case, dns_name is not passed on to the script as dns entries are not recreated. However, our script needs dns_name for creating sites, partners etc. Providing default dns_name in whitelabel playbook didn't solve the problem either because it recreated site entries in database, consequently we were getting MultipleObjectsReturned exception.

After doing some debugging, I came up with this PR: https://github.com/edx/configuration/pull/4462/files

The default sandbox creation is currently failing, so I was not able to test this extensively so far.

PR #334, opened 2018-04-05
Add analytics_api as optional install for sandboxes

georgebabey:
Runway for the analytics_api config changes in https://github.com/edx/configuration/pull/4443


feanil:
@georgebabey I've re-seeded the job but without also making changes to https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh your change here isn't gonna do much good.  Not sure if you have that PR already and it hasn't landed?

georgebabey:
Thanks @feanil!  I have a PR out (https://github.com/edx/configuration/pull/4443) but landing this first will let me point the configuration & analytics versions to my branches to verify all is well

PR #336, opened 2018-04-11
add makefile for build jenkins

pwnage101:
This adds a Makefile for starting and configuring the build_jenkins docker image for local development or for the openedx community.  I tried my best to make it generic enough to be applicable to both tools and build jenkins, but in many ways the two are still quite different.

The changes in this PR have enabled me to build a pipeline job for retirement (https://github.com/edx/jenkins-job-dsl/pull/337) currently by running:

1. `make docker.run.jenkins_build`
2. create ansible_overrides.yml and bump the version of the job-dsl plugin
3. `make docker.configure.jenkins_build`
4. open the manually_seed_one_job job and manually build the retirement job

At least step 2 will be eliminated via https://github.com/edx/jenkins-configuration/pull/83.

I was hoping to automate step 4, but I already spent too much time figuring that out so I'll skip for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,79 @@
+SHELL := /usr/bin/env bash
+.DEFAULT_GOAL := help
+
+# List of services defined in docker-compose.yml.  Guard this with a check for
+# the existence of the docker-compose program since we don't want to exit
+# before showing help text.
+ifneq (, $(shell which docker-compose))
+    DOCKER_SERVICES := $(shell docker-compose config --services)
+endif
+
+.PHONY: help clean docker.clean $(DOCKER_SERVICES:%=docker.run.%) $(DOCKER_SERVICES:%=docker.stop.%) $(DOCKER_SERVICES:%=docker.running.%) $(DOCKER_SERVICES:%=docker.configure.%)
+
+help:
+	@echo 'Makefile for launching and configuring edX-related jenkins services.'
+	@echo ''
+	@echo 'These jenkins services are launched via Docker, and are primarily intended for'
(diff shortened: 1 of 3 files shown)


pwnage101:
FYI @edx/testeng

Makefile
```diff
@@ -0,0 +1,72 @@
...
+# docker.install_cli.$service
+# "/jnlpJars/jenkins-cli.jar" is the standard location for the jar.
+# Every jenkins master has this file.
+# More info: https://jenkins.io/doc/book/managing/cli/#using-the-cli-client
+$(DOCKER_SERVICES:%=docker.install_cli.%) : docker.install_cli.% :
+	docker-compose exec $* /bin/bash -c 'curl --output /opt/jenkins-cli.jar http://localhost:8080/jnlpJars/jenkins-cli.jar'
+
+# docker.seed_job.$service
+$(DOCKER_SERVICES:%=docker.seed_job.%) : docker.seed_job.% :
+	docker-compose exec $* /bin/bash -c 'su -l jenkins --shell=/bin/bash -c "java -jar /opt/jenkins-cli.jar -auth admin:$$(cat /var/lib/jenkins/secrets/initialAdminPassword) build -v -s -p DSL_SCRIPT='$(DSL_SCRIPT)' manually_seed_one_job"'
```
pwnage101: @estute can you imagine a more succinct and robust way of writing this?  Specifically, it does not handle the case where the administrator changes the admin password from the default.  I looked over all the CLI options, but it looks like one way or another the local client needs to authenticate, even though we have direct access to the filesystem.

PR #337, opened 2018-04-12
Initial shell of a retirement job for the GDPR pipeline

pwnage101:
It looks like a pipeline!

2018-04-12-164431_1332x609_scrot

### DIFF
--- a/platform/jobs/RetirementJob.groovy
+++ b/platform/jobs/RetirementJob.groovy
@@ -1,0 +1,39 @@
+/* UserRetirementJobs.groovy
+ *
+ * Defines jobs which orchestrate user retirement.
+ */
+
+package platform
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_TEAM_SECURITY
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+// This is the job DSL responsible for creating the main pipeline job.
+pipelineJob('User Retirement') {
+    description('Retire a single user which has confirmed account deletion')
+
+    // Only a subset of edx employees should be allowed to control this job,
+    // but customer support can read and discover.
+    authorization {
+        blocksInheritance(true)
+        List membersWithFullControl = ['edx/platform-team']
+        membersWithFullControl.each { emp ->
+            permissionAll(emp)
+        }
+        // TODO PLAT-2036: uncomment the following two lines when we add the
+        // appropriate github group.
+        //permission('hudson.model.Item.Read', 'edx/customer-support')
+        //permission('hudson.model.Item.Discover', 'edx/customer-support')
+    }
+
+    // Parameters are defined in the pipeline DSL, so leave this line commented
+    // out.
+    // parameters { }
+
+    definition {
+        cps {
+            sandbox()  // enable the groovy sandbox
+            script(readFileFromWorkspace('platform/resources/RetirementPipelineDefinition.groovy'))
+        }
+    }
+}
--- a/platform/resources/RetirementPipelineDefinition.groovy
+++ b/platform/resources/RetirementPipelineDefinition.groovy
@@ -1,0 +1,83 @@
+/*
+ * Main user retirement pipeline
+ */
+pipeline {
+    agent {
+        // Try to run on standard jenkins-worker instances, otherwise the
+        // backup-runner will suffice as fallback.  Master is currently
+        // designated as a backup-runner.
+        label 'jenkins-worker'
+    }
+    parameters {
+        string(
+            name: 'environment',
+            defaultValue: '',  // caller must specify environment
+            description: 'The environment-deployment to retire the user from'
+        )
+        string(name: 'original_username', defaultValue: '', description: 'The original username of the user to retire.')
+        string(name: 'original_email', defaultValue: '', description: 'The original email address of the user to retire.')
+        string(name: 'retired_username', defaultValue: '', description: 'The retired username of the user to retire, containing the hash.')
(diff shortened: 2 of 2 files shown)


platform/jobs/RetirementJob.groovy
```diff
@@ -0,0 +1,33 @@
...
+/* This is the job DSL responsible for creating the pipeline job.
+ */
+pipelineJob('retirement') {
+    description(
+        'Retire a group of users which has been queued up.'
+    )
+
+    /* Give all job control a read permissions to all edx employees. */
+    /* TODO: make this configurable */
+    authorization JENKINS_PUBLIC_TEAM_SECURITY.call(['edx'])
```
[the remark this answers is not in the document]
estute: @pwnage101 in the past, the thinking has been that these jobs are first and foremost job configurations specific for edx infrastructure, but that they can be modified to suit others needs. That being said, perhaps we should make these easier to parameterize. It would be nice to pass values (in this case, the team name) in to the seed job (or make it available as a variable or something) so that members of the open source community don't have to fork or manually edit the DSL each time they wish to seed.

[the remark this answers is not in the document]
estute: As for your other question about the auth strategy, it would probably be something like this (using dummy names):

// Make sure that you block the inherited security scheme (it may be more permissive than what you want)
blocksInheritance(true)

// Either manually feed the users you want to grant control to OR create github group(s) and put them in this list
List membersWithFullControl = ['employee1', 'employee2']                                                                                                                                                                                                                                                                                                    
membersWithFullControl.each { emp ->
    permissionAll(emp)
}
// individually give 'discover' (find the job on the Jenkins job list/hit the URL) and 'read' (view logs/artifacts of the build) to customer support
permission('hudson.model.Item.Read', 'customer-support')
permission('hudson.model.Item.Discover', 'customer-support')

estute:
@pwnage101 do you want me to review this even though it is only a shell?

pwnage101:
Yes,  but use your best judgment regarding what looks like a TODO  (in a bit I'll update this PR to make those sections clearer).  We've structured our ticketing to split the jenkins work into a few separate tickets, including a separate ticket for each stage.

In this PR, I'm focusing on hopefully getting auth nailed.

platform/resources/RetirementPipelineDefinition.groovy
```diff
@@ -0,0 +1,83 @@
+/*
+ * Main user retirement pipeline
+ */
+pipeline {
+    agent {
+        // Try to run on standard jenkins-worker instances, otherwise the
+        // backup-runner will suffice as fallback.  Master is currently
+        // designated as a backup-runner.
+        label 'jenkins-worker||backup-runner'
```
pwnage101: jenkins-worker instances are the default label for this job, but the build jenkins docker container won't work with AWS out of the box.  I noticed that the jenkins master had the "backup-runner" label, so I decided to make it a fallback runner for this pipeline.

platform/jobs/RetirementJob.groovy
```diff
@@ -0,0 +1,39 @@
...
+    authorization {
+        blocksInheritance(true)
+        List membersWithFullControl = ['edx/platform-team']
+        membersWithFullControl.each { emp ->
+            permissionAll(emp)
+        }
+        // TODO PLAT-2036: uncomment the following two lines when we add the
+        // appropriate github group.
+        //permission('hudson.model.Item.Read', 'edx/customer-support')
+        //permission('hudson.model.Item.Discover', 'edx/customer-support')
```
pwnage101: I added this authorization section per your recommendation, but it's quite specific to edX, as you warned.  For now I think it will suffice to get this off the ground.

platform/resources/RetirementPipelineDefinition.groovy
```diff
@@ -0,0 +1,83 @@
...
+                // checkout secure repo to get secure defaults such as API auth client keys
+                sh 'echo "running the Checkout stage"'
+
+            }
+        }
+        stage('LMS - lock account') {
+            when { expression { return params.do_lms_lock_account } }
+            steps {
+                sh 'echo "running the LMS - lock account stage"'
+                // TODO: PLAT-2047
```
pwnage101: Tried to drop TODOs in places where we have ticketed work.

platform/resources/RetirementPipelineDefinition.groovy
estute: I understand why you did this, but am concerned that it won't be treated as a fallback but rather as OR by the label picker. In the case that all jenkins-workers are occupied, this would run on the master node. Is this job going to be particularly resource heavy?

pwnage101:
actually, I was hoping to keep backup-runner a label for this job, which would help with continuous integration testing of the retirement workflow if we end up using the docker image for that.

pwnage101:
do you think the backup-runner label is a no-go for actual build jenkins?  I was hoping that it might not ever run these jobs on master as long as it can provision a jenkins-worker.

estute:
We are often in a scenario in which all of the jenkins workers in a cluster are occupied. In this case, I *believe* that jenkins would favor running it on master, as it is available, rather than spinning up a new worker.

estute:
Because I am unsure of the implications of this job, I tend to think it would be better to prevent this from happening. We do run a few jobs on master, especially ones that manipulate jenkins (and workers) themselves, but try to limit it. I doubt this job will be a real resource hog, but it's possible that something like a change to a script/job could cause master to become non-performant/experience errors.

pwnage101:
I went ahead and created https://github.com/edx/configuration/pull/4491 to solve this problem for now, and I'll merge this PR.  We can always come back and change this pipeline if needed.

PR #338, opened 2018-04-13
Create hawthorn ci jobs

michaelyoungstrom:


michaelyoungstrom:
@estute We decided to disable them when we refactored quality, with our main reasoning being that most commits are cherry picked onto release branches anyways. Ned was okay with it. But since hawthorn has the change maybe we should add it back.

PR #340, opened 2018-04-17
Change git checkouts on edx-platform CI jobs from http to ssh

michaelyoungstrom:
When we have numerous concurrent http git clones, several fail. We saw online that ssh may be more reliable, so this is a test to see for ourselves.

Also, I refactored the subset job to no longer rely on a secret (similar to what we did for other platform jobs).


michaelyoungstrom:
@estute PR flow jobs only need to checkout edx-platform from their respective shards, which is handled by the subset job.

estute:
Ah yes. And we don't really need to care about the testeng-ci scripts because our working theory is that this only affects large clones?

michaelyoungstrom:
@estue The failure is only happening on edx-platform clones, so I figured we should start there.

PR #342, opened 2018-04-19
Retirement pipeline: secrets configuration

pwnage101:
This PR establishes a minimally viable model for retirement pipeline credentials.

For this iteration, I did not implement loading credentials from a repository because it would add about 50-100 lines of boilerplate for every stage since we'd have to check out a secure git repo for every stage for security reasons.  I also acknowledge that this has taken a really long time to get over the line, and I'm tired of reading shoddy jenkins documentation for a DSL based on a DSL based on Java, so I present to you something that technically works.

Also, it ended up being easier to define one YAML blob containing all client credentials across all IDAs and environments.

The credential file is to be defined in testeng-secure, and groovy init scripts will populate those credentials automatically.  Stu, does this sound correct?

### DIFF
--- a/platform/resources/RetirementPipelineDefinition.groovy
+++ b/platform/resources/RetirementPipelineDefinition.groovy
@@ -3,10 +3,12 @@
  */
 pipeline {
     agent {
-        // Try to run on standard jenkins-worker instances, otherwise the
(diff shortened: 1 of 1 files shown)


pwnage101:
closing this since we're pivoting off of jenkins pipeline

PR #343, opened 2018-04-20
create pingdom alerts on merge

syed-awais-ali:


devops/jobs/CreatePingdomAlerts.groovy
```diff
@@ -0,0 +1,97 @@
...
+
+    Variables without defaults are marked (required)
+
+    Variables consumed for this job:
+        * NOTIFY_ON_FAILURE: (Required - email address)
+        * SECURE_GIT_CREDENTIALS: (Required - jenkins name of git credentials)
+        * PINGDOM_EMAIL: (Required - The Pingdom Email)
+        * PINGDOM_API_KEY: (Required - Pingdom API Key)
+        * PINGDOM_PASSWORD: (Required - Pingdom Password)
+        * PINGDOM_ALERT_CONFIG_FILE (The pingdom Alert config file)
```
jibsheet: I assumed this would be a config file in the configuration-internal repo
Given that, we need to clone that repo too, right?

devops/jobs/CreatePingdomAlerts.groovy
```diff
@@ -0,0 +1,97 @@
...
+import static org.edx.jenkins.dsl.Constants.common_wrappers
+import static org.edx.jenkins.dsl.Constants.common_logrotator
+
+
+class CreatePingdomAlerts {
+    public static def job = { dslFactory, extraVars ->
+        assert extraVars.containsKey('PINGDOM_EMAIL') : "Required PINGDOM_EMAIL setting missing from configuration"
+        assert extraVars.containsKey('PINGDOM_ALERT_CONFIG_FILE') : "Required pingdom alert config file(PINGDOM_ALERT_CONFIG_FILE) missing from configuration"
+        dslFactory.job(extraVars.get("FOLDER_NAME","Monitoring") + "/create_pingdom_alerts") {
+            assert extraVars.containsKey('CONFIGURATION_REPO_URL') : "Please define CONFIGURATION_REPO_URL"
```
jibsheet: Please document this with the other vars above, and note from some of the other webhook trigger jobs which format of URL you need (I believe it's documented in the DockerCI jobs)

devops/jobs/CreatePingdomAlerts.groovy
```diff
@@ -0,0 +1,121 @@
...
+                        url(extraVars.get("CONFIGURATION_REPO_URL"))
+                        branch(config_branch)
+                    }
+                    extensions {
+                        relativeTargetDirectory('configuration')
+                    }
+                }
+            }
+
+            triggers merge_to_master_trigger(config_branch)
```
jibsheet: We only get to autorun on one repo in this version of Jenkins so we have to choose between configuration and configuration-internal
I assume we want to autorun when internal changes the config file, right?
You'll also need to update properties above since it's tied in to the trigger

jibsheet: I'm not sure what changed, since it appears that the trigger still goes off of master on the configuration repo, and I assume we want to trigger on internal merges.  Can you confirm that?  You should be able to seed this in the docker container and see the jenkins job's trigger and properties settings.

syed-awais-ali: @jibsheet I have defined the value of config_branch to be master and on [line 64](https://github.com/edx/jenkins-job-dsl/pull/343/files#diff-894a46f08e932836c2696f97bd875d6bR64)  I have changed the repo to CONFIGURATION_INTERNAL_REPO. I am now going to test this locally in docker

jibsheet: Cool, thanks, let me know how it goes and I'll do the final review.
I suspect we'll need to do some work when it goes live to get the webhooks configured fully, but that's more for after it gets to tools-edx-jenkins

syed-awais-ali:
Relevant Internal PR: https://github.com/edx/edx-internal/pull/388

Jenkins-dsl-internal PR: https://github.com/edx/jenkins-job-dsl-internal/pull/168

PR #344, opened 2018-04-20
[loadtestDriver.groovy] change default hatch rate to 5

pwnage101:
Values greater than about 5 cause problems with auto_auth and/or
enrollment of locust users.


amangano-edx:
Seems reasonable to me, though I don't know anything about this code.

PR #351, opened 2018-04-26
Add python pipeline job

michaelyoungstrom:
This is only triggered by the comment "jenkins run pipeline python"

### DIFF
--- a/platform/jobs/pipelines/edxPlatformPythonPipelinePr.groovy
+++ b/platform/jobs/pipelines/edxPlatformPythonPipelinePr.groovy
@@ -1,0 +1,66 @@
+package platform
+
+import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_TEAM_SECURITY
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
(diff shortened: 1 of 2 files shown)


platform/jobs/pipelines/edxPlatformPythonPipelinePr.groovy
```diff
@@ -0,0 +1,63 @@
...
+                git {
+                    remote {
+                        credentials('jenkins-worker')
+                        github('edx/edx-platform', 'ssh', 'github.com')
+                        refspec('+refs/heads/*:refs/remotes/origin/* +refs/pull/*:refs/remotes/origin/pr/*')
+                        branch('\${sha1}')
+                    }
+                }
+            }
+            scriptPath('Jenkinsfile')
```
jmbowman: Are we ultimately just going to have a single Jenkinsfile containing the pipeline(s) for all test types and quality checks, or are we going to need multiple Jenkinsfiles?  If the latter, we may want to group them in a directory with appropriate names.

michaelyoungstrom: I'm not really sure yet.

As we do more discovery in areas such as the external workspace plugin... we may find ways to cut down CPU time by avoiding redundant steps (like a git checkout). To me it could make sense to have one file if there's numerous shared steps like that.. but if not, splitting them out could make it easier to follow.

We can always change how this is organized as we add more complexity too. What do you think?

jmbowman: Sure, we can keep it this way for now to match the usual convention and then change it later if appropriate.

PR #352, opened 2018-04-27
Add Retirement Jobs DSL

pwnage101:
This contains both the master (collector) and slave (driver) job DSLs for user retirement.

PLAT-2100 - slave DSL
PLAT-2099 - master DSL

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -1,0 +1,265 @@
+/* UserRetirementJobs.groovy
+ *
+ * Defines jobs which orchestrate user retirement.
+ *
+ * Prerequisites:
+ *
+ * 1. A global credential of type "secret file" with ID
+ * "user-retirement-secure-default.yml".  The contents of this file is used by
+ * default as the credentials configuration for all user retirement scripts.
+ *
+ * 2. A global credential of type "SSH username with private key" and ID
+ * "jenkins-worker", used for checking out the secrets repo.
+ *
+ */
+
+package platform
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_TEAM_SECURITY
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+
+// ########### user-retirement-driver ###########
+// This defines the job for retiring individual users.
+job('user-retirement-driver') {
+    description('Drive the retirement of a single user which is ready for retirement immediately.')
+
+    // Only a subset of edx employees should be allowed to control this job,
+    // but customer support can read and discover.
+    authorization {
+        blocksInheritance(true)
+        List membersWithFullControl = ['edx/platform-team']
+        membersWithFullControl.each { emp ->
+            permissionAll(emp)
+        }
+        // TODO PLAT-2036: uncomment the following two lines when we add the
+        // appropriate github group.
+        //permission('hudson.model.Item.Read', 'edx/customer-support')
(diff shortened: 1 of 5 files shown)


platform/resources/user-retirement-collector.sh
```diff
@@ -0,0 +1,16 @@
...
+pip install -r requirements.txt
+
+# Create the directory where we will populate properties files, one per
+# downstream build.
+mkdir $LEARNERS_TO_RETIRE_PROPERTIES_DIR
+
+# Call the script to collect the list of learners that are to be retired.
+scripts/get_learners_to_retire.py
+    --secrets=$WORKSPACE/user-retirement-secure/$ENVIRONMENT.yml
+    --output_dir=$LEARNERS_TO_RETIRE_PROPERTIES_DIR
```
pwnage101: @doctoryes this was the CLI I was predicting for get_learners_to_retire.py.  This doesn't actually cover HTTP basic auth, but I'm tempted to also squeeze that into the user-retirement-secure repo in the same yaml blobs containing the client_secret.

The file names for each properties file under `$LEARNERS_TO_RETIRE_PROPERTIES_DIR` isn't important, other than they don't collide.

platform/jobs/RetirementJobs.groovy
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    parameters {
+        stringParam('TUBULAR_BRANCH', 'master', 'Repo branch for the tubular scripts.')
+        stringParam('ORIGINAL_USERNAME', '', '')
+        stringParam('RETIRED_USERNAME', '', '')
+        stringParam('ORIGINAL_EMAIL', '', '')
+        stringParam('RETIRED_EMAIL', '', '')
+        stringParam('CURRENT_STATE', 'PENDING', 'Current retirement state for this user before starting this build.')
+        stringParam('LAST_STATE', '', 'Last retirement state for this user before starting this build.')
```
pwnage101: @doctoryes except for TUBULAR_BRANCH, these are the names of the keys that should be present in each parameter file from the master python script.  Whether or not these are actually right as-written, this list needs to be be in sync with whatever the parameter files contain.

platform/jobs/RetirementPipelineJobs.groovy
```diff
@@ -1,4 +1,8 @@
-/* UserRetirementJobs.groovy
+// NOTE: These pipeline-enabled jobs have been retired in favor of the simpler
+// jenkins-native jobs in RetirementJobs.groovy which utilize Parameterized
+// Trigger plugin and a more python centric approach.
```
estute: Since these jobs have been retired (temporarily?), I would recommend disabling them directly in the DSL (https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.jobs.FreeStyleJob.disabled), so they don't get run

pwnage101: done.

platform/resources/user-retirement-driver.sh
```diff
@@ -0,0 +1,16 @@
...
+cp $USER_RETIREMENT_SECURE_EXTRA $WORKSPACE/user-retirement-secure/secure-extra.yml
+
+# prepare tubular
+cd $WORKSPACE/tubular
+pip install -r requirements.txt
+
+# Call the script to retire one learner.  This assumes the following build
+# parameters / environment variable is set: RETIREMENT_USERNAME.
+python scripts/retire_one_learner.py \
+    --config_file=$WORKSPACE/user-retirement-secure/$ENVIRONMENT.yml
```
pwnage101: @bmedx just double checking that all that's needed here is the config_file and username (via RETIREMENT_USERNAME) arguments.

PR #358, opened 2018-05-09
Upgrade DSL to 1.67

estute:
The two main changes to the DSL from versions 1.47 -> 1.67 are:
* GHPRB support dropped. Now need to use the slightly different GHPRB syntax
* Hipchat support droppped. Now you need to use Dynamic DSL


gradle.properties
```diff
@@ -1,2 +1,2 @@
-jobDslVersion=1.47
-jenkinsVersion=1.625
+jobDslVersion=1.67
+jenkinsVersion=2.50
```
michaelyoungstrom: Why 2.50? Minimum supported version?

michaelyoungstrom:
The main problem that was causing CI to fail was that:

1- we need to include jgrapht as a testCompile dependency (for the build flow plugin)

2- In order to prevent this error:
```
platform.edxPlatformMasterJobSpec > test seeding edxPlatformBokChoyMaster.groovy FAILED
    Expected no exception to be thrown, but got 'javaposse.jobdsl.dsl.DslScriptException'
        at spock.lang.Specification.noExceptionThrown(Specification.java:118)
        at platform.edxPlatformMasterJobSpec.test dsl creates correct xml(edxPlatformMasterSpec.groovy:42)
        Caused by:
        javaposse.jobdsl.dsl.DslScriptException: (script, line 103) plugin 'build-flow-plugin' needs to be installed
```
We need to actually install the build flow plugin on the test instance for CI. I followed the example that the jenkins DSL contributors give in order to set up the new testPlugins option: https://github.com/sheehan/job-dsl-gradle-example/blob/master/build.gradle

src/test/groovy/platform/edxPlatformMasterSpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+        noExceptionThrown()
+        generatedItems.jobs.size() == numJobs
+
+        where:
+        dslFile                                   | numJobs
+        'edxPlatformAccessibilityMaster.groovy'   | 5
+        'edxPlatformBokChoyMaster.groovy'         | 5
+        'edxPlatformJsMaster.groovy'              | 5
+        'edxPlatformLettuceMaster.groovy'         | 5
+        'edxPlatformPythonUnitTestsMaster.groovy' | 5
```
jmbowman: What triggered the change in job counts?

test-requirements.txt
```diff
@@ -2,7 +2,7 @@ coverage==3.7.1
 ddt==0.8.0
 nose==1.3.4
 pep8==1.5.7
-edx-lint==0.4.3
+edx-lint==0.5.5
```
jmbowman: We should switch this over to pip-tools later.

estute: ^ actually, we might be able to rip that out altogether. There are some python scripts in here that I am not convinced that the mobile team still needs. If that is the case, we can remove all things python from this repo (but in a separate PR)

src/test/groovy/platform/edxPlatformMasterSpec.groovy
estute: I had changed the test spec to focus on the Master jobs (which have different job counts) instead of the PR jobs. The reason for this was that the newer version of the GHPRB uses it's own syntax which is a lot harder to test using the Jenkins test harness

PR #363, opened 2018-05-17
[RetirementJobs] give educator build access to retirement driver jobs

pwnage101:
This is done primarily for allowing anyone in educator to conveniently run user-retirement-driver against stage, but this implicitly gives prod access as well.

Pinging @schenedx and @macdiesel since this expands job privileges to a wider set of users.  Alternatively, I could just give @sandroroux alone the build privileges to the job if educator seems too broad.


schenedx:
@pwnage101 I am fine with giving all educator the access to build.

PR #368, opened 2018-05-23
[Retirement] collector job send emails on failure

pwnage101:
Job failure is defined by any single one of the downstream driver jobs
failing to retire a learner, since this upstream job blocks on all
downstream jobs and sets the status to fail if any blocking job fails.

This depends on https://github.com/edx-ops/testeng-secure/pull/180

PLAT-2127

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -15,9 +15,33 @@
 
 package platform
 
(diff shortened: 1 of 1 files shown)


pwnage101:
I just tested this on test-jenkins, and it works.  It sent the email to the wrong address, but that's a separate PR to fix: https://github.com/edx-ops/testeng-secure/pull/181

PR #373, opened 2018-06-05
WIP:

coryleeio:


devops/resources/check-rabbit.sh
```diff
@@ -11,7 +11,7 @@ export EC2_CACHE_PATH="ec2-cache"
 assume-role ${ROLE_ARN}
 
 
-cd $WORKSPACE/configuration/playbooks/edx-east
+cd $WORKSPACE/configuration/playbooks
```
jibsheet: I wonder if we can delete this file/job instead of even fixing it?

PR #374, opened 2018-06-07
Add job to check and notify mongo agents update(OPS-3162)

nadeemshahzad:


devops/jobs/MongoAgentsUpdateCheck.groovy
```diff
@@ -0,0 +1,67 @@
+package devops.jobs
+import static org.edx.jenkins.dsl.Constants.common_logrotator
```
jibsheet: Please use the common_logrotator, otherwise we'll never retire logs

devops/jobs/MongoAgentsUpdateCheck.groovy
```diff
@@ -0,0 +1,67 @@
...
+                        dslFactory.readFileFromWorkspace("devops/resources/mongo-agents-update-check.sh")
+                    )
+                }
+            }
+
+            publishers {
+
+                extendedEmail {
+                    recipientList('devops@edx.org')
+                    replyToList('devops@edx.org')
```
jibsheet: This is a job in the public repo.  Please don't hardcode our email address.  Pass it in from the configuration file used by other monitoring jobs.  The standard is NOTIFY_ON_FAILURE https://github.com/edx/edx-internal/blob/master/tools-edx-jenkins/monitoring.yml#L77 Feanil asks that we use something like devops+mms@edx.org to help filtering

nadeemshahzad:
@jibsheet Sorry for the disclosure of email id. I have updated the PRs with fixes. thanks

PR #379, opened 2018-06-14
Add GHPRB configure block for cancel on update var

michaelyoungstrom:
We need overrideGlobal set to false to actually do any deduping. This was actually added in the GHPRB repo as a DSL option a few days ago, there just isn't a version for it yet.

Also I removed some dead code.

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
@@ -190,22 +190,10 @@ class JenkinsPublicConstants {
         }
     }
 
-    public static final Closure GHPRB_BLACKLIST_BRANCH(String branchRegex) {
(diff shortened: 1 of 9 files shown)


src/main/groovy/org/edx/jenkins/dsl/JenkinsPublicConstants.groovy
```diff
@@ -221,19 +209,4 @@ class JenkinsPublicConstants {
...
-        return {
-            it / 'properties' / 'hudson.model.ParametersDefinitionProperty' / parameterDefinitions << 'hudson.model.PasswordParameterDefinition' {
-                name param.name
-                defaultValue Secret.fromString(param.default.toString()).getEncryptedValue()
-                description param.description
-            }
-        }
-
-    }
-
```
estute: Are you sure we should pull this out? I *think* it was used by loadtest or gdpr

michaelyoungstrom: I don't see it in use anywhere..

PR #382, opened 2018-06-20
TE-2560 Show quality failures on overview, enable Warnings plugin

jmbowman:
Make a few changes to make the nature of quality job failures more obvious:

* Run xUnit pass/fail checks against the new quality check results files added in https://github.com/edx/edx-platform/pull/18421
* Propagate those results files from the sharded quality jobs to the main quality job
* Replace the unsupported `Violations` plugin with the newer `Warnings` plugin that replaces it
* Enable warnings analysis on pull requests again; [TE-754](https://openedx.atlassian.net/browse/TE-754) notes that we had disabled it because `Violations` didn't work correctly with concurrent builds, but it seems that `Warnings` should

This PR depends on both the edx-platform PR noted above and the configuration PR to add the `Warnings` plugin: https://github.com/edx/configuration/pull/4637

### DIFF
--- a/platform/jobs/edxPlatformQualityDiff.groovy
+++ b/platform/jobs/edxPlatformQualityDiff.groovy
@@ -2,6 +2,7 @@ package platform
 
 import org.yaml.snakeyaml.Yaml
 import static org.edx.jenkins.dsl.JenkinsPublicConstants.GENERAL_PRIVATE_JOB_SECURITY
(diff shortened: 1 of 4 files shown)


jmbowman:
Made a couple of tweaks:

* The quality diff job doesn't need the XML results from the sharded quality jobs, but it **does** need to report its own results.
* We want the Warnings plugin to generate output even (especially) if the quality job fails.

jmbowman:
Another change: we don't need to run violations against both the console log and the output files, just the files.  I suspect that's been causing the doubling up of graphs and such in the UI, attempting to confirm.

PR #383, opened 2018-06-26
add job for running codecov metrics script

estute:
Job for running https://github.com/edx/testeng-ci/pull/158

### DIFF
--- a/testeng/jobs/gatherCodecovMetrics.groovy
+++ b/testeng/jobs/gatherCodecovMetrics.groovy
@@ -1,0 +1,50 @@
+package testeng
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.GENERAL_PRIVATE_JOB_SECURITY
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_MASKED_PASSWORD
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_HIPCHAT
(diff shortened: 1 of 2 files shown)


testeng/jobs/gatherCodecovMetrics.groovy
```diff
@@ -0,0 +1,50 @@
...
+
+    description('Gather metrics on how reliably codecov is posting statuses on pull requests')
+
+    authorization GENERAL_PRIVATE_JOB_SECURITY()
+
+    parameters {
+        stringParam('PULL_REQUEST_TIME_FRAME', '3600',
+                    'blah')
+        stringParam('CODECOV_TIME_FRAME', '1800',
+                    'blah')
```
jmbowman: Shouldn't "blah" get replaced with actual descriptions for these?

testeng/jobs/gatherCodecovMetrics.groovy
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    steps {
+        shell(readFileFromWorkspace('testeng/resources/gather-codecov-metrics.sh'))
+    }
+
+    publishers {
+        mailer('testeng@edx.org')
+        hipChatNotifier JENKINS_PUBLIC_HIPCHAT('')
```
jmbowman: What is the argument here supposed to represent?

estute: from the docs 'when left empty, the token specified in the global configuration will be used.'.

testeng/jobs/gatherCodecovMetrics.groovy
estute: thanks. that was a placeholder until i though of something better

PR #385, opened 2018-07-05
Email Optin DSL.

HassanJaveed84:


dataeng/jobs/AnalyticsEmailOptinMaster.groovy
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+        git {
+            remote {
+                url('https://github.com/edx/edx-analytics-exporter.git')
+                branch('$EXPORTER_BRANCH')
+            }
+            extensions {
+                pruneBranches()
+                relativeTargetDirectory('analytics-exporter')
```
brianhw: My ignorance: is there a reason this repo doesn't have a "credentials('1')" call here, and all others do?

dataeng/resources/exporter-worker.sh
```diff
@@ -0,0 +1,14 @@
...
+
+env | sort
+
+${EXPORTER_VENV}/bin/exporter \
+   ${EXTRA_OPTIONS} ${CONFIG_PATH} ${ORG_CONFIG_PATH} \
+   --org=${ORG} \
+   --task=OrgEmailOptInTask \
+   --output-bucket=${OUTPUT_BUCKET} \
+   --output-prefix=${OUTPUT_PREFIX} \
+   --external-prefix=databases/${DATE:-$TODAY} \
```
brianhw: Just curious if this is actually needed...

dataeng/resources/exporter-worker.sh
```diff
@@ -0,0 +1,14 @@
+TODAY=$(date +%d)
+
+env | sort
+
+${EXPORTER_VENV}/bin/exporter \
+   ${EXTRA_OPTIONS} ${CONFIG_PATH} ${ORG_CONFIG_PATH} \
+   --org=${ORG} \
+   --task=OrgEmailOptInTask \
```
brianhw: Just wondering why this particular task is in a file named "exporter-worker.sh"?    Is there a more specific filename that should be used?

dataeng/resources/setup-platform-venv.sh
```diff
@@ -0,0 +1,20 @@
...
+#pip install --exists-action w -r requirements/edx/django.txt
+#pip install --exists-action w -r requirements/edx/base.txt
+
+# Remove if https://github.com/edx/edx-platform/pull/7465 has been merged
+#if [[ -f requirements/edx/post.txt ]]; then
+#  pip install --exists-action w -r requirements/edx/post.txt
+#fi
+
+#pip install --exists-action w -r requirements/edx/github.txt
+#pip install --exists-action w -r requirements/edx/local.txt
```
brianhw: Is there a reason to leave in these commented-out pip installs?   Are they now replaced by 'make requirements'?   I looked at the Makefile in master for edx-platform, and I only found a 'clean' target.

dataeng/resources/setup-exporter-properties.sh
```diff
@@ -0,0 +1,35 @@
...
+# Export job configuration files 
+exporter-properties \
+    --output-bucket=${OUTPUT_BUCKET} \
+    --output-prefix=${OUTPUT_PREFIX} \
+    --orgs="${ORGS}" \
+    --include=platform_venv \
+    --include=exporter_vars \
+    ${CONFIG_PATH} \
+    ${WORKSPACE}/${ORG_CONFIG} \
+    organizations
```
brianhw: nit: It may be good to add a note here about "organizations".  It is the "Directory where to save Jenkins property files", and is used by the fileBuildParameterFactory in the master job to generate parameters for each of the worker jobs.  (Took me a while to figure that out....)

dataeng/jobs/AnalyticsEmailOptinMaster.groovy
HassanJaveed84: So the current Email Optin job uses https for this repo(I went with what we have), we can skip the credentials while using https. If we use 'git@github.com' url, we need to provide credentials.

dataeng/resources/exporter-worker.sh
HassanJaveed84: Went with what we have configure currently. 
Do we not need it anymore ?

Code:
https://github.com/edx/edx-analytics-exporter/blob/aec7c41332e3594ff6f884ecf2064d61a490b476/exporter/main.py#L29

dataeng/resources/exporter-worker.sh
HassanJaveed84: Would 'email_optin_worker' be more appropriate ?

dataeng/resources/setup-platform-venv.sh
HassanJaveed84: Oh, my bad. Accidentally commented out.

dataeng/jobs/AnalyticsEmailOptinMaster.groovy
azafty468: We are safe to continue using https as we don't have the TLS 1.0, TLS 1.1 issue on the new box.  However we may need to switch to SSH to take advantage of the SSH pem key in the global credentials.

dataeng/jobs/AnalyticsEmailOptinMaster.groovy
```diff
@@ -0,0 +1,105 @@
...
+            )
+        }
+        virtualenv {
+            nature("shell")
+            command(
+                readFileFromWorkspace("dataeng/resources/setup-exporter-properties.sh")
+            )
+        }
+        downstreamParameterized {
+            trigger('analytics-email-optin-worker') {
```
azafty468: There are some updates to the xml job configs, most notably that the newer version of Jenkins requires that all parameters are declared on all child jobs (previous versions allowed you to pass in parameters and they were automatically included).  If you look at the current job configs on the new jenkins box you can see the changes I've made.

dataeng/resources/exporter-worker.sh
azafty468: Yeah we probably want to make this named email opt-in specific.

dataeng/resources/exporter-worker.sh
azafty468: Hmmm...I'm not actually sure where this is used...I can't find it in the analytics exporter code or in the platform code.  I suspect it's unnecessary.  Also as an FYI, I have modifications to the analytics-exporter code currently in the zafft/testing branch.  Some of them will be carried forward (the removal of SUDO on all commands) and some of them will not be carried forward (the change of retries from 3 to 1).  We should resync this script before I make my PR.  As of right now there are no changes to any of the shell scripts.

dataeng/resources/setup-exporter-properties.sh
azafty468: Yeah I second this comment.  I had to follow the file path on the actual Jenkins server to figure this one out.  The local path is determined by the config file in the secure repo.  I mention this because I updated the path because it was hard coded to a temporal mount on the admin Jenkins box (/mnt/tmp/analytics-exporter/course-data/...).  I think we should reference where this guy is defined.

dataeng/resources/setup-platform-venv.sh
azafty468: As, so much cleaner with all these pip installs removed :)

azafty468: Actually wait, so the latest version of the exporter I successfully got to work with a current version of the platform + a cherry picked modification from Alex.  In this new version we only need to use make requirements.  The rest of the pip install commands for edx-platform go away.

dataeng/resources/exporter-worker.sh
brianhw: Oh, it's in the analytics exporter code, and is used by the CopyS3FileTask-derived task to copy CWSM data from the pipeline job output to include in the exports.  I'm just saying that it's not needed for the OrgEmail workflow, and it's there because of a copy-and-paste from the regular export.

dataeng/resources/exporter-worker.sh
brianhw: I was assuming we would have the regular export moving to this box as well.  This name would have been misleading (to me).

dataeng/resources/setup-platform-venv.sh
brianhw: Oh, I was looking at master.  Sigh.

dataeng/jobs/AnalyticsEmailOptinMaster.groovy
```diff
@@ -0,0 +1,106 @@
...
+                block {
+                    buildStepFailure('FAILURE')
+                    failure('FAILURE')
+                    unstable('UNSTABLE')
+                }
+                parameters {
+                    predefinedProp('MASTER_WORKSPACE', '${WORKSPACE}')
```
HassanJaveed84: Should these be removed as these are also declared on the worker ?

dataeng/jobs/AnalyticsEmailOptinWorker.groovy
```diff
@@ -0,0 +1,49 @@
...
+        timestamps()
+        buildName('#${BUILD_NUMBER} ${ENV,var="ORG"}')
+    }
+
+    steps {
+        shell(readFileFromWorkspace("dataeng/resources/email-optin-worker.sh"))
+    }
+
+    publishers {
+        textFinder("Task OrgEmailOptInTask failed fatally", '', true, false, false)
```
HassanJaveed84: Do we want to mark this build as unstable if the text is matched ?

HassanJaveed84:
This is ready for another review. I'll update the job names once it's finalized.

dataeng/jobs/AnalyticsEmailOptinWorker.groovy
azafty468: No I think when it actually fails fatally then the master should be marked as failure.  I think unstable doesn't happen on the Email Optin job.  The output for that job is pretty stable.  I think unstable tends to happen when there are partially built courses being exported during the rest of the exporter process.

dataeng/jobs/AnalyticsEmailOptinMaster.groovy
azafty468: Great question.  I suspect yes because the shell code is rolling them into the environment variables.  But also let's run a confirmation run after these have been removed.

HassanJaveed84:
Updated to use secret values from secure repo. 

See: [Secure Repo PR](https://github.com/edx-ops/analytics-secure/pull/164)

dataeng/jobs/AnalyticsEmailOptin.groovy
```diff
@@ -0,0 +1,182 @@
...
+                credentials('1')
+            }
+            extensions {
+                pruneBranches()
+                relativeTargetDirectory('analytics-exporter')
+            }
+        }
+        git {
+            remote {
+                url(commonConfigMap.get('SECURE_REPO_URL'))
```
brianhw: So this was my question from earlier this morning -- I was asking whether these would just be job parameters or something else.   $SECURE_BRANCH is a job parameter (on the next line), so I thought this could be as well.  Then it is up to the place that defines the actual parameters to pass to the job to provide these URLs as well.    So now I think I see that there are no job parameters being passed in when these define jobs that are automatically scheduled, so the values used in the parameter defaults have to come from some external location at job definition time instead of job runtime.  The purpose of the parameters, then, is to provide customization for when running jobs when they are *not* scheduled.  So, we get one file to define common variable values, pointed to by the environment variable COMMON_VARS.    

I guess in future, we could load a second file that was more customized if we found that we needed it.   The same global logic, but we would read COMMON_VARS_DIR, append an "analytics-email-optin-vars.yaml" path to the end, and try to read that if it exists, into say jobConfigMap.   Then have a method that looks in both commonConfigMap and jobConfigMap to get values, looking at jobConfigMap for overrides.   We might find that it's easier than shoving everything into one file with lots of per-job values.

HassanJaveed84: I could make this a parameter and provide the default. I just made it exactly like what we have on the admin box right now. I was thinking along the same lines that we would have one common file and if we needed more customization we would have yaml files per job.

brianhw: Sorry, that wasn't my intention.  This doesn't need to be a parameter since we don't expect anyone running this job manually to ever need to change that value.  My comment was more about customization in future with yaml files per job.   It sounds like we agree on that.

HassanJaveed84: Alright, so @azafty468 and I had a discussion and we agreed on having a COMMON_VARS_DIR parameter in the seed job. The individual DSL scripts would then append the yaml file name to the path(as per your suggestion) and read in the secret stuff for configuration the individual jobs. We can then add more yaml files on a per job basis(and have a common.yaml one like we do now) without needing to change the seed job. Does that seem appropriate ?

HassanJaveed84:
Updated to use ``COMMON_VARS_DIR`` parameter for reading secrets.

HassanJaveed84:
No change needed on the DSL side then ?

azafty468:
I'm going to say no changes needed.  I'm a bit vague here because I'm not 100% sure of all the uses of the analytics exporter.  I'll confirm by grepping the scheduler job configs.  But I'd say go ahead with this and if we do find something drastic that prevents me from merging my changes to analytics-exporter then we'll open another PR.

PR #386, opened 2018-07-05
Add retirement-partner-reporter job and shell script

bmedx:
WIP

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -330,3 +330,111 @@ job('user-retirement-collector') {
         mailer(mailingListMap['retirement_jobs_mailing_list'])
     }
 }
+
+// ########### retirement-partner-reporter ###########
+// Defines the job for running partner reporting
+job('retirement-partner-reporter') {
+    description('Run the partner reporting job and push the results to Google Drive.')
+
+    authorization {
+        blocksInheritance(true)
+        // Only core teams have full control of the reporter.
+        List membersWithFullControl = ['edx*platform-team', 'edx*testeng', 'edx*devops']
+        membersWithFullControl.each { emp ->
+            permissionAll(emp)
+        }
+        // Other engineering teams can view.
+        List extraMembersCanView = ['edx*learner']
+        extraMembersCanView.each { emp ->
+            permission('hudson.model.Item.Read', emp)
+            permission('hudson.model.Item.Discover', emp)
+        }
+    }
+
+    // This label is configured to disallow more than one simultaneous job
+    // per instance, and generally there are always available jenkins-worker
+    // instances idling.
+    label('jenkins-worker')
+
+    // Only one of these jobs should be running at a time per environment
+    concurrentBuild(true)
+    configure { project ->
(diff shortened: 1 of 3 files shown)


platform/resources/retirement-partner-reporter.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Create the directory where we will store reports, one per partner
+rm -rf $PARTNER_REPORTS_DIR
+mkdir $PARTNER_REPORTS_DIR
+
+# Call the script to generate the reports and upload them to Google Drive
+python scripts/retirement_partner_report.py \
+    --config_file=$WORKSPACE/user-retirement-secure/$ENVIRONMENT.yml \
+    --google_secrets_file=$WORKSPACE/user-retirement-secure/google-service-accounts/service-account-$ENVIRONMENT.json \
+    --output_dir=$PARTNER_REPORTS_DIR
```
bmedx: `--output_dir` is not supported in the script that is currently in PR, it will need to be added before this is merged.

platform/jobs/RetirementJobs.groovy
```diff
@@ -330,3 +330,109 @@ job('user-retirement-collector') {
...
+        project / 'properties' / 'hudson.plugins.throttleconcurrents.ThrottleJobProperty' <<
+            'limitOneJobWithMatchingParams'('true')
+    }
+
+    // keep jobs around for 30 days
+    logRotator JENKINS_PUBLIC_LOG_ROTATOR(30)
+
+    wrappers {
+        buildUserVars() /* gives us access to BUILD_USER_ID, among other things */
+        buildName('#${BUILD_NUMBER}')
```
bmedx: Not sure if there's a better build name here, maybe add the environment?

pwnage101: We use just the build number as the build name for most things, but yea probably adding the edx environment (`#${BUILD_NUMBER}, ${ENV,var="ENVIRONMENT"}`) would be helpful, just like in the user-retirement-collector job.

platform/jobs/RetirementJobs.groovy
```diff
@@ -330,3 +330,109 @@ job('user-retirement-collector') {
...
+            name('retirement-partner-reporter')
+            nature('shell')
+            command(readFileFromWorkspace('platform/resources/retirement-partner-reporter.sh'))
+        }
+    }
+
+    publishers {
+        // After all the build steps have completed, cleanup the workspace in
+        // case this worker instance is re-used for a different job.
+        wsCleanup()
```
pwnage101: Can you add a TODO comment about coming back to this list of publishers and adding an email alert?  Reference the following tickets:

* feature addressed in PLAT-2138
* which is blocked by TE-2618

platform/jobs/RetirementJobs.groovy
```diff
@@ -330,3 +330,111 @@ job('user-retirement-collector') {
...
+        }
+    }
+
+    // This label is configured to disallow more than one simultaneous job
+    // per instance, and generally there are always available jenkins-worker
+    // instances idling.
+    label('jenkins-worker')
+
+    // Only one of these jobs should be running at a time per environment
+    concurrentBuild(true)
```
estute: I think you want to flip this value

PR #390, opened 2018-07-13
added check for ymal file syntax

syed-awais-ali:


devops/resources/syntax-check-config.sh
```diff
@@ -32,6 +32,17 @@ else
   FAIL=1
 fi
 
+for i in `find . -name "*.yml" | sed '/ansible/d'`
+  do
+    python -c "import sys,yaml; yaml.load(open('$i'))" $i > /dev/null
```
coryleeio: It might be good to print the thing you are about to parse just so it's clear which file produced the error when it fails to parse.  Looks good otherwise

PR #399, opened 2018-07-20
Add user-retirement-bulk-status job and shell script

bmedx:
This PR creates a Jenkins job that allows us to move learners who are in the retirement process from one retirement state to another in bulk. Use cases for this are things like: bugs or connectivity issues causing numerous learners to end up in the ERRORED state or adding a new step to the process and needing to re-run users in COMPLETE. It also requires a date range that is used to limit the number of learners acted upon to those whose retirements were created within those dates.

Like the other retirement jobs, this one calls a Tubular Python script via a shell script. The Tubular side of this is here: https://github.com/edx/tubular/pull/254

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -599,3 +599,123 @@ job('retirement-partner-report-cleanup') {
         }
     }
 }
+
+// ########### user-retirement-bulk-status ###########
+// This defines the retirement bulk status change job for users in the retirement queue.
+job('user-retirement-bulk-status') {
+    description('Moves learners in the retirement queue from one retirement state to another.')
+
+    // Only a subset of edx employees should be allowed to control this job,
+    // but customer support can read and discover.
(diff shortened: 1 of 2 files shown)


platform/jobs/RetirementJobs.groovy
```diff
@@ -599,3 +599,122 @@ job('retirement-partner-report-cleanup') {
...
+        }
+    }
+
+    steps {
+        // This step calls the shell script which talks to LMS
+        virtualenv {
+            name('user-retirement-bulk-status')
+            nature('shell')
+            command(readFileFromWorkspace('platform/resources/user-retirement-bulk-status.sh'))
+        }
```
estute: Missing a '}'

bmedx: Should be fixed now

bmedx:
@estute I've updated the description, let me know if you need more info. 
@doctoryes @pwnage101 can you review?

PR #400, opened 2018-07-23
Remove count-workers job

michaelyoungstrom:
As part of the deduper clean-up, I realized this job really isn't needed anymore, since the Splunk plugin does a far more detailed analysis of our jenkins workers.


testeng/jobs/countWorkers.groovy
```diff
@@ -1,36 +0,0 @@
...
-    triggers{
-        cron('*/5 * * * *')
-    }
-
-    wrappers {
-        timestamps()
-    }
-
-    steps {
-        shell(readFileFromWorkspace('testeng/resources/count-workers.sh'))
```
jmbowman: Can this shell script go away too?  It in turn seems to try running a Python script that I can't find.

michaelyoungstrom: Ah yeah, great catch. It calls the script I am removing here: https://github.com/edx/testeng-ci/pull/163

PR #402, opened 2018-07-31
Add exporter DSL

HassanJaveed84:
TODO:

- [ ] DRY.


dataeng/resources/course-exporter-worker.sh
```diff
@@ -0,0 +1,12 @@
...
+
+env | sort
+
+${EXPORTER_VENV}/bin/exporter \
+   --org=${ORG} \
+   --output-bucket=${OUTPUT_BUCKET} \
+   --external-prefix=databases/${DATE:-$TODAY} \
+   --django-admin=${PLATFORM_VENV}/bin/django-admin.py \
+   --django-pythonpath=${PLATFORM_VENV}/edx-platform \
+   --gpg-keys=${GPG_KEYS_PATH} \
```
brianhw: Hmm.   I wonder if this will be actually needed.   One of the ways that the course exporter differs from the regular exporter is that the output is not encrypted.   That is useful for RDX, where we need to actually read and transform the dumps.  And useful for general testing of exporter changes, for example.  (Like the addition we're supposed to do for cohorts, IIRC.)

Oh, wait.   My mistake.   This is actually called by the exporter-slave job, not by the course exporter.  I was misled by the name of this file -- this script is actually exporting an org, not a list of courses.   The output does produce files at the level of individual courses, but the presence of job that takes courses as input made this confusing to me.

dataeng/resources/setup-course-exporter.sh
```diff
@@ -0,0 +1,32 @@
+# Create destination directory
+WORKING_DIRECTORY=/mnt/ephemeral-01/analytics-course-exporter
+mkdir -p ${WORKING_DIRECTORY}/course-data
+
+# Install requirements into this (exporter) virtual environment
+pushd analytics-exporter/
+pip install -r github_requirements.txt
+pip install --allow-external mysql-connector-python -e .
```
brianhw: As I think you and Andrew already noted, the --allow-external option is no longer used.  We pulled it from pipeline makefile in Jan. 2017:  https://github.com/edx/edx-analytics-pipeline/commit/fcebbb7859be66c54d4fa3e30e50f67ed20278e1#diff-b67911656ef5d18c4ae36cb6741b7965.

dataeng/jobs/AnalyticsExporter.groovy
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    wrappers {
+        timestamps()
+        buildName('#${BUILD_NUMBER} ${ENV,var="ORG"}')
+    }
+
+    steps {
+        shell(readFileFromWorkspace("dataeng/resources/course-exporter-worker.sh"))
```
brianhw: Should analytics-exporter-slave call org-exporter-worker.sh, instead of course-exporter-worker.sh, or something like that?  The name course-exporter-worker.sh seems a better fit to be called by analytics-exporter-course.

dataeng/jobs/AnalyticsExporter.groovy
```diff
@@ -0,0 +1,257 @@
...
+        virtualenv {
+            nature("shell")
+            command(
+                readFileFromWorkspace("dataeng/resources/setup-platform-venv.sh")
+            )
+        }
+        virtualenv {
+            nature("shell")
+            command(
+                readFileFromWorkspace("dataeng/resources/setup-course-exporter.sh")
```
brianhw: While this sets up the course-exporter, it also runs it.   I would suggest calling it run-course-exporter, or something like that.

dataeng/resources/setup-exporter.sh
```diff
@@ -0,0 +1,42 @@
+# Check that the DATE_MODIFIER is set to a Sunday, or if blank the current day is a Sunday
+DAYOFWEEK=$(date +%u ${DATE_MODIFIER})
+if [ $DAYOFWEEK != 7 ]; then
+   exit 1
+fi
```
brianhw: What is this about?   The job can only use data that was generated on Sundays?  I hope that job never breaks....

dataeng/jobs/AnalyticsExporter.groovy
```diff
@@ -0,0 +1,257 @@
...
+        timestamps()
+        buildName('#${BUILD_NUMBER} ${ENV,var="ORG"}')
+    }
+
+    steps {
+        shell(readFileFromWorkspace("dataeng/resources/course-exporter-worker.sh"))
+    }
+
+    publishers {
+        textFinder("\\[WARNING\\]", '', true, false, true)
```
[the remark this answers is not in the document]
HassanJaveed84: Added comment.

dataeng/resources/course-exporter-worker.sh
HassanJaveed84: renamed.

dataeng/resources/setup-course-exporter.sh
HassanJaveed84: Done.

dataeng/jobs/AnalyticsExporter.groovy
HassanJaveed84: Updated the name of script.

dataeng/resources/setup-exporter.sh
HassanJaveed84: I'm not sure why we have it, the CWSM dumps are only generated on sundays ?

HassanJaveed84:
This is ready for a final review.

dataeng/resources/setup-exporter.sh
azafty468: I wrote that because the job will run but fail to upload to S3 unless it's run on a date that's a sunday.

azafty468:
This looks good but I want to make sure the job history section is fixed before we merge this code into production (unless that investigation takes too long).

dataeng/jobs/createJobs.groovy
```diff
@@ -0,0 +1,30 @@
+import static analytics.AnalyticsEmailOptin.job as AnalyticsEmailOptinJob
+import static analytics.AnalyticsExporter.job as AnalyticsExporterJob
```
brianhw: nit:   why does github highlight the characters on either side of " as " on these two lines?  If they were just spaces I wouldn't expect a red square.

dataeng/resources/setup-exporter.sh
brianhw: The CWSM dumps are scheduled on Sundays, but if they fail on Sunday and are run on a Monday, then this prevents us from being able to change the date modifier to some other date.   I think that some new users (maybe Emilio?) when first running this specified today's date instead of the Sunday date, and I think Andrew put that in originally to deal with that.   I'm fine with it present or absent, until it becomes a problem one way or the other.

dataeng/jobs/createJobsTest.groovy
```diff
@@ -0,0 +1 @@
+import static analytics.AnalyticsEmailOptin.job as TestJob
```
HassanJaveed84: I'm not sure what's going on. I created this file using vim and the same red blocks appear.

dataeng/jobs/test.groovy
```diff
@@ -0,0 +1 @@
+import static analytics.AnalyticsEmailOptin.job as SomeJob
```
HassanJaveed84: This file was created using github itself.

dataeng/jobs/createJobs.groovy
HassanJaveed84: Not sure what's going on, files created using vim and github show same behavior. I found this: https://github.com/textmate/groovy.tmbundle/issues/17 which made me think it was textmate related(I use textmate). But doesn't look like that it is. olivex internal dsl repo show the same red blocks.

PR #403, opened 2018-07-31
add job for scanning for plugin upgrades

estute:


testeng/jobs/scanPluginUpdatesJob.groovy
```diff
@@ -0,0 +1,52 @@
...
+    label('jenkins-worker')
+    concurrentBuild(false)
+
+    scm {
+        git {
+            remote {
+                url('https://github.com/edx/jenkins-configuration.git')
+            }
+            branch('*/master')
+            browser()
```
jmbowman: What does the repository browser gain us in this case?

estute: This is a carryover from before we upgraded to Jenkins 2.x and all of the newer plugin versions. In the past, there was a bug that would cause an exception on jobs using git SCM w/o this in the DSL. I just checked, and it is no longer the case, so we can take it out

PR #404, opened 2018-08-01
specify diff quality target branch

estute:
Looks like this was left out when we moved from freestyle to flow jobs.


michaelyoungstrom:
image

PR #405, opened 2018-08-02
update groovy used in plugin scanning script

estute:
And finally, once https://github.com/edx/configuration/pull/4723 has merged in, this will use the correct Groovy installation


estute:
The actual fix was too use a System groovy script.

PR #407, opened 2018-08-09
[RetirementJobs] make sure the python io encoding is utf-8 compatible

pwnage101:
This makes sure that both PYTHONIOENCODING and LC_CTYPE are both set to utf-8. Adding these to environmentVariables{} might have been overkill, but I had already written that and decided to keep it.


macdiesel:
*fingers crossed*

PR #409, opened 2018-08-22
On failures also notify individuals with commits in change.

feanil:
This should hopefully reduce the burden on IT to fix issues introduced by developers.


feanil:
Their description:
```
If this option is checked, the notification e-mail will be sent to individuals who have committed changes for the broken build (by assuming that those changes broke the build).
```

But it's unclear how smart it is.  If it's dumb, we can just switch to not having merge commits, I think.

PR #410, opened 2018-08-24
Enable the partner reporting weekly trigger for prod-edx

pwnage101:
We should be good to enable this trigger now that we don't expect the job to fail and require manual cleanup (i.e. manually deleting rows from the partner reporting statuses table since the script would error before that step).


platform/jobs/RetirementJobEdxTriggers.groovy
```diff
@@ -32,7 +32,7 @@ List jobConfigs = [
         environmentDeployment: 'prod-edx',
         extraMembersCanBuild: [],
         cron: '0 8 * * 2',  // 09:00 UTC every Tuesday.
```
pwnage101: FYI, 09:00 UTC is 05:00 EDT.

PR #412, opened 2018-08-29
Fix virtualenv read only access to DB job (OPS-3354)

nadeemshahzad:


nadeemshahzad:
Job fails installing requirements.txt with python virtualenv 3.5.

PR #418, opened 2018-09-27
Pipeline jobs DSL

HassanJaveed84:
Included @rao-abdul-mannan user-activity DSL work. 
Added ability to deploy to multiple environments, environments will be specified in analytics-secure extra vars files.

### DIFF
--- a/dataeng/jobs/analytics/UserActivity.groovy
+++ b/dataeng/jobs/analytics/UserActivity.groovy
@@ -1,0 +1,27 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.date_interval_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+
+class UserActivity {
+    public static def job = { dslFactory, allVars ->
+        allVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("user-activity-$environment") {
+                logRotator common_log_rotator(allVars)
+                parameters common_parameters(allVars, env_config)
+                parameters date_interval_parameters(allVars)
+                multiscm common_multiscm(allVars)
+                triggers common_triggers(allVars, env_config)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    shell(dslFactory.readFileFromWorkspace("dataeng/resources/user-activity.sh"))
+                }
+            }
+        }
+    }
+}
--- a/dataeng/jobs/analytics/VideoTimeline.groovy
+++ b/dataeng/jobs/analytics/VideoTimeline.groovy
@@ -1,0 +1,30 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.date_interval_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+
+class VideoTimeline {
+    public static def job = { dslFactory, allVars ->
+        allVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("video-timeline-$environment") {
+                logRotator common_log_rotator(allVars)
+                parameters common_parameters(allVars, env_config)
+                parameters date_interval_parameters(allVars)
+                multiscm common_multiscm(allVars)
+                triggers common_triggers(allVars, env_config)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/video-timeline.sh'))
+                    if (env_config.get('SNITCH')) {
+                        shell('curl https://nosnch.in/' + env_config.get('SNITCH'))
+                    }
+                }
+            }
+        }
+    }
+}
--- a/dataeng/resources/user-activity.sh
+++ b/dataeng/resources/user-activity.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY*2 ))
+fi
+
+env | sort
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+  InsertToMysqlCourseActivityTask --local-scheduler \
+  --end-date $(date +%Y-%m-%d -d "$TO_DATE") \
+  --weeks 52 \
+  --n-reduce-tasks $NUM_REDUCE_TASKS \
+  --overwrite-mysql
--- a/dataeng/resources/video-timeline.sh
+++ b/dataeng/resources/video-timeline.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY * 2 ))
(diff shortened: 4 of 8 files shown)


src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public static def common_parameters = { extraVars, env=[:] ->
+        return {
+            stringParam('CLUSTER_NAME', env.get('CLUSTER_NAME', extraVars.get('CLUSTER_NAME')), 'Name of the EMR cluster to use for this job.')
+            stringParam('CONFIG_BRANCH', '$ANALYTICS_CONFIGURATION_RELEASE', 'e.g. tagname or origin/branchname, or $ANALYTICS_CONFIGURATION_RELEASE')
+            stringParam('CONFIG_REPO', 'git@github.com:edx/edx-analytics-configuration.git', '')
+            stringParam('EMR_MASTER_INSTANCE_TYPE', extraVars.get('EMR_MASTER_INSTANCE_TYPE', 'm4.2xlarge'), 'EC2 Instance type used for master.')
+            stringParam('EMR_WORKER_INSTANCE_TYPE_1', extraVars.get('EMR_WORKER_INSTANCE_TYPE_1', 'm4.2xlarge'), 'EC2 instance type used by workers.')
+            stringParam('EMR_WORKER_INSTANCE_TYPE_2', extraVars.get('EMR_WORKER_INSTANCE_TYPE_2', 'm4.4xlarge'), 'EC2 instance type used by workers.')
```
brianhw: Do we want to require that all jobs have two instance types?   I think there are some where we just want to define one type, and if the second is not defined, I think we should allow it to be missing.  Hardcoding a default makes that harder to do.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,104 @@
...
+            textParam('EMR_HADOOP_ENV_CONFIG', extraVars.get('EMR_HADOOP_ENV_CONFIG'), 'EMR Hadoop env configuration.')
+            textParam('EMR_MAPRED_SITE_CONFIG', extraVars.get('EMR_MAPRED_SITE_CONFIG'), 'EMR mapred-site configuration')
+            textParam('EMR_YARN_SITE_CONFIG', extraVars.get('EMR_YARN_SITE_CONFIG'), 'EMR yarn-site configuration.')
+            textParam('EXTRA_VARS', extraVars.get('EMR_EXTRA_VARS'), $/Extra variables to pass to the EMR provision/terminate ansible playbook.
```
brianhw: nit: This is why I was having trouble with the naming of extraVars elsewhere.    Are they really extra, or are they really 'allVars'?    It made sense when they were just the vars specified for the specific job, but the extraVars here are actually equal to "commonConfigMap + extraVars".

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,104 @@
...
+This text may reference other parameters in the task as shell variables, e.g.  $$CLUSTER_NAME./$)
+            stringParam('NOTIFY', '$PAGER_NOTIFY', 'Number of EMR instances to use for this job.')
```
brianhw: comment is wrong here.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+    }
+
+    public static def common_parameters = { extraVars, env=[:] ->
+        return {
+            stringParam('CLUSTER_NAME', env.get('CLUSTER_NAME', extraVars.get('CLUSTER_NAME')), 'Name of the EMR cluster to use for this job.')
+            stringParam('CONFIG_BRANCH', '$ANALYTICS_CONFIGURATION_RELEASE', 'e.g. tagname or origin/branchname, or $ANALYTICS_CONFIGURATION_RELEASE')
+            stringParam('CONFIG_REPO', 'git@github.com:edx/edx-analytics-configuration.git', '')
+            stringParam('EMR_MASTER_INSTANCE_TYPE', extraVars.get('EMR_MASTER_INSTANCE_TYPE', 'm4.2xlarge'), 'EC2 Instance type used for master.')
```
brianhw: Shouldn't this and other instance types vary by env as well as extraVars?  Or is the only way to scale differently for edge versus prod is to change the number of instances?

dataeng/jobs/analytics/UserActivity.groovy
```diff
@@ -0,0 +1,28 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.date_interval_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+
+class UserActivity {
+    public static def job = { dslFactory, extraVars ->
```
brianhw: I mention it below, but I wonder if we can call this something like 'allVars' instead of 'extraVars'.   We use extraVars elsewhere to mean the vars that are in addition to the common vars.    Here it's not clear that they are extra, as they are the only vars.

dataeng/jobs/analytics/VideoTimeline.groovy
```diff
@@ -0,0 +1,29 @@
...
+class VideoTimeline {
+    public static def job = { dslFactory, extraVars ->
+        extraVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("video-timeline-$environment") {
+                logRotator common_log_rotator(extraVars)
+                parameters common_parameters(extraVars, env_config)
+                parameters date_interval_parameters(extraVars)
+                multiscm common_multiscm(extraVars)
+                triggers {
+                    cron(env_config.get('JOB_FREQUENCY', ''))
```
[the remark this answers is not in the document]
HassanJaveed84: It wouldn't schedule it, that's why I went with this. I can try to make this conditional.

dataeng/resources/video-timeline.sh
```diff
@@ -0,0 +1,8 @@
+if [ "x$NUM_REDUCE_TASKS" = "x" ]; then
```
estute: I would recommend adding a shebang here. Unless explicitly told, Jenkins default shell env is pulled from the path, and in certain cases, that is dash, instead of bash (or even sh), depending on how you configured your system.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    public static def common_publishers = { allVars ->
+        return {
+            mailer('$NOTIFY')
+        }
+    }
+
+    public static def common_triggers = { allVars, env=[:] ->
```
estute: Nit: This might end up being more confusing in the long run.  If someone has to quickly check a job, and they see `common_triggers` in the job DSL, they won't know if the job is scheduled or not, without digging further. What do you think about making the job explicitly add a cron statement, and then only add the closure to the DSL if you want to add the trigger?

HassanJaveed84: Do you mean something like this in the Job ?
```groovy
job_frequency = extraVars.get('JOB_FREQUENCY')
if (job_frequency) {
   triggers common_triggers(job_frequency)
}
```

pwnage101: @estute, it was done this way because each environment specified in job config uses a different trigger frequency.  It really just boils down to whether that data is stored in analytics-secure vs. jenkins-job-dsl in the job DSL groovy file.  I'm okay with either option.

estute: Understood. If it makes more sense for the team using these jobs to check analytics-secure for configurations, that's fine by me.

dataeng/resources/user-activity.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+if [ "x$NUM_REDUCE_TASKS" = "x" ]; then
```
pwnage101: ditto here, for consistency.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -0,0 +1,114 @@
+package org.edx.jenkins.dsl
+
+class AnalyticsConstants {
+
+    public static def common_log_rotator = { allVars ->
+        return {
+            daysToKeep(30)
```
pwnage101: FYI, I'm making this closure configurable in my PR: https://github.com/edx/jenkins-job-dsl/pull/422/files#diff-4bbef760511c0e8e8f5a2d4a527bf72c

PR #432, opened 2018-10-16
Reduce the number of builds we keep for tasks that run once per minute.

feanil:
Having 10,000+ builds isn't super useful and slows down the UI.


jdmulloy:
Maybe we could make it 3 days to cover weekends, although it could be argued that the on-call person should look at it closer to when it's paging. Is there some way to tell Jenkins to keep the build longer from the script? It might be nice to keep it longer if a queue is stuck.

feanil:
I don't think it makes sense to store them for very long.  I think we look at them when they are failing and if we want to keep one, we can mark the build to keep it forever which will override the auto-cleanup.

PR #433, opened 2018-10-16
Add two more dataeng jobs

pwnage101:
Added the following jobs:

* answer-distribution
* database-export-courseware-studentmodule

DE-914
DE-916

---

See also the corresponding analytics-secure PR: https://github.com/edx-ops/analytics-secure/pull/181

So far, I tested seeding the jobs:

http://jenkins.analytics.edx.org/view/all/job/answer-distribution-production/
http://jenkins.analytics.edx.org/view/all/job/answer-distribution-edge/
http://jenkins.analytics.edx.org/view/all/job/answer-distribution-release/

http://jenkins.analytics.edx.org/view/all/job/database-export-courseware-studentmodule-production/
http://jenkins.analytics.edx.org/view/all/job/database-export-courseware-studentmodule-edge/

### DIFF
--- a/dataeng/jobs/createJobs.groovy
+++ b/dataeng/jobs/createJobs.groovy
@@ -5,6 +5,8 @@ import static analytics.VideoTimeline.job as VideoTimelineJob
 import static analytics.UserLocationByCourse.job as UserLocationByCourseJob
 import static analytics.Enrollment.job as EnrollmentJob
 import static analytics.ModuleEngagement.job as ModuleEngagementJob
+import static analytics.DatabaseExportCoursewareStudentmodule.job as DatabaseExportCoursewareStudentmoduleJob
+import static analytics.AnswerDistribution.job as AnswerDistributionJob
 import org.yaml.snakeyaml.Yaml
 import org.yaml.snakeyaml.error.YAMLException
 
@@ -32,6 +34,8 @@ def taskMap = [
     USER_LOCATION_BY_COURSE_JOB: UserLocationByCourseJob,
     ENROLLMENT_JOB: EnrollmentJob,
     MODULE_ENGAGEMENT_JOB: ModuleEngagementJob,
+    DATABASE_EXPORT_COURSEWARE_STUDENTMODULE_JOB: DatabaseExportCoursewareStudentmoduleJob,
+    ANSWER_DISTRIBUTION_JOB: AnswerDistributionJob,
 ]
 
 for (task in taskMap) {
--- a/dataeng/resources/answer-distribution.sh
+++ b/dataeng/resources/answer-distribution.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY * 2 ))
+fi
+NUM_REDUCE_TASKS=$(( $NUM_REDUCE_TASKS * 2 ))
+
+now=$(date +%H)
+
+name=hourly
+src="$SOURCES"
+dest=${DESTINATION_PREFIX}/$now
+manifest=${DESTINATION_PREFIX}/$now/manifest.txt
+
+# Remove any previous intermediate output for this hour
+pip install awscli
+aws s3 rm --recursive $dest || true
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+  AnswerDistributionWorkflow --local-scheduler \
+  --src $src --dest $dest --name $name --output-root $OUTPUT_URL \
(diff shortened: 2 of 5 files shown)


dataeng/resources/answer-distribution.sh
```diff
@@ -0,0 +1,30 @@
...
+dest=${DESTINATION_PREFIX}/$now
+if [ -z "$MANIFEST_PREFIX" ]; then
+    MANIFEST_PREFIX="$DESTINATION_PREFIX"
+fi
+manifest=${MANIFEST_PREFIX}/$now/manifest.txt
+
+# Remove any previous intermediate output for this hour
+pip install awscli
+aws s3 rm --recursive $dest || true
+aws s3 rm $manifest || true
```
pwnage101: Normally $manifest is under $dest, but in the -release version of the job it is not.  It should not hurt to attempt to delete the file twice anyway.

dataeng/jobs/createJobs.groovy
```diff
@@ -25,13 +27,15 @@ try {
...
-    USER_LOCATION_BY_COURSE_JOB: UserLocationByCourseJob,
-    ENROLLMENT_JOB: EnrollmentJob,
-    MODULE_ENGAGEMENT_JOB: ModuleEngagementJob,
+//    ANALYTICS_EMAIL_OPTIN_JOB: AnalyticsEmailOptinJob,
+//    ANALYTICS_EXPORTER_JOB: AnalyticsExporterJob,
+//    USER_ACTIVITY_JOB: UserActivityJob,
+//    VIDEO_TIMELINE_JOB: VideoTimelineJob,
+//    USER_LOCATION_BY_COURSE_JOB: UserLocationByCourseJob,
+//    ENROLLMENT_JOB: EnrollmentJob,
+//    MODULE_ENGAGEMENT_JOB: ModuleEngagementJob,
```
brianhw: Oh, interesting.  You have to comment these out so they don't get overwritten?   Or so they don't fail?

dataeng/resources/answer-distribution.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY * 2 ))
+fi
+NUM_REDUCE_TASKS=$(( $NUM_REDUCE_TASKS * 2 ))
```
brianhw: We don't have to do this now, but at some point we should really fix this logic so that we figure out the multiplier for NUM_TASK_CAPACITY to determine NUM_REDUCE_TASKS, depending on the instance type.  Rather than hardcoding it here like this.

dataeng/resources/answer-distribution.sh
brianhw: I don't remember the specific ways in which the different jobs have different ways of structuring their intermediate output, but I don't think we need to go out of our way to preserve it, especially for the dev-edx.cfg version.  (I'm not sure if that's what you mean by -release.)   I expect there is one way that is more sensible, and I would hope it is the production setup.   Ultimately it's all currently going to edx-analytics-scratch bucket anyway, which is not consistent with other jobs.  Keep in mind that this was the first job we ever built, back four plus years ago.   So don't work too hard to preserve the idiosyncracies. : )

dataeng/resources/answer-distribution.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Remove any previous intermediate output for this hour
+pip install awscli
+aws s3 rm --recursive $dest || true
+aws s3 rm $manifest || true
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+  AnswerDistributionWorkflow --local-scheduler \
+  --src $src --dest $dest --name $name --output-root $OUTPUT_URL \
+  --include '[\"*tracking.log*20131*.gz\",\"*tracking.log*2014*.gz\",\"*tracking.log*2015*.gz\",\"*tracking.log*2016*.gz\",\"*tracking.log*2017*.gz\",\"*tracking.log*2018*.gz\"]' \
```
brianhw: We only have to update this file once a year! : )

dataeng/jobs/createJobs.groovy
pwnage101: This is part of a second commit that i would delete before merging.  Our job seeding infrastructure updates all jobs at once, but for testing I want to limit the blast radius.  Perhaps we should revisit how aggressive the createJobs job is.

dataeng/resources/answer-distribution.sh
pwnage101: hah, yea, i don't necessarily know what to do about that.  Also the first file looks like it has a typo in it (the original one did too).

PR #435, opened 2018-10-17
Finance report DSL.

HassanJaveed84:
Secure repo PR:  https://github.com/edx-ops/analytics-secure/pull/183


HassanJaveed84:
Good catch @brianhw. I've added it to common.yaml in the secure repo.

PR #437, opened 2018-10-19
Vertica read replica import dsl

rao-abdul-mannan:


dataeng/resources/vertica-replica-import.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+if [ "x$NUM_REDUCE_TASKS" = "x" ]; then
```
pwnage101: Same thing, use `-z`.

dataeng/jobs/analytics/VerticaReplicaImport.groovy
```diff
@@ -0,0 +1,39 @@
...
+            dslFactory.job("${db.toLowerCase()}-read-replica-import") {
+                logRotator common_log_rotator(allVars)
+                parameters common_parameters(allVars, db_config)
+                parameters {
+                    stringParam('SCHEMA', db_config.get('SCHEMA'), 'Name of Vertica schema to write to.')
+                    stringParam('MARKER_SCHEMA', db_config.get('MARKER_SCHEMA'))
+                    stringParam('DB_CREDENTIALS', db_config.get('DB_CREDENTIALS'), 'Credentials for reading from MySQL database.')
+                    stringParam('DATABASE', db_config.get('DATABASE'), 'Name of MySQL database to copy from.')
+                    stringParam('EXCLUDE_FIELD', db_config.get('EXCLUDE_FIELD'), 'List of regular expressions matching field names that should not be copied.')
+                    stringParam('EXCLUDE', db_config.get('EXCLUDE'), 'List of regular expressions matching table names that should not be copied.')
```
brianhw: In the description string, is it possible to include an example?   It would also be good to note that the corresponding flag also needs to be included, e.g. --exclude or --exclude-field (or whatever it is again....).   (The reason for this is that it's truly optional, and in theory could be completely empty and should then be omitted.)

dataeng/resources/vertica-replica-import.sh
brianhw: Actually, why is this here?   Either we should not calculate this at all, or we should use it in the call to run-automated-task below (which we're not currently doing).  Then again, if NUM_TASK_INSTANCES here is zero, then there's little point in specifying 0 for --num-reduce-tasks (or whatever it is).

dataeng/jobs/analytics/VerticaReplicaImport.groovy
brianhw: The default for EXCLUDE is "".   But if one wants to set it, one needs to set "--exclude (list-of-patterns)".  The flag ("--exclude") needs to be included in the parameter value.    

Second, it might be useful to give an example of how to set a list of patterns, since it requires escaping quotes and not embedding spaces and so forth.

rao-abdul-mannan:
This is ready for final review.

PR #440, opened 2018-10-22
Add generate-warehouse-docs job

pwnage101:
DE-922

---

Seeded job: http://jenkins.analytics.edx.org/view/all/job/generate-warehouse-docs/

This job is triggered by the sql-scripts job defined in https://github.com/edx/jenkins-job-dsl/pull/444

See job config PR: https://github.com/edx-ops/analytics-secure/pull/188

### DIFF
--- a/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
+++ b/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
@@ -1,4 +1,6 @@
 package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
 
 class AnalyticsEmailOptin {
     public static def job = { dslFactory, allVars ->
@@ -61,8 +63,6 @@ class AnalyticsEmailOptin {
                         'Branch from the edx-analytics-exporter repository. For tags use tags/[tag-name]. Should be environment/production.')
                 stringParam('PLATFORM_BRANCH','zafft/analytics-exporter-settings-hotfix',
                         'Branch from the edx-platform repository. For tags use tags/[tag-name]. Should be release.')
-                stringParam('SECURE_BRANCH','release',
-                        'Branch from the analytics-secure repository, where the configuration settings reside. For tags use tags/[tag-name]')
                 stringParam('CONFIG_FILENAME','default.yaml', 'Name of configuration file in analytics-secure/analytics-exporter.')
                 stringParam('OUTPUT_BUCKET', allVars.get('EMAIL_OPTIN_OUTPUT_BUCKET'), 'Name of the bucket for the destination of the email opt-in data.')
                 stringParam('OUTPUT_PREFIX','email-opt-in-', 'Optional prefix to prepend to output filename.')
@@ -73,8 +73,9 @@ class AnalyticsEmailOptin {
                 stringParam('ORG_CONFIG','data-czar-keys/config.yaml', 'Path to the data-czar organization config file')
                 stringParam('DATA_CZAR_KEYS_BRANCH','master', 'Branch of the Data-czar-keys repository to use')
             }
+            parameters secure_scm_parameters(allVars)
 
-            multiscm{
+            multiscm secure_scm(allVars) << {
                 git {
                     remote {
                         url('git@github.com:edx/edx-platform.git')
@@ -95,17 +96,6 @@ class AnalyticsEmailOptin {
                     extensions {
                         pruneBranches()
                         relativeTargetDirectory('analytics-exporter')
-                    }
-                }
-                git {
-                    remote {
-                        url(allVars.get('SECURE_REPO_URL'))
-                        branch('$SECURE_BRANCH')
-                        credentials('1')
-                    }
-                    extensions {
-                        pruneBranches()
-                        relativeTargetDirectory('analytics-secure')
                     }
                 }
                 git {
--- a/dataeng/resources/answer-distribution.sh
+++ b/dataeng/resources/answer-distribution.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -z "$NUM_REDUCE_TASKS" ]; then
     NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY * 2 ))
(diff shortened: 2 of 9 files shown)


src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -85,13 +85,18 @@ class AnalyticsConstants {
...
             stringParam('NOTIFY', '$PAGER_NOTIFY', 'Space separated list of emails to send notifications to.')
             stringParam('NUM_TASK_CAPACITY', env.get('NUM_TASK_CAPACITY', allVars.get('NUM_TASK_CAPACITY')), 'Number of EMR instance capacity to use for this job.')
-            stringParam('SECURE_BRANCH', '$ANALYTICS_SECURE_RELEASE', 'e.g. tagname or origin/branchname, or $ANALYTICS_SECURE_RELEASE when released.')
             stringParam('SECURE_CONFIG', env.get('SECURE_CONFIG', 'analytics-tasks/prod-edx.cfg'), '')
-            stringParam('SECURE_REPO', allVars.get('SECURE_REPO_URL'), '')
             stringParam('TASKS_BRANCH', '$ANALYTICS_PIPELINE_RELEASE', 'e.g. tagname or origin/branchname,  e.g. origin/master or $ANALYTICS_PIPELINE_RELEASE')
             stringParam('TASKS_REPO', 'https://github.com/edx/edx-analytics-pipeline.git', 'Git repo containing the analytics pipeline tasks.')
             stringParam('TASK_USER', allVars.get('TASK_USER'), 'User which runs the analytics task on the EMR cluster.')
             booleanParam('TERMINATE', true, 'Terminate the EMR cluster after running the analytics task?')
+        } >> secure_scm_parameters(allVars)
```
pwnage101: I'm not super happy about this solution---I'd rather just include secure_scm_parameters() in every job that actually uses secure_scm, if that makes sense to everybody.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -42,7 +42,7 @@ class AnalyticsConstants {
         return {
             git {
                 remote {
-                    url(allVars.get('SECURE_REPO_URL'))
+                    url('$SECURE_REPO')
```
pwnage101: This was previously ignoring the SECURE_REPO parameter which could have caused confusion.

brianhw: Looks like a good catch.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
brianhw: Not sure what you're proposing here.  You pulled out two parameters because they were the only ones used for this particular job.  It might be possible to break up this common chunk into other smaller fragments as well.   But are you suggesting we should add the ">> secure_scm_parameters(allVars)" line in all the files that currently use common_parameters?   Just want to understand before I think too deeply about it. : )

dataeng/resources/answer-distribution.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
brianhw: Out of curiosity, what does this do?   I'm not familiar with "/usr/bin/env bash"...

pwnage101: `/usr/bin/env whatever` respects the PATH set in the current environment.  It's basically less strict than an absolute `/bin/whatever` about which executable to use.  Ever since the `/usr/bin/bash` "scare" a few years back, `/usr/bin/env bash` won many followers who were afraid that `/bin/bash` would disappear, but realistically OSes just symlinked `/usr/bin/bash` to `/bin/bash`.

I'm indifferent about it---I just tried to make them all consistent.

pwnage101: `#!/usr/bin/env python` is what really shows the true power of this style of shebang because it can respect your current virtualenv.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
pwnage101: I had assumed there were other consumers of the $SECURE_REPO, but there aren't.  In most places the repo variable is pulled directly out of allVars.

dataeng/resources/answer-distribution.sh
brianhw: Cool!   Thanks for the background -- good stuff to know.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
brianhw: But I still want to be sure I understand what "I'd rather just include secure_scm_parameters() in every job that actually uses secure_scm" means.

HassanJaveed84: I'm fine with majority consensus. For some additional context, whichever job calls run-automated-task.sh needs to have SECURE_REPO and SECURE_BRANCH parameters defined.

dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
```diff
@@ -61,8 +63,6 @@ class AnalyticsEmailOptin {
                         'Branch from the edx-analytics-exporter repository. For tags use tags/[tag-name]. Should be environment/production.')
                 stringParam('PLATFORM_BRANCH','zafft/analytics-exporter-settings-hotfix',
                         'Branch from the edx-platform repository. For tags use tags/[tag-name]. Should be release.')
-                stringParam('SECURE_BRANCH','release',
-                        'Branch from the analytics-secure repository, where the configuration settings reside. For tags use tags/[tag-name]')
```
brianhw: Note that removing this here may result in a different default value, depending on what ANALYTICS_SECURE_RELEASE is set to.

pwnage101: ANALYTICS_SECURE_RELEASE is set to 'release' as well.  It's set programmatically via:

https://github.com/edx-ops/analytics-secure/blob/master/ansible/analytics-jenkins-override-vars.yml#L102-L103

I suspect setting this variable globally is more confusing that its worth (adds another questionable layer of abstraction), but I could be missing context, and changing it would creep the scope of this PR even more than it has already.

pwnage101:
I made one small change.  Instead of this job requesting to be triggered by sql-scripts (via `upstream()`), I modified sql-scripts to trigger this job in its own job config (via `downstream()`) to be more consistent with the rest of the dataeng jobs.

PR #442, opened 2018-10-24
 DSL for loading events into S3.

HassanJaveed84:


dataeng/resources/load-json-events-to-s3.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY*2 ))
+fi
```
brianhw: Interesting that this is here and not in the other job.   I don't think it really matters, though, for these incremental jobs.   It's one reducer per day, and we're running one day.    (So the non-zero number of instances is helping only with the mapping phase.)

brianhw: Oh, interesting.  Apart from this line, the two bash scripts are identical.  You could just get rid of this script altogether, best I can tell, and have both jobs use load-events-to-s3.sh.

HassanJaveed84:
Ready for a final review.

dataeng/jobs/analytics/LoadEvents.groovy
```diff
@@ -0,0 +1,142 @@
...
+  Examples:
+
+  * --event-record-type JsonEventRecord
+  * --max-bad-records=1
+  * --overwrite
+  /$)
+                stringParam('DATASET', allVars.get('DATASET', ''))
+            }
+            multiscm common_multiscm(allVars)
+            triggers common_triggers(allVars)
```
pwnage101: does this still need common_triggers if it's being triggered by upstream?

HassanJaveed84: Good catch. Removed

PR #443, opened 2018-10-24
Add job to archive aged-off user retirements to S3.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-2237

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -620,6 +620,139 @@ job('retirement-partner-report-cleanup') {
     }
 }
 
+// ########### user-retirement-archiver ###########
+// Defines the job for archiving retired user statuses
+job('user-retirement-archiver') {
+    description('Archive the old user retirement status rows to AWS S3.')
+
+    authorization {
+        blocksInheritance(true)
+        // Only core teams have full control of the archiver.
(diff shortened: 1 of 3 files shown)


platform/jobs/RetirementJobs.groovy
```diff
@@ -620,6 +620,139 @@ job('retirement-partner-report-cleanup') {
...
+        colorizeOutput('xterm')
+        credentialsBinding {
+            file('USER_RETIREMENT_SECURE_DEFAULT', 'user-retirement-secure-default.yml')
+        }
+    }
+
+    parameters {
+        stringParam('TUBULAR_BRANCH', 'master', 'Repo branch for the tubular scripts.')
+        stringParam('ENVIRONMENT', 'secure-default', 'edx environment from which to archive users, in ENVIRONMENT-DEPLOYMENT format.')
+        stringParam('COOL_OFF_DAYS', '60', 'Number of days after which a retired learner status should be archived in S3.')
```
bmedx: Just to confirm that this is the expectation, this would be 60 days total since the retirement was requested, not 60 days after the retirement was completed (since we have the additional cool off days before the retirement is run).

doctoryes: Good point. I'll change to value to 67 and the comment to reflect this point.

doctoryes: Done.

PR #450, opened 2018-10-26
Add read only access job should fail when ssh key missing

coryleeio:


jdmulloy:
What does this do/what does it do before this change? Does it keep going with other users without failing so it's easy to miss a missing key?

coryleeio:
@jdmulloy This script is only used by the corresponding job, which is only used by yonkers.  It adds and revokes SSH users based on their having or lacking a key.  I want the job to blow up when it is misconfigured as I see numerous users that dont have SSH keys and it is silently failing to add them.

coryleeio:
What you said it correct - it continues without failing

jdmulloy:
Are these missing in github? I agree that we should make it more visible, but I'm concerned with blocking changes to correctly configured users because one user is misconfigured. Can we make the job fail and send an email or something, but not block other users?

coryleeio:
Actually - i'm going to close this because we have absent users, and we can't fail when they don't have an SSH key.

coryleeio:
Thanks @jdmulloy

PR #451, opened 2018-10-29
Vertica to BQ schema copy Business Intelligence dsl

rao-abdul-mannan:


dataeng/jobs/analytics/VerticaToBigquerySchemaCopyBusinessIntelligence.groovy
```diff
@@ -0,0 +1,33 @@
...
+                stringParam('GCP_CREDENTIALS', allVars.get('GCP_CREDENTIALS'), 'The path to the GCP credentials file.')
+                stringParam('EXCLUDE', allVars.get('EXCLUDE'))
+                stringParam('RUN_DATE', allVars.get('RUN_DATE'))
+            }
+            multiscm common_multiscm(allVars)
+            triggers common_triggers(allVars)
+            wrappers common_wrappers(allVars)
+            publishers common_publishers(allVars)
+            steps {
+                shell(dslFactory.readFileFromWorkspace('dataeng/resources/vertica-to-bigquery-schema-copy-business-intelligence.sh'))
```
brianhw: I would suggest renaming the script to remove business-intelligence from the name.   If we want a second job to copy another schema (e.g. finance or curated), then we should be able to use the same script with a different SCHEMA argument.

rao-abdul-mannan:
This could use another review.

PR #454, opened 2018-11-01
Backup Vertica DSL

rao-abdul-mannan:


rao-abdul-mannan:
This could use another review. I've consolidated partial backup dsl with this & closed that PR.

PR #455, opened 2018-11-01
Total events daily report DSL

HassanJaveed84:


dataeng/resources/total-events-daily-report.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
+    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY + 1 ))
+fi
+
+rm -rf venvs
```
brianhw: nit:  just curious why this is here?   Is it needed, or is the idea that venvs is not used and can be deleted?

PR #456, opened 2018-11-01
Remove references to hipchat on build jenkins

michaelyoungstrom:


platform/jobs/exampleJob.groovy
```diff
@@ -117,7 +114,7 @@ secretMap.each { jobConfigs ->
        }
 
        /* Post Build Steps */
-       /* Archive artifacts, archive jUnit Reports, send an email, message on hipchat */
+       /* Archive artifacts, archive jUnit Reports, send an email */
```
jmbowman: Should the example job be updated to also message Slack, since it was doing HipChat before?

PR #460, opened 2018-11-05
Enrollment validation events DSL.

HassanJaveed84:


dataeng/jobs/analytics/EnrollmentValidationEvents.groovy
```diff
@@ -0,0 +1,50 @@
...
+                multiscm common_multiscm(allVars)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                publishers {
+                    postBuildTask {
+                        if (environment == "production") {
+                            task('OVER THRESHOLD', 'exit 1', true)
+                        }
+                        if (environment == "edge") {
+                            task('Traceback', 'exit 1', true)
```
brianhw: nit: I think this is a holdover from the days we needed this for detecting failures.

dataeng/jobs/analytics/EnrollmentValidationEvents.groovy
```diff
@@ -0,0 +1,43 @@
...
+                        ' files as being "normal".)')
+                }
+                multiscm common_multiscm(allVars)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                publishers {
+                    postBuildTask {
+                        task('OVER THRESHOLD', 'exit 1', true)
+                        if (environment == "edge") {
+                            task('synthetic_enroll.log', 'exit 1', true)
```
HassanJaveed84: @brianhw do we still need this for edge only ?

HassanJaveed84:
Addressed feedback, ready for final review.

dataeng/resources/enrollment-validation-events.sh
```diff
@@ -0,0 +1,23 @@
...
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+  EnrollmentValidationWorkflow --local-scheduler \
+  --output-root $OUTPUT_ROOT/$INTERVAL \
+  --interval $INTERVAL \
+  --credentials $CREDENTIALS \
+  --n-reduce-tasks $NUM_REDUCE_TASKS
+
+pip install awscli
+aws s3 ls $OUTPUT_ROOT/$INTERVAL/ | grep "synthetic_enroll"  || true
```
brianhw: What does this line do?  Is it somehow making sure there are synthesized enrollment files present?   If not, will it skip the code below?

HassanJaveed84: From what I know of double pipe is that it will run ```true``` only if the previous command(grep in this case) fails. The code below would still be executed.

HassanJaveed84: We also have a post build task that checks for synthetic_enroll.log in console output. For edge.

brianhw: Oh, of course.   *That's* what this line is for -- just to get output into the console log.   Sorry to be slow.   It might be worth a comment on that line so I remember the next time I ask.

dataeng/resources/enrollment-validation-events.sh
```diff
@@ -0,0 +1,23 @@
...
+
+MAX_FILE_SIZE=`aws s3 ls $OUTPUT_ROOT/$INTERVAL/ | cut -b20-30 | grep -v PRE | sort -nr | head -1`
+if [ "$MAX_FILE_SIZE" -gt "$FILE_THRESHOLD" ]; then echo "OVER" "THRESHOLD"; fi
```
brianhw: This is going to break on edge, with or without a post-build step, or in any case where there are no synthetic events.  Maybe we can put an if-then-else here to handle if MAX_FILE_SIZE="" and FILE_THRESHOLD>="0" then do nothing.

dataeng/jobs/analytics/EnrollmentValidationEvents.groovy
brianhw: So I think we have to put something in the script.   As a result, I don't think we will need this.

dataeng/resources/enrollment-validation-events.sh
HassanJaveed84: I've removed the post build task that grepped for 'synthetic_enroll.log' for edge. Does it make sense to remove this as well ?

dataeng/resources/enrollment-validation-events.sh
```diff
@@ -0,0 +1,24 @@
...
+  --output-root $OUTPUT_ROOT/$INTERVAL \
+  --interval $INTERVAL \
+  --credentials $CREDENTIALS \
+  --n-reduce-tasks $NUM_REDUCE_TASKS
+
+pip install awscli
+aws s3 ls $OUTPUT_ROOT/$INTERVAL/ | grep "synthetic_enroll"  || true
+
+MAX_FILE_SIZE=`aws s3 ls $OUTPUT_ROOT/$INTERVAL/ | cut -b20-30 | grep -v PRE | sort -nr | head -1`
+if [ -z "$MAX_FILE_SIZE" ] && [ "$FILE_THRESHOLD" -gt "0" ]; then :
```
HassanJaveed84: @brianhw does this look reasonable ?

brianhw: I *think* it does the right thing, but I think it would be clearer if it didn't check for file threshold.    That is, I think it could just be:  if [ -z "$MAX_FILE_SIZE" ]; then :   which should be true if there are no synthetic events.  It shouldn't matter what the FILE_THRESHOLD value is.

HassanJaveed84: I've combined it into a single if else.

dataeng/resources/enrollment-validation-events.sh
HassanJaveed84: Removed.

PR #462, opened 2018-11-06
TE-2793 Clear quality reports from prior jobs

jmbowman:
Quality reports from previous jobs on the same worker were sometimes sticking around and being included in the results for later jobs.  Explicitly delete all these at the start of each quality job run.  The copying of artifacts from other jobs in the quality diff job occurs after this cleanup.


platform/jobs/edxPlatformQualityDiff.groovy
```diff
@@ -113,6 +116,10 @@ jobConfigs.each { jobConfig ->
             }
             timestamps()
             colorizeOutput('gnome-terminal')
+            preBuildCleanup {
+                includePattern(deleteReports)
```
estute: Am I reading this right? We 'include' the following paths in the 'delete'?

jmbowman: Yes, counter-intuitively enough: https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.wrapper.WrapperContext.preBuildCleanup

> Makes cleanup more selective by specifying file to be deleted using Ant syntax.

PR #464, opened 2018-11-07
Enterprise enrollment dsl

rao-abdul-mannan:


dataeng/jobs/analytics/EnterpriseEnrollment.groovy
```diff
@@ -0,0 +1,32 @@
...
+            authorization {
+                allVars.get('PROJECT_BASED_ACCESS').each { roles ->
+                    permission(roles.get('ROLE'), roles.get('GITHUB_ID'))
+                }
+            }
+            logRotator common_log_rotator(allVars)
+            parameters common_parameters(allVars)
+            parameters {
+                stringParam('REPORT_DATE', allVars.get('REPORT_DATE'), '')
+                stringParam('EXTRA_ARGS', allVars.get('EXTRA_ARGS'), '')
```
HassanJaveed84: So I haven't been including EXTRA_ARGS in any of the jobs, unless it wasn't already set to something. 
And I'm not exactly sure why we have these in the first place, perhaps to support adding parameters for custom runs(failed jobs?), overwrite_n_days etc ? @brianhw do you know more ?

brianhw: Yes, these were added to support adding parameters for custom runs on specific jobs.   Yes, overwriting a value of overwrite_n_days is one of those scenarios, for when we want to run with a value different from the default of 3 (either more or less).  But we hadn't been adding them on all jobs and I don't know offhand why this job would need it.   I suspect it was a cut-and-paste.

rao-abdul-mannan: Although we don't need it right now, but in case we have to run a job with different param so it can be handy & we don't have to manually change the job.

dataeng/jobs/analytics/EnterpriseEnrollment.groovy
```diff
@@ -0,0 +1,32 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+
+class EnterpriseEnrollment {
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("enterprise-enrollment") {
+            authorization {
+                allVars.get('PROJECT_BASED_ACCESS').each { roles ->
+                    permission(roles.get('ROLE'), roles.get('GITHUB_ID'))
```
brianhw: Oh, right.   We talked about changing this to be a double nested list.    That is, loop through IDs, and then loop through roles per id.

dataeng/jobs/analytics/Enterprise.groovy
```diff
@@ -0,0 +1,40 @@
...
+                }
+                multiscm common_multiscm(allVars)
+                triggers common_triggers(allVars)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    if (job == 'enrollment') {
+                        shell(dslFactory.readFileFromWorkspace('dataeng/resources/enterprise-enrollment.sh'))
+                    } else {
+                        shell(dslFactory.readFileFromWorkspace('dataeng/resources/enterprise-user.sh'))
```
brianhw: Instead of hardcoding knowledge of the file structure, can we just use the job name to create the shell name here?    e.g.   "enterprise-" + job + ".sh" ?   That way, if they add another workflow, it's clear what to do to extend it.

rao-abdul-mannan: Right, I've updated it.

PR #465, opened 2018-11-07
Move archiver job to Tools Jenkins

pwnage101:
PLAT-2380


devops/jobs/UserRetirementArchiver.groovy
```diff
@@ -0,0 +1,153 @@
...
+                            timeout(10)
+                        }
+                        cleanBeforeCheckout()
+                    }
+                }
+            }
+        
+            steps {
+                virtualenv {
+                    pythonName('System-CPython-3.5')
```
pwnage101: Lets treat ourselves a little because we deserve it.

pwnage101:
It's now ready for re-review.  I have since added a new commit which reflects my decision to pivot to a one-job-per-environment approach, away from our old trigger job approach.

src/main/groovy/org/edx/jenkins/dsl/UserRetirementConstants.groovy
```diff
@@ -0,0 +1,112 @@
+/*
```
pwnage101: The bulk of the DSL closures have moved into this shared class so that we don't have to re-type them for every single job.

platform/jobs/RetirementJobEdxTriggers.groovy
```diff
@@ -57,22 +57,6 @@ List jobConfigs = [
...
-        cron: '0 9 * * *',  // 10:00 UTC every day.
-        disabled: true
-    ],
-    [
-        downstreamJobName: 'user-retirement-archiver',
-        triggerJobNamePrefix: 'user-retirement-archiver-trigger',
-        environmentDeployment: 'prod-edge',
-        extraMembersCanBuild: [],
-        cron: '30 9 * * *',  // 10:30 UTC every day.
-        disabled: true
```
pwnage101: After merging this, I'll try to remember to go back and delete these old trigger jobs.

devops/jobs/UserRetirementArchiver.groovy
```diff
@@ -0,0 +1,80 @@
...
+class UserRetirementArchiver {
+    public static def job = { dslFactory, extraVars ->
+
+        // Job config assertions:
+        assert (extraVars.get('MAILING_LIST', '') instanceof String) :
+            "Make sure MAILING_LIST is a single string (containing a comma-delimited list) and not a proper Yaml " +
+            "list or some other non-string type."
+
+        def job_name = extraVars.get("FOLDER_NAME") + '/archiver-' + extraVars.get('ENVIRONMENT_DEPLOYMENT')
+        dslFactory.job(job_name) {
```
pwnage101: Note that i removed the `user-retirement-` prefix from the job name because now all user retirement jobs will be namespaced to the folder.  After expansion, this job should end up with a name like `User-Retirement/archiver-prod-edx`.

src/main/groovy/org/edx/jenkins/dsl/UserRetirementConstants.groovy
```diff
@@ -0,0 +1,115 @@
...
+
+            if (extraVars.containsKey('MAILING_LIST')) {
+                // Send an alerting email upon failure.
+                extendedEmail {
+                    recipientList(extraVars.get('MAILING_LIST'))
+                    triggers {
+                        failure {
+                            attachBuildLog(false)  // build log contains PII!
+                            compressBuildLog(false)  // build log contains PII!
+                            subject('Failed build: ' + job_name + ' #${BUILD_NUMBER}')
```
HassanJaveed84: nit: We can probably use JOB_NAME environment variable like you use BUILD_NUMBER and skip adding the method parameter altogether.

pwnage101:
@estute removal of throttling was offset by adding concurrent(false), and is motivated by the fact that tools-jenkins does not utilize any workers as it is a one-instance jenkins.  For the archiver, this is not a big deal because we would only build one archiver job per environment per day.

PR #471, opened 2018-11-13
Event export incremental large dsl

rao-abdul-mannan:


brianhw:
Just curious if this would be easier to define with the other event-export jobs.   If they change, this should change, and should only differ in the size of the cluster.   

Emilio and others in support also raised issues with being able to access jobs on the jenkins -- both the event export jobs and the weekly data export jobs.   We may need to bring in job-level permissions for them on those, similar to what was done for enterprise.  

Speaking of that, I can see it would be much simpler to administer if there were some way to be able to use some kind of group membership, rather than having to grant to individual github users.  I don't know if there is such a concept.  Out of scope for this work, but just raising it.

dataeng/jobs/analytics/EventExportIncrementalLarge.groovy
```diff
@@ -0,0 +1,45 @@
...
+                    stringParam('ONLY_ORGS', allVars.get('ONLY_ORGS'), "i.e. --org-id [\"FooX\",\"BarX\"]")
+                    stringParam('DATA_CZAR_KEYS_BRANCH', allVars.get('DATA_CZAR_KEYS_BRANCH'), '')
+                }
+                multiscm common_multiscm(allVars) >> secure_scm(allVars) >> data_czar_keys_scm(allVars)
+
+                triggers common_triggers(allVars, env_config)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    shell(dslFactory.readFileFromWorkspace("dataeng/resources/event-export-incremental-large-${environment}.sh"))
```
brianhw: Do we need separate scripts, or can we just use the existing scripts.  And for that matter, is there a reason that we don't just have one single script for all four event export jobs?   Seems that they differ only in the --environment variable and the ONLY_ORGS.   Which should be blank for the other jobs, so would work fine there with a blank default.

dataeng/resources/event-export-incremental-large-edge.sh
```diff
@@ -0,0 +1,36 @@
...
+
+env | sort
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+ EventExportTask --local-scheduler \
+ --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
+ --output-root ${OUTPUT_ROOT} \
+ --config ${EXPORTER_CONFIG_PATH} \
+ --gpg-key-dir ${GPG_KEYS_PATH} \
+ --environment edge ${ONLY_ORGS} \
```
brianhw: nit:   put ONLY_ORGS on a separate line.  This makes it look like it's an argument to --environment.

rao-abdul-mannan:
@brianhw There are couple of other differences with event-export-incremental jobs like master & worker instance types, hadoop env & mapreduce configuration & task params that you mentioned. I think seeding all 4 jobs with a single dsl script & vars file will introduce unnecessary complexity.

dataeng/jobs/analytics/EventExportIncrementalLarge.groovy
```diff
@@ -0,0 +1,46 @@
...
+                    stringParam('DATA_CZAR_KEYS_BRANCH', allVars.get('DATA_CZAR_KEYS_BRANCH'), '')
+                    stringParam('ENVIRONMENT', env_config.get('ENVIRONMENT'), '')
+                }
+                multiscm common_multiscm(allVars) >> secure_scm(allVars) >> data_czar_keys_scm(allVars)
+
+                triggers common_triggers(allVars, env_config)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    shell(dslFactory.readFileFromWorkspace("dataeng/resources/event-export-incremental-large.sh"))
```
brianhw: Do we need to use event-export-incremental-large.sh here?   It looks identical to event-export-incremental.sh.

dataeng/resources/event-export-incremental-large.sh
```diff
@@ -0,0 +1,37 @@
...
+env | sort
+
+${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
+ EventExportTask --local-scheduler \
+ --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
+ --output-root ${OUTPUT_ROOT} \
+ --config ${EXPORTER_CONFIG_PATH} \
+ --gpg-key-dir ${GPG_KEYS_PATH} \
+ --environment ${ENVIRONMENT} \
+ ${ONLY_ORGS} \
```
brianhw: nit:  thanks for doing this.   It would be good to do this in event-export-incremental.sh, especially if we end up using just that file.

dataeng/jobs/analytics/EventExportIncrementalLarge.groovy
rao-abdul-mannan: Good catch.

PR #476, opened 2018-11-15
Update users dsl

rao-abdul-mannan:


src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -69,6 +69,20 @@ class AnalyticsConstants {
...
 
+    public static def analytics_configuration_scm = { allVars ->
+        return {
+            git {
+                remote {
+                    url('$CONFIG_REPO')
+                    branch('$CONFIG_BRANCH')
+                }
+                extensions {
+                    pruneBranches()
```
brianhw: Is there a reason this doesn't match the extensions in common_configuration_scm?  

And just out of curiosity, is there a way to define common_configuration_scm in a way that uses this definition?   It would be a little DRYer.

PR #480, opened 2018-11-21
actually read and respect the DISABLED config var

pwnage101:
This fixes a bug which would cause all User-Retirement jobs to run according to their schedule regardless of whether we disabled them via config.

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/UserRetirementConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/UserRetirementConstants.groovy
@@ -109,4 +109,10 @@ class UserRetirementConstants {
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/edx/jenkins/dsl/UserRetirementConstants.groovy
```diff
@@ -109,4 +109,10 @@ class UserRetirementConstants {
             }
         }
     }
+    public static def common_closures_extra = { extraVars ->
+        return {
+            disabled(extraVars.get('DISABLED'))  // Jobs may be disabled for testing/rollout.
```
pwnage101: This line is the meat of this PR, the rest is just making the job more pretty.

PR #482, opened 2018-11-27
Deploy cluster DSL

HassanJaveed84:


HassanJaveed84:
Now uses EMR_EXTRA_VARS common definition.

PR #483, opened 2018-11-27
Add dsl for openedx-release-ci job

michaelyoungstrom:
Get started on the DSL for the new openedx-release-ci job. For now, it just allows you to specify a release-name and whether you want the created branches to be kept or deleted at the end of the job.

For now, it is only triggered manually. That will change as the job becomes more useful.

### DIFF
--- a/openedx/jobs/openedxReleaseJob.groovy
+++ b/openedx/jobs/openedxReleaseJob.groovy
@@ -1,0 +1,35 @@
+package openedx
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+// This is the job DSL responsible for creating the main pipeline job.
+pipelineJob('openedx-release-ci') {
+
+    definition {
+
+        parameters {
+            stringParam('OPENEDX_RELEASE_NAME', 'test-release',
+                        'open-release/${OPENEDX_RELEASE_NAME}')
+            choiceParam('DELETE_OR_KEEP', ['delete', 'keep'],
+                        'What should we do with the new release branches')
+
+        }
+
+        logRotator JENKINS_PUBLIC_LOG_ROTATOR(7)
+
+        cpsScm {
+            scm {
+                git {
+                    branch('master')
+                    remote {
+                        credentials('jenkins-worker')
+                        github('edx/repo-tools', 'ssh', 'github.com')
+                        refspec('+refs/heads/master:refs/remotes/origin/master')
+                    }
+                }
+            }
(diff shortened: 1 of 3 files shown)


openedx/jobs/openedxReleaseJob.groovy
```diff
@@ -0,0 +1,42 @@
...
+PrintStream out = config['out']
+
+// This is the job DSL responsible for creating the main pipeline job.
+pipelineJob('openedx-release-ci') {
+
+    definition {
+
+        parameters {
+            stringParam('OPENEDX_RELEASE_NAME', 'test-release',
+                        'open-release/${OPENEDX_RELEASE_NAME}')
```
nedbat: Most of this file I don't understand :)   Does this stringParam mean that the default will be "open-release/test-release"?

michaelyoungstrom: Yep! That's what the default is. Happy to change if you have a preference. Also, this is where the Jenkinsfile in repo-tools gets the branch name for its command.

openedx/jobs/openedxReleaseJob.groovy
```diff
@@ -0,0 +1,42 @@
+package openedx
+
+import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+
+/* stdout logger */
+Map config = [:]
+Binding bindings = getBinding()
+config.putAll(bindings.getVariables())
+PrintStream out = config['out']
```
estute: ^ setting up the stdout logger was pretty much only used when debugging issues with parsing the secret files during the seeding process, which if they were configured incorrectly, would cause the seeding fail without explanation. Seeing that you have no secret files, you can delete this. Also, the majority of the logic is in the Jenkinsfile itself, so you can get rid of the yaml import as well

openedx/jobs/openedxReleaseJob.groovy
```diff
@@ -0,0 +1,35 @@
...
+        logRotator JENKINS_PUBLIC_LOG_ROTATOR(7)
+
+        cpsScm {
+            scm {
+                git {
+                    remote {
+                        credentials('jenkins-worker')
+                        github('edx/repo-tools', 'ssh', 'github.com')
+                        refspec('+refs/heads/*:refs/remotes/origin/* +refs/pull/*:refs/remotes/origin/pr/*')
+                        branch('master')
```
jmbowman: Shouldn't `branch` be under `git` instead of `remote`?  And if we're only going to ever use master for this, can't we use a narrower `refspec` to speed up checkouts?

PR #485, opened 2018-11-30
Increase current exporter builds from 4 to 6.

brianhw:
This should be enough to get started on exporting MIT sooner.


brianhw:
This was blocked by https://openedx.atlassian.net/browse/DE-1131, because the load on the database was assumed to be coming from the research exports.   It turns out that it doesn't seem to.   (The load comes from the courseware-studentmodule dumps instead.)

coryleeio:
@brianhw I think based on @brianhw 's comments I'm good with this.  I'll put something in the event log and let the on-call know.

coryleeio:
to be clear I'm referring to https://openedx.atlassian.net/browse/DE-1131

PR #489, opened 2018-12-03
Many more fixes and add basic smoke test

pwnage101:
This adds a basic "smoke test" spec which simply seeds the job DSL in the test environment, as well as several additional fixes to get the test to pass.  The test is not at all exhaustive, but it facilitated a substantial leap-forward towards getting this job to work.

### DIFF
--- a/devops/jobs/UserRetirementArchiver.groovy
+++ b/devops/jobs/UserRetirementArchiver.groovy
@@ -33,7 +33,10 @@ class UserRetirementArchiver {
             "Make sure MAILING_LIST is a single string (containing a comma-delimited list) and not a proper Yaml " +
             "list or some other non-string type."
 
-        def job_name = extraVars.get("FOLDER_NAME") + '/archiver-' + extraVars.get('ENVIRONMENT_DEPLOYMENT')
+        def job_name = 'archiver-' + extraVars.get('ENVIRONMENT_DEPLOYMENT')
+        if (extraVars.containsKey("FOLDER_NAME")) {
+            job_name = extraVars.get("FOLDER_NAME") + '/' + job_name
+        }
         dslFactory.job(job_name) {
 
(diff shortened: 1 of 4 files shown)


src/test/groovy/devops/UserRetirementArchiverJobSpec.groovy
```diff
@@ -0,0 +1,45 @@
...
+
+        setup:
+        JenkinsJobManagement jm = new JenkinsJobManagement(System.out, [:], new File('.'))
+        loader = new DslScriptLoader(jm)
+
+        when:
+        File dslTestSeederScriptPath = new File(
+            "src/test/resources/devops/seeders/UserRetirementArchiverTestSeeder.groovy"
+        )
+        GeneratedItems generatedItems = loader.runScript(dslTestSeederScriptPath.text)
```
pwnage101: FYI @estute this ended up not needing any string concatenation.  The class path resolution error was caused by the test seeder file being located where the testing framework expected spec files, so It was collecting it and trying to run it like a spec.  I simply moved it under `resources` and the error went away.

devops/jobs/UserRetirementArchiver.groovy
```diff
@@ -33,7 +33,10 @@ class UserRetirementArchiver {
             "Make sure MAILING_LIST is a single string (containing a comma-delimited list) and not a proper Yaml " +
             "list or some other non-string type."
 
-        def job_name = extraVars.get("FOLDER_NAME") + '/archiver-' + extraVars.get('ENVIRONMENT_DEPLOYMENT')
+        def job_name = 'archiver-' + extraVars.get('ENVIRONMENT_DEPLOYMENT')
+        if (extraVars.containsKey("FOLDER_NAME")) {
```
fredsmith: :+1:

PR #562, opened 2018-12-17
Combine pipeline DSL files for platform CI jobs

michaelyoungstrom:
These jobs are too redundant to create 1 file per job. I propose using maps, similar to what we do currently with build flow jobs, to keep it DRY.


platform/jobs/pipelines/edxPlatformPipelineMaster.groovy
```diff
@@ -0,0 +1,79 @@
...
+                                sparseCheckoutPaths {
+                                    sparseCheckoutPath {
+                                        path(jobConfig.jenkinsFileDir)
+                                    }
+                                }
+                            }
+                        }
+                        remote {
+                            credentials('jenkins-worker')
+                            github('edx/edx-platform', 'ssh', 'github.com')
```
jmbowman: We probably also want to mimic the ability of the build flow jobs to parameterize the repository name and associated variables.

PR #566, opened 2018-12-20
Define EXTRA_ARGS as a common parameter.

brianhw:


brianhw:
Some scripts have ${} and others just use $.   So I tried to make EXTRA_ARGS consistent with the other arguments in those scripts.

HassanJaveed84:
Makes sense. I'm good with this.

PR #569, opened 2018-12-21
inject environment name in user retirement archiver job

pwnage101:
This variable was left behind when we pivoted away from the "trigger
job" model which supplied this variable via a string parameter.


pwnage101:
@estute they're defined here: https://github.com/edx/edx-internal/blob/master/tools-edx-jenkins/user-retirement.yml#L29

And then get passed into the seed job at build time, which in turn passes them into this job DSL.

PR #571, opened 2018-12-21
add ghprb stub for edx-app-android job

estute:
This is a simple job for calling the Jenkinsfile in edx/edx-app-android on PR activity. I just want to get this in place, so that we can iterate on the CI process.


mobileApp/jobs/edxAppAndroidPr.groovy
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        cpsScm {
+            scm {
+                git {
+                    remote {
+                        credentials('jenkins-worker')
+                        github('edx/edx-app-android', 'https', 'github.com')
+                        refspec('+refs/heads/*:refs/remotes/origin/* +refs/pull/*:refs/remotes/origin/pr/*')
```
jmbowman: Do we need this broad of a refspec?  Seems like we could get away with just the current PR branch, like https://github.com/edx/jenkins-job-dsl/blob/master/platform/jobs/pipelines/edxPlatformPipelinePr.groovy#L116

estute: done

PR #572, opened 2018-12-27
Add dsl job to delete merged git branches (OPS-3223)

nadeemshahzad:


devops/resources/delete-merged-git-branches.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+WHITELIST="open-release|origin/release$|olive|mckinsey"
```
feanil: `mckinsey` is not needed here.

devops/resources/delete-merged-git-branches.sh
```diff
@@ -0,0 +1,14 @@
...
+
+for branch in $(git branch -r --merged origin/master | grep -v HEAD | grep -v origin/master | grep -vE "${WHITELIST}"); do
```
feanil: Can you add some comments in this file to explain what all the git commands are doing?  Not everyone will know that this is listing all branches that have been merged to origin/master for example.

feanil:
re-based to re-run tests since there looks like there was an upstream change to codenarc.

PR #576, opened 2019-01-02
Rename mckinsey bastion jobs

jdmulloy:


feanil:
Rebased since it looks like some things were changed with codenarc that  weren't picked up here.

PR #579, opened 2019-01-07
Add new jobs for missing_newrelic_alerts

syedimranhassan:


devops/jobs/MissingNewRelicAlerts.groovy
```diff
@@ -0,0 +1,63 @@
...
+                        credentialsBinding {
+                            string("NEW_RELIC_API_KEY", "${deployment}-new-relic-api-key")
+                            file("AWS_CONFIG_FILE","tools-edx-jenkins-aws-credentials")
+                            def variable = "${deployment}-missing-newrelic-alerts"
+                            string("ROLE_ARN", variable)
+                        }
+                    }
+
+                    triggers {
+                        cron("0 0 * * 0")
```
coryleeio: Every night midnight utc on Sunday is that intentional?


I would atleast change the first 0 to an H
eg:
'H 0 * * 0'

This will distribute it to an arbitrary minute within the hour based on a hash and is a good practice for scheduling jenkins jobs so they dont all stack up.

devops/resources/missing-newrelic-alerts.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+cd $WORKSPACE/configuration/util/jenkins/
+
+pip install -r check_celery_progress/requirements.txt
```
coryleeio: I would probably generate your own requirements.txt and put your script and requirements directory in their own directory instead of using the check_celery_progress/requirements.txt

This is so that if you want to delete one of the scripts later you dont have to pull them apart.

PR #583, opened 2019-01-11
Fix git checkout for private jobs

michaelyoungstrom:
Some small fixes. Makes sure the git checkout is correct for private pipelines, changes the branch for js/a11y private merge jobs, adds a private view for jobs.

### DIFF
--- a/platform/views/platformViews.groovy
+++ b/platform/views/platformViews.groovy
@@ -56,3 +56,12 @@ listView("edx-platform-pipeline-pr-tests") {
     }
     columns DEFAULT_VIEW.call()
 }
+
+listView("edx-platform-private-tests") {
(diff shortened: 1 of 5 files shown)


michaelyoungstrom:
jenkins run all

platform/views/platformViews.groovy
```diff
@@ -56,3 +56,12 @@ listView("edx-platform-pipeline-pr-tests") {
     }
     columns DEFAULT_VIEW.call()
 }
+
+listView("edx-platform-private-tests") {
+    description('Jobs for running tests on edx-platform-private')
+
+    jobs {
+        regex("edx-platform-*_private")
```
estute: can't recall, do we have private hawthorn/ironwood jobs? if so, will they match this regex?

PR #590, opened 2019-01-23
Make bastion access run automatically OPS-3656

jdmulloy:


devops/jobs/BastionAccessWatcher.groovy
```diff
@@ -0,0 +1,97 @@
+/*
+
+ This job checks out the app permissions repository and checks for merges to master with github webhooks. Upon a merge to master,
```
feanil: Update the comment since this is not the app permissions job.

PR #597, opened 2019-02-04
Add openedx.yaml file

michaelyoungstrom:


openedx.yaml
```diff
@@ -0,0 +1,7 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# https://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html#specification
+
+oeps: {}
+owner: edx/testeng
+tags: [tools]
+track-pulls: true
```
nedbat: There's no need for the track-pulls key any more.  Please delete it.

PR #603, opened 2019-02-11
Delete build flow jobs

michaelyoungstrom:
Removing in favor of pipelines


platform/views/platformViews.groovy
```diff
@@ -46,11 +46,11 @@ branchList.each { branch ->
 
 }
 
-listView("edx-platform-pipeline-pr-tests") {
+listView("edx-platform-pr-tests") {
```
jmbowman: A view of this name already exists, created by iteration over `branchList` above.  Do we need to keep this separate one anymore?

michaelyoungstrom: Good point. Removed. No need to separate anymore

PR #613, opened 2019-02-22
Update jobs to point to new repo

syedimranhassan:


syed-awais-ali:
@feanil yeah, I agree. I suggested this initially but then in standup jenkins-job-dsl was referred for moving this. I think we should move this to internal repo instead of jenkins-job-dsl repo.

PR #615, opened 2019-02-26
Update job to send email only on failure

syedimranhassan:


devops/jobs/MissingRDSAlarms.groovy
```diff
@@ -63,11 +63,26 @@ class  MissingRDSAlarms {
                             )
                         }
                     }
+                   publishers {
+                        extendedEmail {
+                            recipientList(extraVars.get('NOTIFY_ON_FAILURE'))
+                            contentType('text/plain')
+                            defaultContent('''\
+                                  $BUILD_LOG
+                                  '''.stripIndent())
```
coryleeio: Instead of emailing the build log, which may contain sensitive data - Just notify us that the job is failing, we dont need the whole log in email, a link to the job should be sufficient.

PR #624, opened 2019-03-06
Use mysql rename for video-timeline and enrollments.

HassanJaveed84:
This will bring down the runtime of both workflows by 6-7 hours.

### DIFF
--- a/dataeng/resources/video-timeline.sh
+++ b/dataeng/resources/video-timeline.sh
@@ -8,4 +8,5 @@ ${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
   InsertToMysqlAllVideoTask --local-scheduler \
   --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
   --n-reduce-tasks $NUM_REDUCE_TASKS \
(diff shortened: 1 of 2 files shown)


dataeng/resources/video-timeline.sh
```diff
@@ -8,4 +8,5 @@ ${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
   InsertToMysqlAllVideoTask --local-scheduler \
   --interval $(date +%Y-%m-%d -d "$FROM_DATE")-$(date +%Y-%m-%d -d "$TO_DATE") \
   --n-reduce-tasks $NUM_REDUCE_TASKS \
+  --InsertToMysqlVideoTimelineTask-use-temp-table-for-overwrite \
```
pwnage101: I see that there's a `use_temp_table_for_overwrite` luigi parameter in `MysqlInsertTaskMixin`, but how exactly does `--InsertToMysqlVideoTimelineTask-use-temp-table-for-overwrite` translate to that parameter?

HassanJaveed84: You can set parameters for other classes by using that notation. More here:
https://luigi.readthedocs.io/en/stable/parameters.html#setting-parameter-value-for-other-classes

PR #626, opened 2019-03-07
Added missing NOTIFY param.

HassanJaveed84:


brianhw:
Is this (not) ready to merge?

HassanJaveed84:
I don't recall our discussion from earlier around this. But we could get daily alerts if we put this in(it fails everyday). Should we increase the threshold ? It is currently set to 600.

brianhw:
Fair question.   There are probably a few things that could be improved with this.  One is that there is a huge amount of noise in what it echoes to console that is a lot of debugging stuff that obscures the list of cluster names and their estimated cost.   Second is that this doesn't handle instance fleets at all -- and I have no idea if we ever really *use* instance fleets alternatives in any workflows.  Third is that there is no distinction made between regular operations and development/experimentation work, so it's hard to know what the overall limit should be set to.   That said, I agree that at the very least, we should pick a higher number if this is always failing.   But with Snowflake work (at a minimum), we may be going over the limit anyway in the short term, so perhaps we just have to suck that up.   Oh, wait:   I think that we should only notify the email list, and not page on failures on this job.

PR #627, opened 2019-03-11
Ihassan/ops 3010 slow query to cloudwatch

syedimranhassan:


devops/jobs/ExportRDSSlowQueryLogs.groovy
```diff
@@ -0,0 +1,68 @@
+package devops.jobs
+import static org.edx.jenkins.dsl.Constants.common_wrappers
+
+class EXPORTRDSSLOWQUERYLOGS {
```
coryleeio: The casing of this should match the jenkins-job-dsl-internal repo PR

coryleeio: The casing of this should match the jenkins-job-dsl-internal repo PR

devops/jobs/ExportRDSSlowQueryLogs.groovy
```diff
@@ -0,0 +1,68 @@
...
+                            string("USERNAME", "${deployment}-export-slow-logs-username")
+                            string("PASSWORD", "${deployment}-export-slow-logs-password")
+                            file("AWS_CONFIG_FILE","tools-edx-jenkins-aws-credentials")
+                            def variable = "${deployment}-export-slow-query-logs"
+                            string("ROLE_ARN", variable)
+                        }
+                    }
+
+                    triggers {
+                        cron("H * * * *")
```
coryleeio: Since your script looks back an hour in the logs, this should not run every minute.  It should probably run every 55 minutes or so.  This is because we want to minimize overlap while not losing anything.

coryleeio: Let me know if you need more context on this

devops/resources/export-slow-query-logs.sh
```diff
@@ -0,0 +1,13 @@
...
+
+pip install -r requirements.txt
+. ../assume-role.sh
+
+# Assume the role
+set +x
+assume-role ${ROLE_ARN}
+set -x
+
+python export_slow_query_logs.py  --username ${USERNAME} --password ${PASSWORD}
```
coryleeio: Please load password from an ENV, passing it in this way will expose it to anyone who runs ps aux on the machine(possibly other jobs)

PR #631, opened 2019-03-13
Ihassan/ops 3026 table size monitoring

syedimranhassan:


devops/jobs/CheckTableSize.groovy
```diff
@@ -0,0 +1,68 @@
...
+                    }
+
+                    triggers {
+                        cron("H * * * *")
+                    }
+
+                    environmentVariables {
+                        env('AWS_DEFAULT_REGION', extraVars.get('REGION'))
+                        env('USERNAME', extraVars.get('USERNAME'))
+                        env('PASSWORD', configuration.get('PASSWORD'))
```
feanil: You should pull this from the credential binding.

devops/resources/table-size-monitoring.sh
```diff
@@ -0,0 +1,19 @@
...
+assume-role ${ROLE_ARN}
+set -x
+
+# Set RDSTHRESHOLD if not set in job, need because we're setting -u
+# Otherwise we get an error "RDSTHRESHOLD: unbound variable"
+if [[ ! -v RDSTHRESHOLD ]]; then
+    RDSTHRESHOLD=""
+fi
+
+python check_table_size.py  --username ${USERNAME} --password ${PASSWORD} --threshold ${THRESHOLD} ${RDSTHRESHOLD}
```
coryleeio: I would load the password from an environment variable instead of expecting it as a passed arg, as this makes it known to anyone on the system who can inspect the running processes.

coryleeio: aka other jenkins tasks

devops/jobs/CheckTableSize.groovy
```diff
@@ -0,0 +1,77 @@
...
+                            string("ROLE_ARN", variable)
+                        }
+                    }
+
+                    triggers {
+                        cron("H * * * *")
+                    }
+
+                    def rdsthreshold = ""
+                    redis_config.rds.each { rds, threshold ->
```
coryleeio: I dont see where redis_config is coming from, does it exist?

coryleeio: is it possible to just load this from the config?

PR #634, opened 2019-03-18
Initial username replacment  work

tuchfarber:


tuchfarber:
Two points I was concerned about:
 - Line 12, manually adding the prod.yml and stage.yml files via devops ticket. This is how I understood it from our meeting. Just wanted to confirm this was correct.
 - The authorization. I didn't entirely understand the permissioning system. I imagine this needs to be changed. I'm just not sure what to. 
```
authorization {
    blocksInheritance(true)
    permissionAll('edx')
}
```

devops/jobs/ReplaceUsernames.groovy
```diff
@@ -0,0 +1,97 @@
...
+class ReplaceUsernames {
+    public static def job = { dslFactory, extraVars ->
+        assert extraVars.containsKey('ENVIRONMENTS') : "Please define ENVIRONMENTS. It should be a list of strings."
+        assert !(extraVars.get('ENVIRONMENTS') instanceof String) : "Make sure ENVIRONMENTS is a list and not a string"
+
+        extraVars.get('ENVIRONMENTS').each { environment ->
+            dslFactory.job("${environment}-replace-usernames") {
+                /* TODO: Find out if this is correct */
+                /* Enable project based security */
+                authorization {
```
feanil: Security is handled at the folder level so you don't need this block at all.

devops/jobs/ReplaceUsernames.groovy
```diff
@@ -0,0 +1,97 @@
...
+                            cloneOptions {
+                                shallow()
+                                timeout(10)
+                            }
+                            cleanBeforeCheckout()
+                        }
+                    }
+                }
+
+                logRotator JENKINS_PUBLIC_LOG_ROTATOR()
```
feanil: If this is going to run on tools-jenkins, then you should use the common wrappers and logging config.

eg. https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/CheckSesLimits.groovy#L41-L42

devops/resources/replace-usernames.sh
```diff
@@ -0,0 +1,26 @@
...
+
+mkdir -p $WORKSPACE/username-replacement-secure
+
+# prepare tubular
+cd $WORKSPACE/tubular
+pip install -r requirements.txt
+
+# Call the script to replace the usernames for all users in the CSV file.
+python scripts/replace_usernames.py \
+    --config_file=$WORKSPACE/username-replacement-secure/$ENVIRONMENT.yml # TODO: DEVOPS DROPS FILE HERE?
```
feanil: Rather than doing this, you'll want to do something like this: https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/CheckSesLimits.groovy#L45-L46

In that example a file that we upload to jenkins will be available at a random path and the location of the path will be available as the environment variable `AWS_CONFIG_FILE`  You'll want to do something similar here.

devops/jobs/ReplaceUsernames.groovy
tuchfarber: Removed

devops/jobs/ReplaceUsernames.groovy
tuchfarber: Fixed to use the common ones

devops/resources/replace-usernames.sh
tuchfarber: Changed to match that pattern. If there is to be a separate config file for each environment, like I had with `$ENVIRONMENT.yml`, do I need to create two `credentialsBinding` => `file` objects, or can a different file be at that location for each job. I.e. is the new `tools-edx-jenkins-username-replacement-config` global or per job?

devops/jobs/ReplaceUsernames.groovy
```diff
@@ -0,0 +1,91 @@
...
+
+*/
+class ReplaceUsernames {
+    public static def job = { dslFactory, extraVars ->
+        assert extraVars.containsKey('ENVIRONMENTS') : "Please define ENVIRONMENTS. It should be a list of strings."
+        assert !(extraVars.get('ENVIRONMENTS') instanceof String) : "Make sure ENVIRONMENTS is a list and not a string"
+
+        extraVars.get('ENVIRONMENTS').each { environment ->
+            dslFactory.job("${environment}-replace-usernames") {
+                /* TODO: Find out if this is correct */
```
feanil: Remove these comments.

devops/jobs/ReplaceUsernames.groovy
```diff
@@ -0,0 +1,91 @@
...
+                    stringParam('TUBULAR_BRANCH', 'master', 'Repo branch for the tubular scripts.')
+                    fileParam(
+                        'username_replacements.csv',
+                        'A CSV of [current_username,desired_username] for username replacement'
+                    )
+                }
+
+                wrappers {
+                    credentialsBinding {
+                        file('USERNAME_REPLACEMENT_CONFIG_FILE','tools-edx-jenkins-username-replacement-config')
```
feanil: The credential names are unique for all of jenkins so you'll want something more like what we do here: https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/CheckSesLimits.groovy#L47-L48

devops/resources/replace-usernames.sh
```diff
@@ -0,0 +1,26 @@
...
+# just stores a filename in the environment (rather than the content).
+env
+
+# Make sure that when we try to write unicode to the console, it
+# correctly encodes to UTF-8 rather than exiting with a UnicodeEncode
+# error.
+export PYTHONIOENCODING=UTF-8
+export LC_CTYPE=en_US.UTF-8
+
+mkdir -p $WORKSPACE/username-replacement-secure
```
feanil: Delete this.

devops/resources/replace-usernames.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# prepare tubular
+cd $WORKSPACE/tubular
+pip install -r requirements.txt
+
+# Call the script to replace the usernames for all users in the CSV file.
+python scripts/replace_usernames.py \
+    --config_file=$USERNAME_REPLACEMENT_CONFIG_FILE # TODO: DEVOPS DROPS FILE HERE?
```
feanil: Remove the comment.

devops/resources/replace-usernames.sh
feanil: See new comment, these names are global so you should have a different one pre environment.

devops/jobs/ReplaceUsernames.groovy
tuchfarber: Removed

devops/jobs/ReplaceUsernames.groovy
tuchfarber: Somehow totally missed that on the first read through. Thanks!

devops/resources/replace-usernames.sh
tuchfarber: done

devops/resources/replace-usernames.sh
tuchfarber: done

devops/jobs/ReplaceUsernames.groovy
```diff
@@ -0,0 +1,89 @@
...
+/* JenkinsPublicConstants can be found in src/main/groovy/org/edx/jenkins/dsl*/
+/* It contains values and methods that are used repeatedly through the dsl jobs */
+import org.yaml.snakeyaml.Yaml
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_LOG_ROTATOR
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_WORKER
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.GENERAL_SLACK_STATUS
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.JENKINS_PUBLIC_GITHUB_BASEURL
+
+/*
+Need to manually add files with the following details to $WORKSPACE/username-replacement-secure/$ENVIRONMENT.yml:
```
georgebabey: Is it common to have to manually update a yml file for a jenkins job? Seems odd to me but maybe this is a pattern i just haven't seen

feanil: We don't currently have a good automation story for getting secrets available on jenkins while keeping them secure.  We use the jenkins secureBindings for this but don't have a way of adding new secure content automatically.

feanil: @tuchfarber can you update this comment to indicate that these are the contents of the credentialBinding.

PR #635, opened 2019-03-18
Remove ansible timing log from sandbox builds

coryleeio:


coryleeio:
~217 GB of our Jenkins space is these logs. 1.7GB per run
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/

We hold them for 5 days which means that space available can vary pretty radically depending on how many sandboxes have been built that week.

PR #641, opened 2019-03-25
Monitor bigquery event loading DSL

rao-abdul-mannan:


dataeng/jobs/analytics/MonitorBigqueryEventLoading.groovy
```diff
@@ -0,0 +1,34 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+
+class MonitorBigqueryEventLoading {
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("monitor-bigquery-loading") {
+            logRotator common_log_rotator(allVars)
+            parameters {
+                stringParam('TOOLS_REPO', allVars.get('TOOLS_REPO_URL'), '')
+                stringParam('TOOLS_BRANCH', allVars.get('TOOLS_BRANCH'), 'e.g. tagname or origin/branchname')
+                stringParam('NOTIFY', '$PAGER_NOTIFY', 'Space separated list of emails to send notifications to.')
```
doctoryes: Is this value used? I see that `common_publishers` is no longer imported at the top of the file.

dataeng/jobs/analytics/MonitorBigqueryEventLoading.groovy
```diff
@@ -0,0 +1,34 @@
...
+                stringParam('PARTITION_DATE', allVars.get('PARTITION_DATE'), 'DATE to check bigquery event loading.')
+                stringParam('CREDENTIALS', allVars.get('CREDENTIALS'))
+                stringParam('DATASET', allVars.get('DATASET'))
```
doctoryes: Is this one used as well?

dataeng/resources/monitor-bigquery-loading.sh
```diff
@@ -0,0 +1,19 @@
...
+. $TOOLS_BIN/activate
+
+pip install -r ${WORKSPACE}/analytics-tools/bigquery-monitoring/requirements.txt
+
+# download credentials file to temporary directory
+aws s3 cp $CREDENTIALS /tmp/bigquery_credentials.json
+
+export PARTITION_DATE=$(date +%Y-%m-%d -d "$PARTITION_DATE")
+
+. ${WORKSPACE}/analytics-tools/bigquery-monitoring/bigquery_monitor_event_loading.sh
```
doctoryes: This script doesn't yet exist in a PR, correct?

pwnage101: should be here: https://github.com/edx/analytics-tools/pull/13/files#diff-c5790760e0c72ae4fd6039c8137f2e00

pwnage101: That's also where `$DATASET` is used.

dataeng/jobs/analytics/MonitorBigqueryEventLoading.groovy
pwnage101: Definitely seems unused, I'll remove.  IIRC, the plan was to alarm via cloudwatch threshold.

dataeng/resources/monitor-bigquery-loading.sh
```diff
@@ -0,0 +1,19 @@
...
+mkdir -p $VENV_ROOT
+
+virtualenv $VENV_ROOT/analytics-tools
+TOOLS_BIN=$VENV_ROOT/analytics-tools/bin
+. $TOOLS_BIN/activate
+
+pip install -r ${WORKSPACE}/analytics-tools/bigquery-monitoring/requirements.txt
+
+# download credentials file to temporary directory
+aws s3 cp $CREDENTIALS /tmp/bigquery_credentials.json
```
pwnage101: I don't like that creds are being copied into /tmp, then not cleaned up.  I'm going to rework this.

HassanJaveed84: The script that I used to load data into snowflake also required pulling credentials info off a S3 url.
I used boto to read the file contents from the s3 file, maybe we can do the same and pull the object in the python script. I'm also wondering if we need multiple script files.

pwnage101: If we wanted to just make it one script file in the analytics-tools PR (just .py instead of .py + .sh) then we'd also need to leverage boto to do the `aws cloudwatch put-metric-data` step in addition to the `aws s3 cp`.  Consolidating into one python script is also my preference---any objection?

pwnage101:
This is ready for review again, as well as:

https://github.com/edx-ops/analytics-secure/pull/231
https://github.com/edx/analytics-tools/pull/13

dataeng/jobs/analytics/MonitorBigqueryEventLoading.groovy
brianhw: It is used.   But is it needed?

PR #643, opened 2019-03-28
Add jenkins job for make upgrade

michaelyoungstrom:
Add job and view for a new make upgrade job. Also add a new env var to the db cache job.

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -1,0 +1,69 @@
+package testeng
+
+Map edxPlatform = [
+    org: 'edx',
+    repoName: 'edx-platform',
+    cronValue: '@daily'
+]
+
+List jobConfigs = [
+    edxPlatform
+]
+
+/* Iterate over the job configurations */
+jobConfigs.each { jobConfig ->
+
+    job("${jobConfig.repoName}-upgrade-python-requirements") {
+
+        logRotator {
+            daysToKeep(14)
+        }
+        concurrentBuild(false)
+        label('jenkins-worker')
+        environmentVariables(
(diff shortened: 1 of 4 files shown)


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -0,0 +1,66 @@
...
+                }
+                branch('youngstrom/update-github-scripts')
+                extensions {
+                    cleanBeforeCheckout()
+                    relativeTargetDirectory('testeng-ci')
+                }
+            }
+        }
+        triggers {
+            cron("@daily")
```
jmbowman: Can the frequency be added as a parameter in `jobConfigs`?  We'll definitely want this to vary across repositories, depending on how often we modify them.

testeng/views/testengViews.groovy
```diff
@@ -0,0 +1,17 @@
+package testeng
+
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.DEFAULT_VIEW
+
+listView("upgrade-python-requirements") {
```
jmbowman: Thanks for thinking of this, this should come in handy.  We should also prepare a GitHub search URL that returns all open python dependency upgrade PRs across repos in the edX organization.

testeng/resources/upgrade-python-requirements.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+virtualenv upgrade_venv -q
```
jmbowman: Do we need to explicitly delete this virtualenv when done?  Seems like it could interfere with later jobs on the same worker.

michaelyoungstrom: We don't seem to have any existing jobs that do that but that's a good idea.

PR #646, opened 2019-04-03
disable lettuce jobs

estute:
All of the lettuce tests have now been converted to bokchoy. This will just disable the PR/Master jobs from running automatically (we will also need to change the settings for the expected contexts in the platform repo itself). Once this is done, we can begin work on removing the lettuce infrastructure from our code


jmbowman:
@estute This shouldn't be needed anymore, right?

estute:
Correct

PR #648, opened 2019-04-09
Update job status to PR

syedimranhassan:


devops/jobs/AppPermissionsRunner.groovy
```diff
@@ -97,6 +98,24 @@ class AppPermissionsRunner {
                         env('USER', extraVars.get('USER', 'ubuntu'))
                     }
 
+                    triggers {
+                        githubPullRequest {
```
coryleeio: Wont this cause the app permissions runner to run for every PR?
We don't necessarily want to run the job for unreviewed PRs, but perhaps I misunderstand what this is doing.

syedimranhassan: @coryleeio Yes, you are right, it should run once it merge. I am working to fix this. I have tested this and now working on DSL to update this PR

PR #672, opened 2019-04-25
Seed job for devstack snapshot installers

jmbowman:
Seed job to be able to build an offline devstack installer on demand.  Takes the devstack repo branch as a parameter to enable iteration on changes in devstack PRs; doesn't really support installers for named Open edX releases, but we haven't had demand for that yet either.


michaelyoungstrom:
Looks good. Do we want this to run on a regular schedule, or just on manual triggers for now?

jmbowman:
Just manual triggers; it's a pretty intensive process with large artifacts, don't really want to run it except when needed.  Just can't leave it to the last minute before an event, in case something broke since the last time it ran.  Might switch it to weekly or such later to avoid bit rot.

michaelyoungstrom:
Cool, makes sense to me.

PR #674, opened 2019-04-29
Add more 'make upgrade' jobs

jmbowman:
Add weekly "make upgrade" jobs to the rest of the Tools-managed repositories complying with OEP-18.


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -9,6 +9,14 @@ package testeng
 //     githubTeamReviewers: Comma separated list of Github teams that should be tagged on PR's, e.g.: team1,team2,team3
 // ]
 
+Map bokChoy = [
```
michaelyoungstrom: Nit: I feel like `bokchoy` is pretty commonly used by us and might be simpler rather than the camel case. Up to you

PR #678, opened 2019-05-02
Add more repos to `make upgrade` automation.

feanil:
Add two more repos to the make upgrade automation.

@davestgermain I added you to the list for edx-proctoring since you seem to be a recent contributor to this and probably have more context than me.  Let me know if it makes more sense to have a github team associated with this and what that should be.

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -49,12 +49,28 @@ Map djangoConfigModels = [
     githubTeamReviewers: 'testeng'
 ]
 
+Map edxOrganizations = [
+    org: 'edx',
+    repoName: 'edx-organizations',
+    cronValue: '@weekly',
+    githubUserReviewers: 'feanil',
+    githubTeamReviewers: 'testeng'
+]
+
 Map edxPlatform = [
     org: 'edx',
     repoName: 'edx-platform',
     cronValue: '@daily',
     githubUserReviewers: '',
     githubTeamReviewers: 'testeng'
+]
+
+Map edxProctoring = [
+    org: 'edx',
+    repoName: 'edx-proctoring',
+    cronValue: '@weekly',
+    githubUserReviewers: 'feanil',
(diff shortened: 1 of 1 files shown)


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -57,6 +65,14 @@ Map edxPlatform = [
     githubTeamReviewers: 'testeng'
 ]
 
+Map edxProctoring = [
+    org: 'edx',
+    repoName: 'edx-proctoring',
+    cronValue: '@weekly',
+    githubUserReviewers: 'feanil,davestgermain',
```
davestgermain: You're going to change this to dahlia, right?

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -49,6 +49,14 @@ Map djangoConfigModels = [
     githubTeamReviewers: 'testeng'
 ]
 
+Map edxOrganizations = [
+    org: 'edx',
+    repoName: 'edx-organizations',
+    cronValue: '@weekly',
+    githubUserReviewers: 'feanil',
+    githubTeamReviewers: 'testeng'
```
jmbowman: I realize that the team owner for these two repos is unclear, but we should really start figuring that out so testeng doesn't become the dumping ground for all abandoned repos.

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -57,6 +65,14 @@ Map edxPlatform = [
     githubTeamReviewers: 'testeng'
 ]
 
+Map edxProctoring = [
+    org: 'edx',
+    repoName: 'edx-proctoring',
+    cronValue: '@weekly',
+    githubUserReviewers: 'feanil,davestgermain',
+    githubTeamReviewers: 'testeng'
+]
```
jmbowman: Both of the new maps still need to be added to `jobConfigs`.

PR #680, opened 2019-05-03
check primary keys

syed-awais-ali:


syed-awais-ali:
This was stale. Made https://github.com/edx/jenkins-job-dsl/pull/714 for this work

PR #686, opened 2019-05-07
Add email notifications on make upgrade failures

michaelyoungstrom:


michaelyoungstrom:
@feanil Is this better?

feanil:
@michaelyoungstrom I was thinking more like:

```
Map bokchoy = [
    org: 'edx',
    repoName: 'bok-choy',
    cronValue: '@weekly',
    githubUserReviewers: '',
    githubTeamReviewers: ['testeng', 'team2', 'team3'],
    emailList: ['testeng@edx.org'],
]

And then you can join the lists via commas or spaces depending on how the other function expects inputs.
```

michaelyoungstrom:
@feanil Ahhh totally, sorry for misunderstanding. How's this look?

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -5,80 +5,90 @@ package testeng
 //     org: Github organization,
 //     repoName: Github repository,
 //     cronValue: How often to run the job,
-//     githubUserReviewers: Comma separated list of Githuhb users that should be tagged on PR's, e.g.: user1,user2,user3
-//     githubTeamReviewers: Comma separated list of Github teams that should be tagged on PR's, e.g.: team1,team2,team3
+//     githubUserReviewers: List of Github users that should be tagged on PR's, e.g.: user1,user2,user3
+//     githubTeamReviewers: List of Github teams that should be tagged on PR's, e.g.: team1,team2,team3
+//     emails: List of emails that should be notified when job fails
```
estute: I would suggest updating the example in the comment here, because you changed the format. Either don't include examples (and make people look at existing structures) or use the actual type used for the field. People might get confused here and think they should put a comma-separated string in. Also, if you choose to put examples in, you should do it for emails as well.

PR #700, opened 2019-05-22
Add USER_ROLES for per-job authorization.

brianhw:
Permit data-science and other users to run single SQL scripts.

### DIFF
--- a/dataeng/jobs/analytics/SqlScripts.groovy
+++ b/dataeng/jobs/analytics/SqlScripts.groovy
@@ -37,6 +37,13 @@ class SqlScripts {
 
     public static def single_script_job = { dslFactory, allVars ->
         dslFactory.job("single-sql-script") {
+            authorization {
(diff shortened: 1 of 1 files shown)


dataeng/jobs/analytics/SqlScripts.groovy
```diff
@@ -37,6 +37,13 @@ class SqlScripts {
 
     public static def single_script_job = { dslFactory, allVars ->
         dslFactory.job("single-sql-script") {
+            authorization {
+                allVars.get('USER_ROLES').each { github_id, roles ->
+                    roles.each {
+                        permission(it, github_id)     // it is an implicit parameter corresponding to the current element
```
pwnage101: Nit: can't you just do
```
roles.each { role ->
    permission(role, github_id)
```
to be explicit?  That's also more consistent with at least the DSL that I've been writing.

brianhw: I'll try that.

PR #714, opened 2019-06-12
job for check primary keys

syed-awais-ali:


devops/resources/check_primary_keys.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+cd $WORKSPACE/configuration/util/jenkins/check_table_size
+
+pip install -r requirements.txt
+. ../assume-role.sh
+
+# Assume the role
+set +x
```
jdmulloy: Not sure we need ```set +x``` or ```set-x```

syed-awais-ali: `+x` to not print anything to console.

devops/jobs/CheckPrimaryKeys.groovy
```diff
@@ -0,0 +1,70 @@
...
+
+                    publishers {
+                        mailer(extraVars.get('NOTIFY_ON_FAILURE'), false, false)
+
+                    }
+                }
+           }
+       }
+   }
+}
```
jdmulloy: Missing newline.

PR #718, opened 2019-06-12
Add celery dashboard job

jdmulloy:


devops/jobs/CreateCeleryDashboard.groovy
```diff
@@ -0,0 +1,113 @@
...
+                            systemSitePackages(false)
+
+                            command(
+                                dslFactory.readFileFromWorkspace("devops/resources/create_celery_dashboard.sh")
+                            )
+
+                        }
+                    }
+
+                    // publishers {
```
adzuci: Do you have an opinion on having commented code in a repo / could we remove this or do we plan on uncommenting it once we merge it?  (Not sure if there is a standard workflow or not for this repo)

adzuci:
P.S. Thanks for adding this!   I think it will make a big QOL difference for people on call and it might also be possible to add a link to the dashboard in our alerting or pull a cloudwatch graph into a chatops command.

PR #719, opened 2019-06-13
Google spreadsheet loading job DSL.

HassanJaveed84:
secure repo pr: https://github.com/edx-ops/analytics-secure/pull/247

### DIFF
--- a/dataeng/jobs/analytics/LoadGoogleSpreadsheetToWarehouse.groovy
+++ b/dataeng/jobs/analytics/LoadGoogleSpreadsheetToWarehouse.groovy
@@ -1,0 +1,28 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.to_date_interval_parameter
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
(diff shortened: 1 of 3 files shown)


HassanJaveed84:
I meant to bike cage that. I think it make sense to make it a list, and be configured through luigi override file(prod-edx.cfg) ? We'd change the workflow to work on list of spreadsheets instead. 

But do we store worksheets belong to different spreadsheets in a single schema ? How do we configure different schemas for different worksheets ?

brianhw:
Fair question.  I think it's fair for all worksheets in a spreadsheet to go to the same Schema.  (If you don't want that, as an author, then put them in different spreadsheets. : )   I think that in many cases, it will make sense to load multiple spreadsheets into the same schema as well -- for example, the first one coming from Jackie is expected to appear in the "financial_reporting" schema.   So if we get to a point where there are one-or-more spreadsheets going into one schema and one-or-more spreadsheets going into a second schema, then we would need to either move things from config files into DSL, or change the input from a ListParameter to a DictParameter containing the spreadsheet-key and the destination schema.  The advantage of the latter is that we can more easily add spreadsheets, yes, but we could also add more information for each -- e.g. a spreadsheet name that must match so that we confirm and document what's being loaded (hard to tell from just a key), a prefix to add to worksheet titles when creating output table names so that we don't require the worksheet names to be globally unique, whether the spreadsheet has a column-type row, etc.

brianhw:
Just got confirmation from Jackie in the DS channel:   'we are going to need to load a second google sheet that contains “previously reported” data'.   I thought that might be the case, so if we can figure that out now, we'll be better able to deal with that requirement.  (Mind you, not loads of additional ones, just one.  For now.)

PR #720, opened 2019-06-14
Create a job for Affiliate Window warehouse loading

bmedx:


dataeng/jobs/createJobs.groovy
```diff
@@ -81,6 +82,7 @@ def taskMap = [
     GENERATE_WAREHOUSE_DOCS_JOB: GenerateWarehouseDocsJob,
     SQL_SCRIPTS_JOB: SqlScriptsJob,
     SINGLE_SQL_SCRIPT_JOB: SingleSqlScriptJob,
+    LOAD_AFFILIATE_WINDOW_JOB: LoadAffiliateWindowWarehouseJob,
```
pwnage101: I think you'll probably also want to add `name('affiliate-window')` to the `listView('Warehouse') {` block below so that the job shows up under the "Warehouse" tab in the jenkins UI.

dataeng/jobs/analytics/LoadAffiliateWindowToWarehouse.groovy
```diff
@@ -0,0 +1,31 @@
...
+            parameters {
+                stringParam('OUTPUT_SCHEMA', 'finance', '')
+            }
+            multiscm common_multiscm(allVars)
+            triggers common_triggers(allVars)
+            wrappers common_wrappers(allVars)
+            publishers common_publishers(allVars)
+            steps {
+                shell(dslFactory.readFileFromWorkspace('dataeng/resources/load-affiliate-window.sh'))
+                if (allVars.get('SNITCH')) {
```
pwnage101: Don't bother with snitch, we don't use it anyway, and you don't have a SNITCH variable anyway.

PR #733, opened 2019-07-01
Override default NOTIFY on single_script_job to not page on fail

bmedx:


dataeng/jobs/analytics/SqlScripts.groovy
```diff
@@ -49,6 +49,7 @@ class SqlScripts {
             parameters {
                 stringParam('SOURCE_SCRIPT', '', 'Path to the source script, relative to the root of the repository. Most scripts just require a script name.')
                 stringParam('SCRIPT_NAME')
+                stringParam('NOTIFY', '', 'Space separated list of emails to send notifications to.')
             }
             parameters common_parameters(allVars)
```
pwnage101: common_parameters may override the NOTIFY parameter, and even if it doesn't it's double-defined which isn't clean.  My preference would be to either:

1. set the `NOTIFY` variable in `job-configs/SINGLE_SQL_SCRIPT_JOB_EXTRA_VARS.yaml` in the edx/analytics-secure repo to an empty string, or
2. remove the line below `publishers common_publishers(allVars)` which currently has a sole purpose of sending failure notification emails.

pwnage101: I don't have a preference either way.  Setting the NOTIFY variable is perhaps more verbose.

bmedx: Ah yes I forgot about the job-configs overrides, I can do this there. I don't want to disable notifications in general, people should be able to put in their email if they want to be notified.

bmedx:
Closing in favor of https://github.com/edx-ops/analytics-secure/pull/253

PR #739, opened 2019-07-10
Add course-discovery 'make upgrade' job

mikix:


feanil:
Adding @estute and @michaelyoungstrom since Jeremy is out on paternity.

mikix:
Thanks!

PR #755, opened 2019-07-25
adding auth token

ceguiluzrosas:
In order for clamps to build a jenkins job remotely (GrantSSHAccess to Sandbox), I need to set a job token to my POST Request.

### DIFF
--- a/devops/jobs/CreateSandboxSSHAccess.groovy
+++ b/devops/jobs/CreateSandboxSSHAccess.groovy
@@ -70,7 +70,7 @@ class CreateSandboxSSHAccess {
         sshAgent('sandbox-ssh-keys')
       }
 
-      authenticationToken(extraVars.get('JENKINS_SANDBOX_JOB_KEY', 'S@nD80%'))
+      authenticationToken(extraVars.get('JENKINS_SANDBOX_JOB_KEY'))
 
       environmentVariables {
         env('SSH_USER',extraVars.get('SSH_USER','ubuntu'))
--- a/devops/jobs/CreateSandboxSSHAccess.groovy
+++ b/devops/jobs/CreateSandboxSSHAccess.groovy
@@ -70,6 +70,8 @@ class CreateSandboxSSHAccess {
         sshAgent('sandbox-ssh-keys')
       }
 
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandboxSSHAccess.groovy
```diff
@@ -70,6 +70,8 @@ class CreateSandboxSSHAccess {
         sshAgent('sandbox-ssh-keys')
       }
 
+      authenticationToken("${JENKINS_AUTH_TOKEN}")
```
ceguiluzrosas: According to Stu, there is a [End-to-End Job](https://github.com/edx/jenkins-job-dsl-internal/blob/master/jobs/tools-edx-jenkins.edx.org/createEndToEndTestJobs.groovy#L49) that can be called remotely via an auth token that set in the highlighted line. If you look at line 36, a "${JENKINS_AUTH_TOKEN}" is passed as a parameter to the Ecommerce Config (linked hashmap) above. I've tried "grepping" the code for this token but no success (must be an environmental variable)

ceguiluzrosas: Decided to store a "JENKINS_SANDBOX_JOB_KEY" variable (here's the [PR](https://github.com/edx/edx-internal/pull/1023) that will be read using extraVars.get(xxx)

devops/jobs/CreateSandboxSSHAccess.groovy
```diff
@@ -70,6 +70,8 @@ class CreateSandboxSSHAccess {
         sshAgent('sandbox-ssh-keys')
       }
 
+      authenticationToken(extraVars.get('JENKINS_SANDBOX_JOB_KEY', 'S@nD80%'))
+
```
ceguiluzrosas: The actual key was created using LastPass's Key generator. I added a default key just in case

coryleeio: Dont add default passwords because if you forget to set them then they are known, this repo is public as well. It would be better to blow up if it didn't exist, or skip it if it didn't exist.

ceguiluzrosas: @coryleeio Done 👍

PR #758, opened 2019-07-26
add python3 codejail job stub dsl

estute:
Rather than have the codejail job run each python environment, I am creating one job per environment. This way we can have Github require 2.7 pass for merges, and have 3.x jobs be optional/informative.


estute:
Disregard the travis failures- they are unrelated. Something to do with a travis worker update conflicting w/ gradle/jdk versions. I will debug in another PR. This DSL works and is correctly seeded on build Jenkins

PR #765, opened 2019-08-07
Warn about hyphens in CreateSandbox dns_name

kdmccormick:
Hyphens in the sandbox dns_name cause some weird issues in the microservices, probably due to some nginx regex rules along the lines of `[service_name]-[dns_name]`. Let's warn about that.

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -105,7 +105,8 @@ class CreateSandbox {
                          One reason you might want to override this field is if you are building a sandbox for review or a specific task. \
                          If setting this, you probably want to also set name_tag below. \
                          For example, if you are building a sandbox for pull request 1234 put in 'pr1234' which will setup the sandbox <i>pr1234.sandbox.edx.org</i>.<br /> \
-                         <b>If you are building a sandbox for yourself leave this blank</b><b>Do not use underscores</b>")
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -105,7 +105,8 @@ class CreateSandbox {
                          One reason you might want to override this field is if you are building a sandbox for review or a specific task. \
                          If setting this, you probably want to also set name_tag below. \
                          For example, if you are building a sandbox for pull request 1234 put in 'pr1234' which will setup the sandbox <i>pr1234.sandbox.edx.org</i>.<br /> \
-                         <b>If you are building a sandbox for yourself leave this blank</b><b>Do not use underscores</b>")
+                         <b>If you are building a sandbox for yourself leave this blank.</b> <br /> \
+                         <b>Do not use hyphens or underscores</b>.")
```
schenedx: how about `Do not use hyphens or underscores here`?

timmc-edx:
Do you happen to remember what issues this caused? My sandbox seems to work fine despite the hyphen, but maybe I'm not testing the right thing.

kdmccormick:
@timmc-edx Try accessing a service like `discovery` or one of the weird hyphenated ones like `enterprise-catalog`.

kdmccormick:
If I remember correctly, the issue may have arised when trying to access `portal-designer` on a Sandbox.

PR #768, opened 2019-08-13
TE-2961 Fix tests, add edx-sphinx-theme deps job

jmbowman:
* Upgrade the gradle wrapper to match the gradle version used in Travis CI's new Xenial workers.
* Add a `make upgrade` job for `edx-sphinx-theme`.
* Add a build status badge to the README.

The tests now run under openjdk11 because:

* oraclejdk8 isn't supported on the new Travis CI Xenial workers
* We could downgrade back to Trusty, but that doesn't match what we use in prod
* oraclejdk9 and oraclejdk10 are no longer supported (non-LTS releases)
* oraclejdk11 now [charges for production use](https://www.dariawan.com/blog/important-changes-in-oracle-jdk-11-license/)
* openjdk11 is functionally identical to oraclejdk11, and is currently the default on Travis Xenial workers


jmbowman:
@coryleeio @jdmulloy See above for notes on where the JDK situation ended up.  tl;dr - we'll probably be shifting everything over to openjdk11 eventually.

PR #769, opened 2019-08-13
Automatically upgrade cookiecutter-django-ida requirements

kdmccormick:
@edx/testeng 

In https://github.com/edx/cookiecutter-django-ida/pull/53, I add `make upgrade` to the IDA cookiecutter. It doesn't upgrade the requirements of the repo itself, but upgrades the requirements of to-be-cookie-cut IDA.

I will leave it up to y'all on test eng to decide whether this is worth merging or not.

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -37,6 +37,15 @@ Map cookiecutterDjangoApp = [
     emails: ['testeng@edx.org']
 ]
 
+Map cookiecutterDjangoIda = [
+    org: 'edx',
+    repoName: 'cookiecutter-django-ida',
+    cronValue: '@weekly',
+    githubUserReviewers: [],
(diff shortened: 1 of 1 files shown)


jmbowman:
In cookiecutter-django-app, I explicitly tried to avoid creating the `*.txt` requirements files until actually generating a new repo (partially because some of the requirements files are templated and can't be used in their original form).  I suppose we could take this approach to keeping the IDA cookiecutter dependencies current, but would it work better to just omit them in that cookiecutter also?

kdmccormick:
@jmbowman That makes sense, and keeping the IDA/app cookie-cutters consistent is also probably a good call.

I pushed a commit to my cookie-cutter IDA PR that removes the version pins, and has `make test` (from the root) generate them: https://github.com/edx/cookiecutter-django-ida/pull/53/commits/63dfd82cda228f1f78afb799930cc1eaa805aec0

If that looks good then we can close this PR.

PR #770, opened 2019-08-19
set env var for python version in platform pr jobs

estute:
Add the `PYTHON_VERSION` environment variable in for PR jobs that run w/ python 3. For now, if no value is set, default to python 2.7


platform/jobs/pipelines/edxPlatformPipelinePr.groovy
```diff
@@ -271,6 +274,7 @@ jobConfigs.each { jobConfig ->
             environmentVariables(
                 REPO_NAME: "${jobConfig.repoName}",
                 TOX_ENV: "${jobConfig.toxEnv}"
+                PYTHON_VERSION: "${jobConfig.pythonVersion}"
```
jmbowman: This doesn't trigger an error for configs in which `pythonVersion` wasn't defined?

estute: No. We follow the same pattern for setting `TOX_ENV`. Groovy just evaluates it to null (which I check for in the platform script).

PR #784, opened 2019-09-04
add parameters to decrypt and copy app config from internal (OPS-4261)

nadeemshahzad:


coryleeio:
@nadeemshahzad Can you make it possible for the sandbox job to change their yaml file that they upload from internal?

For instance, instead of sandbox-internal/sandbox-remote-config/notes.yml someone might want to do the following to test a new feature:
sandbox-internal/sandbox-remote-config/sally-notes.yml

coryleeio:
The default should be set to the normal .yml in sandbox-internal though

PR #790, opened 2019-09-05
Add dbt schema builder script and groovy

bmedx:


dataeng/jobs/analytics/SnowflakeSchemaBuilder.groovy
```diff
@@ -0,0 +1,45 @@
...
+                stringParam('DBT_PROFILE', allVars.get('DBT_PROFILE'), 'dbt profile from analytics-secure to work on.')
+                stringParam('DBT_TARGET', allVars.get('DBT_TARGET'), 'dbt target from analytics-secure to work on.')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+            }
+            logRotator common_log_rotator(allVars)
+            multiscm {
+                git {
+                    remote {
+                        url('$TRANSFORMS_REPO')
+                        branch('$TRANSFORMS_BRANCH')
```
brianhw: Should these have WAREHOUSE_ in front of them?

dataeng/resources/snowflake-schema-builder.sh
```diff
@@ -0,0 +1,20 @@
...
+# Create a new git branch
+now=$(date +%d_%m_%Y_%H_%M_%S)
+branchname="builder_$now"
+git checkout -b "$branchname"
+
+# Run the dbt script to update schemas and sql
+python tools/dbt_schema_builder/schema_builder.py build --schemas $SCHEMAS --project $DBT_PROJECT --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/analytics-tasks/
+
+# Commit all changes to the new branch
+git commit -a --message "Schema Builder automated dbt update at $now"
```
brianhw: I assume that most of the time we won't expect to see differences.   Is there something we should do to check for that condition, and not create a PR?   Or will there be dates and such in various comments that will appear and churn every time this is run?

dataeng/resources/snowflake-schema-builder.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Create a PR on Github from the new branch
+hub pull-request --push --no-edit
```
doctoryes: Maybe a `-r <REVIEWERS>` should be added here as well for notification purposes?

dataeng/resources/snowflake-schema-builder.sh
doctoryes: Agreed - could `git status` be used to see if the commit was empty?

bmedx: Makes sense, but it will introduce a bit of work to get the schema files to sort deterministically in the generation script (to make the comparisons work). Hopefully I can just fix it in the query. 🤞

bmedx: I've updated the generation script and just pushed changes here to address this. I've tested it as well as I can, adding, deleting, and changing files.

dataeng/jobs/analytics/SnowflakeSchemaBuilder.groovy
```diff
@@ -0,0 +1,45 @@
...
+                }
+            }
+            triggers common_triggers(allVars)
+            wrappers {
+                timestamps()
+            }
+            publishers common_publishers(allVars)
+            steps {
+                virtualenv {
+                    pythonName('System-CPython-3.6')
```
pwnage101: System-CPython-3.5.2 is what's available under Manage Jenkins -> Global Tool Configuration -> Python Installations

Does this actually work without one that exactly matches the name that you've given?

bmedx: I don't know, @doctoryes suggested this was available. It should work with either.

doctoryes: Other jobs used that Python version? And closer to Python3.7, the better. But either version is fine with me.

pwnage101: Other jobs that are defined for different jenkins instances maybe, but not any of ours.  We have to use the 3.5.2 version unless we install another one under Manage Jenkins -> Global Tool Configuration -> Python Installations.

brianhw: I understand that just doing the installation would make it harder to replicate the recreation of the Jenkins server.  But I wonder if it is more valuable to move ahead with 3.6 (or 3.7), since that's a direction we would be wanting to head in future anyway.

pwnage101: We're probably all synced up offline about this, but just wanted to be clear that the plan currently is to add a newer python to the box and add the corresponding jenkins tool configuration, all while trying to keep the ansible in sync.

bmedx: I have a WIP PR for this that is going through the process now for the ansible side of things. I've manually installed 3.7 and configured Shining Panda through the UI to match what I expect those ansible changes will create in the future.

PR #792, opened 2019-09-11
Snowflake read replica import DSL.

HassanJaveed84:
analytics-secure PR: https://github.com/edx-ops/analytics-secure/pull/277


pwnage101:
@brianhw The properties file is saved into the root of the `${WORKSPACE}` which should be specific to this job.

pwnage101:
@brianhw to clarify, we may normally use `predefinedProp(String key, String value)` and skip writing to a file, but in this case we needed to interpolate the value in a bash context, so writing to a file becomes necessary to communicate the interpolated value.

PR #800, opened 2019-09-25
Break vertica export into three job definitions.

brianhw:
As part of work on DE-1700, this PR reorganizes Vertica export jobs to separate the writing to S3 from the loading into BigQuery and Snowflake. With separate Jenkins jobs to do the export versus the loading, the loading into BQ and Snowflake can then run in parallel.

Should be merged with https://github.com/edx/edx-analytics-pipeline/pull/764.

### DIFF
--- a/dataeng/jobs/analytics/LoadVerticaSchemaToBigquery.groovy
+++ b/dataeng/jobs/analytics/LoadVerticaSchemaToBigquery.groovy
@@ -6,31 +6,25 @@ import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
 
-class LoadVerticaSchemaToSnowflake {
+class LoadVerticaSchemaToBigquery {
     public static def job = { dslFactory, allVars ->
         allVars.get('VERTICA_SCHEMAS').each { schema, schema_config ->
-            dslFactory.job("load-vertica-schema-to-snowflake-$schema") {
+            dslFactory.job("load-vertica-schema-to-bigquery-$schema") {
                 logRotator common_log_rotator(allVars, schema_config)
                 parameters common_parameters(allVars, schema_config)
                 parameters {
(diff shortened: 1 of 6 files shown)


dataeng/resources/load-vertica-schema-to-snowflake.sh
```diff
@@ -3,7 +3,7 @@
 env
 
 ${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
- VerticaSchemaToSnowflakeTask --local-scheduler \
+ LoadVerticaSchemaFromS3ToSnowflakeTask --local-scheduler \
  --date $(date +%Y-%m-%d -d "$RUN_DATE") \
  ${OVERWRITE} \
  --credentials $SNOWFLAKE_CREDENTIALS \
```
doctoryes: Note to self (@doctoryes): This file will require the addition of `scratch_schema` and `run_id`.

brianhw: Yup.    And the code as well, to pass args from Schema to Table task.

dataeng/jobs/analytics/LoadVerticaSchemaToBigquery.groovy
```diff
@@ -0,0 +1,32 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
```
HassanJaveed84: Is this being used ?

brianhw: Ah, good catch.  I'll remove that in my next DSL PR.   (I need to make all of this be in turn triggered by SQL scripts for the next story.)

PR #802, opened 2019-09-29
Add missing scratch_schema/run_id parameters to Groovy.

doctoryes:
These changes were missed in https://github.com/edx/jenkins-job-dsl/pull/801, causing each Snowflake application DB import to fail, such as:

http://jenkins.analytics.edx.org/job/snowflake-credentials-read-replica-import/21/

Merging this PR and re-seeding the relevant jobs should fix the error.


brianhw:
I'm just curious why this contains RUN_ID, if it is never referenced?   The parameter is hardcoded to BUILD_ID instead.

doctoryes:
Drat - you're right, @brianhw . The intent is to always use `BUILD_ID` and not provide an override. I'll remove `RUN_ID` in another PR.

doctoryes:
PR is here: https://github.com/edx/jenkins-job-dsl/pull/803

PR #803, opened 2019-09-30
Remove RUN_ID override from Snowflake import jobs.

doctoryes:
I mistakenly included `RUN_ID` in https://github.com/edx/jenkins-job-dsl/pull/802 . This PR simply removes `RUN_ID` from those jobs. The intent is to always use the Jenkins `BUILD_ID` for the value and never override it.


doctoryes:
Thanks for the catch, @brianhw - this PR removes `RUN_ID`.

PR #808, opened 2019-10-01
Warehouse transforms jobs

hunytalk:
Create warehouse-transforms-{environment} job.

Each DSL will be modeled after the DSLs for sql-scripts-{daily,monthly,quarterly}.  Instead of invoking sql-scripts, these new jobs will invoke dbt models that are tagged either daily, monthly, or quarterly.

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransforms.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransforms.groovy
@@ -1,0 +1,55 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+class WarehouseTransforms{
+    public static def job = { dslFactory, allVars ->
+        allVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("warehouse-transforms-$environment"){
+                logRotator common_log_rotator(allVars)
+                parameters secure_scm_parameters(allVars)
+                parameters {
+                    stringParam('WAREHOUSE_TRANSFORMS_URL', allVars.get('WAREHOUSE_TRANSFORMS_URL'), 'URL for the Warehouse Transforms Repo.')
+                    stringParam('WAREHOUSE_TRANSFORMS_BRANCH', allVars.get('WAREHOUSE_TRANSFORMS_BRANCH'), 'Branch of Warehouse Transforms to use.')
+                    stringParam('MODEL_TAG', env_config.get('MODEL_TAG', allVars.get('MODEL_TAG')), 'Tagged model that will be run.')
+                    stringParam('DBT_PROJECT', env_config.get('DBT_PROJECT', allVars.get('DBT_PROJECT')), 'dbt project in warehouse-transforms to work on.')
+                    stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
+                    stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
+                    stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                }
+                multiscm secure_scm(allVars) << {
+                    git {
+                        remote {
+                            url('$WAREHOUSE_TRANSFORMS_URL')
+                            branch('$WAREHOUSE_TRANSFORMS_BRANCH')
+                            credentials('1')
+                        }
+                        extensions {
+                            relativeTargetDirectory('warehouse-transforms')
+                            pruneBranches()
+                            cleanAfterCheckout()
+                        }
+                    }
+                }
+                triggers common_triggers(allVars, env_config)
+                wrappers common_wrappers(allVars)
+                publishers common_publishers(allVars)
+                steps {
+                    virtualenv {
+                        pythonName('PYTHON_3.7')
+                        nature("shell")
+                        systemSitePackages(false)
+                        command(
+                            dslFactory.readFileFromWorkspace("dataeng/resources/warehouse-transforms.sh")
+                        )
+                    }
+                }
+            }
+        }
+    }
+}
+
--- a/dataeng/resources/warehouse-transforms.sh
+++ b/dataeng/resources/warehouse-transforms.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup
+cd $WORKSPACE/warehouse-transforms
+# To install right version of dbt
+pip install -r tools/dbt_schema_builder/requirements.txt
+
+cd $DBT_PROJECT
+dbt clean
+
+cd $WORKSPACE/warehouse-transforms
+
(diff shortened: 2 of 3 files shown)


dataeng/resources/warehouse-transforms-daily.sh
```diff
@@ -0,0 +1,14 @@
...
+# Setup
+cd $WORKSPACE/warehouse-transforms
+pip install -r tools/dbt_schema_builder/requirements.txt
+
+cd $DBT_PROJECT
+dbt clean
+
+cd $WORKSPACE/warehouse-transforms
+
+dbt run --models tag: $MODEL_TAG --project $DBT_PROJECT --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/analytics-tasks/
```
brianhw: I mentioned this before, but I'll mention it again before we run out of runway:   is it possible to not store dbt profiles in the same place we store our pipeline stuff?   Could we start pointing profiles-dir somewhere like $WORKSPACE/analytics-secure/dbt-profiles/?

dataeng/jobs/analytics/WarehouseTransformsDaily.groovy
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+            publishers common_publishers(allVars)
+            steps {
+                virtualenv {
+                    pythonName('PYTHON_3.7')
+                    nature("shell")
+                    systemSitePackages(false)
+                    command(
+                        dslFactory.readFileFromWorkspace("dataeng/resources/warehouse-transforms-daily.sh")
```
brianhw: Hmm.  I commented on this, but it didn't show up in my review, so I'll try again:   I think this whole thing should be structured like the SQL scripts stuff, to have the individual jobs be defined in analytics-secure, and have this loop over all the warehouse-transform jobs to create.  Each would provide a name, and a schedule, as well as the various parameters.   Note that nothing here is actually "daily", so that should probably be removed from the shell script, the groovy file, and the job name.   And at present, we have two daily SQL script jobs that run completely independently of each other, and I think we should expect there may be something similar going forward.

dataeng/resources/warehouse-transforms-daily.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup
+cd $WORKSPACE/warehouse-transforms
+pip install -r tools/dbt_schema_builder/requirements.txt
```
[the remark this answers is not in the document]
hunytalk: I'll comment it.

dataeng/jobs/analytics/WarehouseTransformsDaily.groovy
```diff
@@ -0,0 +1,51 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
```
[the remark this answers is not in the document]
hunytalk: I'll remove it for now.

dataeng/resources/warehouse-transforms-daily.sh
hunytalk: @brianhw It won't be problem for me to change profiles.yaml location to the one you suggested. But i think it would require changes at @bmedx end as well since schema builder task is also using this. 
Also could you explain why would there be a need to have it move in a separate location from our pipeline code!

dataeng/jobs/analytics/WarehouseTransformsDaily.groovy
hunytalk: @brianhw I don't think i follow it completely about what you said here.

dataeng/jobs/analytics/WarehouseTransformsDaily.groovy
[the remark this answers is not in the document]
brianhw: If you remove it, how will this run daily?

dataeng/resources/warehouse-transforms-daily.sh
brianhw: I agree.  This needs a broader discussion. I thought that could happen here, but maybe I'll just bring it up in Slack.

dataeng/jobs/analytics/WarehouseTransformsDaily.groovy
brianhw: Sorry, I'll try again.  I think this should be in a file WarehouseTransforms.groovy, that points to dataeng/resources/warehouse-transforms.sh.   It should read from a WAREHOUSE_TRANSFORMS_JOB_EXTRA_VARS.yaml that is structured similarly to SQL_SCRIPTS_JOB_EXTRA_VARS.yaml, since warehouse-transforms jobs are intended to replace sql-scripts jobs.  Each ENVIRONMENT has a job "name" (e.g. "bi-daily", "financial-reporting-daily", etc.), and each of these contains the relevant parameters for that job.  These include MODEL_TAG, DBT_PROJECT, DBT_PROFILE, DBT_TARGET, but also JOB_FREQUENCY.   The code in WarehouseTransforms.groovy would include a wrapper for processing multiple jobs:
{code}
    public static def job = { dslFactory, allVars ->
         allVars.get('ENVIRONMENTS').each { environment, env_config ->
            dslFactory.job("warehouse-transforms-$environment") {
                ...
                triggers common_triggers(allVars, env_config)
{code}

The addition of the trigger and the defining of JOB_FREQUENCY is what gets the job to trigger.  Without that, it will only run manually.

hunytalk: @brianhw Thanks for the explanation. 
Since you mentioned another daily job (financial-reporting-daily) running independently of "daily" job, i've added this job here also. 
For now i've taken exact same job schedule for these job that we have in "sql-scripts-{$environment}".

hunytalk: By here i mean, added it as environment in analytics-secure.

hunytalk: @brianhw It's ready for re-review.

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -0,0 +1,54 @@
...
+            dslFactory.job("warehouse-transforms-$environment"){
+                logRotator common_log_rotator(allVars)
+                parameters secure_scm_parameters(allVars)
+                parameters {
+                    stringParam('WAREHOUSE_TRANSFORMS_URL', allVars.get('WAREHOUSE_TRANSFORMS_URL'), 'URL for the Warehouse Transforms Repo.')
+                    stringParam('WAREHOUSE_TRANSFORMS_BRANCH', allVars.get('WAREHOUSE_TRANSFORMS_BRANCH'), 'Branch of Warehouse Transforms to use.')
+                    stringParam('MODEL_TAG', allVars.get('MODEL_TAG'), 'Tagged model that will be run.')
+                    stringParam('DBT_PROJECT', allVars.get('DBT_PROJECT'), 'dbt project in warehouse-transforms to work on.')
+                    stringParam('DBT_PROFILE', allVars.get('DBT_PROFILE'), 'dbt profile from analytics-secure to work on.')
+                    stringParam('DBT_TARGET', allVars.get('DBT_TARGET'), 'dbt target from analytics-secure to work on.')
```
brianhw: I think these need to reference env_config first, then fall back to allVars.

dataeng/resources/warehouse-transforms.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup
+cd $WORKSPACE/warehouse-transforms
+pip install -r tools/dbt_schema_builder/requirements.txt
```
brianhw: nit: suggesting adding a comment here indicating that this is used to install the right version of dbt.

PR #814, opened 2019-10-15
Update README to include structure section.

adzuci:


feanil:
Another simpler version of that paragraph that I might suggest is:

# Making a new job?  Make it look like this one(link) NOT like this one(link).

As an addition not as a replacement of the context you provide.

README.md
```diff
@@ -116,3 +116,19 @@ Constant | Purpose | Use
 ------------ | ------------- | -------------
 credential | Allow access to private git repositories | In the credential() function in the Git Scm Context
 
+### Structure
+
+TL;DR Making a new job? Make it look like [this one](https://github.com/edx/jenkins-job-dsl/blob/master/testeng/jobs/backupJenkins.groovy#L47) NOT like [this one](https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/AggregateDailyTrackingLogs.groovy#L11).
+
+There are two common ways to write jobs:
+1: The Build jenkins jobs are often flattened (like [this](https://github.com/edx/jenkins-job-dsl/blob/master/testeng/jobs/backupJenkins.groovy#L47) ) or if there are many jobs that are similar you can loop over a list of Maps like [this](https://github.com/edx/jenkins-job-dsl/blob/master/testeng/jobs/upgradePythonRequirements.groovy#L206).  This has the benefit of keeping all the logic in one file and being easy to understand.
```
pwnage101: nit, use the name of the file in the link text.  general rule of thumb is that I should be able to list all links on a page out of context and understand essentially what they're pointing to.

adzuci:
> Seems good and reasonable to me. Is there a plan to migrate the old jobs to the preferred paradigm?

No explicit plan, but the hope is to at least keep new jobs as simple as possible and possibly encourage others to simplify old jobs as they edit them.

PR #837, opened 2019-10-23
fix schema builder cli option

estute:
A flag on the schema-builder will change once this PR (https://github.com/edx/warehouse-transforms/pull/22) merges. From the comments:

```
The schema-builder CLI changed:

    --schemas becomes --raw-schemas and the list of schemas passed into it must include a _RAW suffix.

```

### DIFF
--- a/dataeng/resources/snowflake-schema-builder.sh
+++ b/dataeng/resources/snowflake-schema-builder.sh
@@ -5,7 +5,9 @@ set -ex
 cd $WORKSPACE/warehouse-transforms
 pip install -r tools/dbt_schema_builder/requirements.txt
 
-cd $DBT_PROJECT
+cd $WORKSPACE/warehouse-transforms/app_views_project
+dbt clean
+cd $WORKSPACE/warehouse-transforms/warehouse_transforms_project
 dbt clean
 
 cd $WORKSPACE/warehouse-transforms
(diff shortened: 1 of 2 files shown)


dataeng/resources/snowflake-schema-builder.sh
```diff
@@ -16,7 +16,7 @@ branchname="builder_$now"
 git checkout -b "$branchname"
 
 # Run the dbt script to update schemas and sql
-python tools/dbt_schema_builder/schema_builder.py build --schemas $SCHEMAS --project $DBT_PROJECT --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
+python tools/dbt_schema_builder/schema_builder.py build --raw-schemas $SCHEMAS --project $DBT_PROJECT --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
```
bmedx: If this is merging after the big batch of changes I think `--project` is going away too.

estute:
@doctoryes yes, unfortunately. This script is contained in the dsl repo, which only gets referenced at seed-time. I am in favor of moving this script into the schema builder repo itself.

dataeng/resources/snowflake-schema-builder.sh
estute: Good catch. Fixed it (and in https://github.com/edx-ops/analytics-secure/pull/300)

PR #841, opened 2019-10-24
only run schema builder for the app views dbt project

estute:
I assumed that since we split the dbt project into two separate projects, we would need to run the schema builder for both. However, when I ran the schema builder on the `warehouse_transforms_project`, it failed spectacularly, with all kinds of threading errors. 

Should we be running schema builder on this project yet?


doctoryes:
No, we shouldn't - the schema builder script should only be working in the app_views project, as you've discovered. The warehouse_transforms project will just use plain dbt with no script.

PR #864, opened 2019-12-18
Remove edx-platform Py2 jobs

jmbowman:
Stop running the edx-platform tests under Python 2 on master and PRs against it, now that edx-platform is running on Python 3 in production.  The reduction in test job count should help reduce the number of PRs that hit problems with flaky tests, and switching the Python 3 jobs from tox to run directly in a Python 3.5 virtualenv should also somewhat reduce the time to feedback.  The Ironwood jobs still run under Python 2.

Also removed some obsolete artifact handling code related to pa11ycrawler, which we don't run anymore.

### DIFF
--- a/platform/jobs/edxPlatformAccessibilityPr.groovy
+++ b/platform/jobs/edxPlatformAccessibilityPr.groovy
@@ -47,7 +47,8 @@ Map publicJobConfig = [
     workerLabel: 'js-worker',
     whitelistBranchRegex: /^((?!open-release\/).)*$/,
     context: 'jenkins/a11y',
-    triggerPhrase: /.*jenkins\W+run\W+a11y.*/
+    triggerPhrase: /.*jenkins\W+run\W+a11y.*/,
+    pythonVersion: '3.5',
 ]
 
 Map privateJobConfig = [
--- a/platform/jobs/edxPlatformAccessibilityPr.groovy
+++ b/platform/jobs/edxPlatformAccessibilityPr.groovy
@@ -102,7 +102,7 @@ jobConfigs.each { jobConfig ->
         logRotator JENKINS_PUBLIC_LOG_ROTATOR(7)
         concurrentBuild()
         environmentVariables {
-            env('PYTHON_VERSION', jobConfig.pythonVersion),
+            env('PYTHON_VERSION', jobConfig.pythonVersion)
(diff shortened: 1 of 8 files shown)


platform/jobs/edxPlatformAccessibilityPr.groovy
```diff
@@ -57,18 +57,8 @@ Map privateJobConfig = [
...
-
-Map python3JobConfig = [
-    open : true,
-    jobName : 'edx-platform-python3-accessibility-pr',
-    repoName : 'edx-platform',
-    workerLabel: 'js-worker',
-    whitelistBranchRegex: /^((?!open-release\/).)*$/,
-    context: 'jenkins/python3.5/a11y',
-    triggerPhrase: /.*jenkins\W+run\W+py35-django111\W+a11y.*/,
-    toxEnv: 'py35-django111'
```
estute: ^ i'm out of the loop. Have we stopped using tox?

platform/jobs/edxPlatformAccessibilityPr.groovy
```diff
@@ -57,18 +57,8 @@ Map privateJobConfig = [
...
+    triggerPhrase: /.*jenkins\W+run\W+a11y.*/,
+    pythonVersion: '3.5',
```
estute: ^ if we are including `pythonVersion` in the private job, we need to have it in the public job too, right?

platform/jobs/edxPlatformAccessibilityPr.groovy
jmbowman: We were using tox for the Python 3 jobs because we modeled them after the Django upgrade jobs where we had to use it.  But they should work fine by just using a Python 3 virtualenv, also.  We'll start using tox again for optional Django 2.0, 2.1, and 2.2 jobs shortly.

platform/jobs/edxPlatformAccessibilityPr.groovy
jmbowman: Yes, apparently missed that one when making the updates.  Fixing.

platform/jobs/edxPlatformAccessibilityPr.groovy
estute: oh, right. Thanks

PR #868, opened 2020-01-02
Add program reports job

Zacharis278:
@edx/masters-devs 

Adds DSL to generate a job to run the program reporting pipeline task here: https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/programs/program_reports.py

one open question on this, it appears there is a default env variable for the trigger cron.  I'm not sure where that is defined yet.

### DIFF
--- a/dataeng/jobs/analytics/ProgramEnrollmentReports.groovy
+++ b/dataeng/jobs/analytics/ProgramEnrollmentReports.groovy
@@ -1,0 +1,38 @@
+
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_multiscm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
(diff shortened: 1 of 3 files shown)


dataeng/resources/program-enrollment-reports.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+if [ -z "$NUM_REDUCE_TASKS" ]; then
```
estute: wouldn't this always be the case? You only set it within this if-block.

estute:
You should be all set with the values you supplied in your analytics-secure PR. I don't think there actually is a default value for `JOB_FREQUENCY`, despite `def job_frequency = env.get('JOB_FREQUENCY', allVars.get('JOB_FREQUENCY'))` looking like there is

dataeng/jobs/analytics/ProgramEnrollmentReports.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+        dslFactory.job('program-enrollment-reports') {
+
+            // disabled until downstream reporting work is complete
+            disabled(true)
+
+            logRotator common_log_rotator(allVars)
+            parameters common_parameters(allVars)
+            parameters {
+                stringParam('OUTPUT_ROOT', env_config.get('OUTPUT_ROOT'), 'Report output location')
```
doctoryes: Looks like `env_config` here is a copy error? Shouldn't it be `allVars`?

PR #871, opened 2020-01-12
Update all User Retirement jobs to use Python 3

pwnage101:
This is a response to setuptools 45 now requiring python 3. It is easier
to just update these jobs which are already tested under python 3 to
actually run python 3.  It helps that the PYTHON_3.5 tool in
build-jenkins also already exists.

### DIFF
--- a/platform/jobs/RetirementJobs.groovy
+++ b/platform/jobs/RetirementJobs.groovy
@@ -174,6 +174,7 @@ job('user-retirement-driver') {
             name('user-retirement-driver')
             nature('shell')
             command(readFileFromWorkspace('platform/resources/user-retirement-driver.sh'))
+            pythonName('PYTHON_3.5')
         }
     }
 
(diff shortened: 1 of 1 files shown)


platform/jobs/RetirementJobs.groovy
```diff
@@ -174,6 +174,7 @@ job('user-retirement-driver') {
             name('user-retirement-driver')
             nature('shell')
             command(readFileFromWorkspace('platform/resources/user-retirement-driver.sh'))
+            pythonName('PYTHON_3.5')
```
doctoryes: Tubular is tested under Python 3.6 as well as Python 3.5:
https://github.com/edx/tubular/blob/410cb3f11b05386ab14a675bc851cfdb08faab7b/tox.ini#L2
Maybe instead use `System-CPython-3.6`?

Or update tubular to test using Python 3.7 - and use `PYTHON_3.7`?

PR #875, opened 2020-01-15
Add missing parameters to Snowflake Spreadsheet loading job.

pwnage101:
DE-1919

---

edx-analytics-pipeline: https://github.com/edx/edx-analytics-pipeline/pull/794/
analytics-secure: https://github.com/edx-ops/analytics-secure/pull/327 (merged)
analytics-secure x2: https://github.com/edx-ops/analytics-secure/pull/328
jenkins-job-dsl: This PR

### DIFF
--- a/dataeng/jobs/analytics/LoadGoogleSpreadsheetToSnowflake.groovy
+++ b/dataeng/jobs/analytics/LoadGoogleSpreadsheetToSnowflake.groovy
@@ -17,6 +17,9 @@ class LoadGoogleSpreadsheetToSnowflake {
             parameters to_date_interval_parameter(allVars)
             parameters {
                 stringParam('GOOGLE_CREDENTIALS', allVars.get('GOOGLE_CREDENTIALS'))
+                stringParam('SNOWFLAKE_CREDENTIALS', allVars.get('SNOWFLAKE_CREDENTIALS'), 'The path to the Snowflake user credentials file.')
+                stringParam('SNOWFLAKE_WAREHOUSE', allVars.get('SNOWFLAKE_WAREHOUSE'), 'The warehouse to use for loading data into Snowflake.')
+                stringParam('SNOWFLAKE_ROLE', allVars.get('SNOWFLAKE_ROLE'), 'The role to use for loading data into Snowflake.')
             }
             multiscm common_multiscm(allVars)
             triggers common_triggers(allVars)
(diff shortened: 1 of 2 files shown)


dataeng/jobs/analytics/LoadGoogleSpreadsheetToSnowflake.groovy
```diff
@@ -17,6 +17,10 @@ class LoadGoogleSpreadsheetToSnowflake {
             parameters to_date_interval_parameter(allVars)
             parameters {
                 stringParam('GOOGLE_CREDENTIALS', allVars.get('GOOGLE_CREDENTIALS'))
+                stringParam('SNOWFLAKE_CREDENTIALS', db_config.get('SNOWFLAKE_CREDENTIALS', allVars.get('SNOWFLAKE_CREDENTIALS')), 'The path to the Snowflake credentials file.')
+                stringParam('SNOWFLAKE_WAREHOUSE', db_config.get('SNOWFLAKE_WAREHOUSE', allVars.get('SNOWFLAKE_WAREHOUSE')), '')
+                stringParam('SNOWFLAKE_ROLE', db_config.get('SNOWFLAKE_ROLE', allVars.get('SNOWFLAKE_ROLE')))
+                stringParam('SNOWFLAKE_DATABASE', db_config.get('SNOWFLAKE_DATABASE', allVars.get('SNOWFLAKE_DATABASE')))
```
brianhw: Looks to me like this value only comes from config, not from the command line.

dataeng/resources/load-google-spreadsheet-snowflake.sh
```diff
@@ -4,5 +4,10 @@ env
 ${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
  LoadGoogleSpreadsheetsToSnowflakeWorkflow --local-scheduler \
  --date $(date +%Y-%m-%d -d "$TO_DATE") \
+ --sf-credentials $SNOWFLAKE_CREDENTIALS \
+ --sf-warehouse $SNOWFLAKE_WAREHOUSE \
+ --sf-role $SNOWFLAKE_ROLE \
+ --sf-database $SNOWFLAKE_DATABASE \
```
brianhw: Where is this coming from?  And why is it passed if it is not defined in Luigi?

brianhw: That said, it would be nice in general to be able to override values that were specified in config files, rather than hardcoding the gets to only pull values from config files.  It may matter less for this specific workflow, but in general when testing or doing a copy from Vertica, one may want to point to a more ad-hoc location in Snowflake.  Like now, where I'm working on loading dbt validation datasets.   (That may need to be added to those jobs....) .  But just for the record....

pwnage101: Hm, I thought I fixed this already, probably didn't add it to the git index.  Not sure how this even worked when I tested it.

pwnage101: To your second point, I also do not like that we are using prod-edx.cfg for this job, but the intention of the ticket was to re-use the same "code" from the vertica job to minimize difficulties.  In retrospect this ticket was already so damn difficult that I could have just re-done the config.

PR #887, opened 2020-01-31
Assorted OEP-18 job updates

jmbowman:
Updated the reviewers and emails for several of the `make upgrade` jobs, and added 3 more for repositories that should already be otherwise OEP-18 compliant.  Switched the ecommerce job to Python 3.5 as that repository is about to drop 2.7 support.


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -107,8 +127,8 @@ Map edxOrganizations = [
     pythonVersion: '2.7',
     cronValue: '@weekly',
     githubUserReviewers: ['feanil'],
```
robrap: Remove feanil?

PR #895, opened 2020-02-10
MICROBA-104 | Add new Jenkins job for MB ITK script

justinhynes:
Add new Jenkins job for MB ITK script

### DIFF
--- a/dataeng/resources/snowflake-microbachelors-send-coaching-data-itk.sh
+++ b/dataeng/resources/snowflake-microbachelors-send-coaching-data-itk.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -ex
+
+# setup
+cd $WORKSPACE/analytics-tools/snowflake
+pip install -r requirements/microbachelors.txt
+
+# Run the script twice. Once for the student report, then again for the Course report
+python send_coaching_data_itk.py \
+    --key_path $KEY_PATH \
+    --passphrase_path $PASSPHRASE_PATH \
+    --user $USER \
+    --account $ACCOUNT \
+    --report-type student \
+    --send True
+
+python send_coaching_data_itk.py \
+    --key_path $KEY_PATH \
+    --passphrase_path $PASSPHRASE_PATH \
+    --user $USER \
+    --account $ACCOUNT \
+    --report-type course \
+    --send True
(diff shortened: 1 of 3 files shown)


dataeng/resources/snowflake-microbachelors-send-coaching-data-itk.sh
```diff
@@ -0,0 +1,23 @@
...
+pip install -r requirements/microbachelors.txt
+
+# Run the script twice. Once for the student report, then again for the Course report
+python send_coaching_data_itk.py \
+    --key_path $KEY_PATH \
+    --passphrase_path $PASSPHRASE_PATH \
+    --user $USER \
+    --account $ACCOUNT \
+    --report-type student \
+    --send True
```
crice100: Should we default both _send_ params to _False_ for now?

justinhynes: Yeah, that makes sense until we are ready to go.

dataeng/resources/snowflake-microbachelors-send-coaching-data-itk.sh
```diff
@@ -0,0 +1,23 @@
...
+    --account $ACCOUNT \
+    --report-type student \
+    --send False
+
+python send_coaching_data_itk.py \
+    --key_path $KEY_PATH \
+    --passphrase_path $PASSPHRASE_PATH \
+    --user $USER \
+    --account $ACCOUNT \
+    --report-type course \
```
bmedx: If this fails for some reason and needs to be re-run, will it be clear to the downstream consumers which files they should use? Will someone manually have to go delete the previously generated files from S3 to prevent confusion? 

Actually, I'm not seeing where these files actually get uploaded. Currently it looks like they get generated to the local Jenkins filesystem and just left there?

justinhynes: We broke the stories into a couple of chunks. This was just the first piece of the work to get the ball rolling. We don't have the creds/info from InsideTrack to know where we are uploading these files, but this work is captured in MICROBA-173.

Similarly, the cleanup of the generated reports is captured in MICROBA-176 and is going to be a fast follow up to this PR. 

We wouldn't be launching our coaching integration without either of these stories being finished. Let me know if you have any other concerns or want to chat further!

PR #905, opened 2020-02-25
Separate out source tests from tag tests

bmedx:


dataeng/resources/warehouse-transforms.sh
```diff
@@ -12,5 +12,10 @@ cd $WORKSPACE/warehouse-transforms/warehouse_transforms_project
 dbt clean
 dbt deps
 dbt seed --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
+
+# Run the source tests first, sadly not just the ones upstream from this tag
```
pwnage101: so close!  dbt just needs to create a full blown query language to expand their selection syntax.

bmedx: I thought I could do `--models +tag:$MODEL_TAG --exclude tag:$MODEL_TAG` but alas, no.

PR #907, opened 2020-02-26
MICROBA-176 | Archive generated reports, limit to last 7 days

justinhynes:
Use common_logrotator which is set up to keep the last 7 days of the job runs. Add a step to archive the reports we are generating (the CSV files).

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
+++ b/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
@@ -5,7 +5,7 @@ that queries Snowflake for MicroBachelors learner data and transmits it to edX's
 
 package analytics
 
-import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.DevopsConstants.common_logrotator
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
 import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
@@ -17,7 +17,7 @@ class SnowflakeMicrobachelorsITK {
 
     public static def job = { dslFactory, allVars ->
         dslFactory.job('snowflake-microbachelors-send-coaching-data-itk') {
-            logRotator common_log_rotator(allVars)
+            logRotator common_logrotator
             authorization common_authorization(allVars)
             parameters secure_scm_parameters(allVars)
             parameters {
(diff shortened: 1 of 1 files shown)


justinhynes:
> One question what about this change limits to last 7?
@staubina I found and imported a different utility function. It is  named _very_ similarly but defaults to 'daysToKeep(7)'. See https://github.com/edx/jenkins-job-dsl/blob/master/src/main/groovy/org/edx/jenkins/dsl/DevopsConstants.groovy#L11

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -5,7 +5,7 @@ that queries Snowflake for MicroBachelors learner data and transmits it to edX's
 
 package analytics
 
-import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.Constants.common_logrotator
```
brianhw: I tried seeding with this branch, but got an error:
workspace:/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy: 8: unable to resolve class org.edx.jenkins.dsl.Constants
 @ line 8, column 1.
   import static org.edx.jenkins.dsl.Constants.common_logrotator

brianhw: Ah, based on your previous comment, you wanted DevopsConstants, not Constants.

justinhynes: Thanks. I wasn't sure if there was a good way to test this locally :-O. 

I just pushed the update to (hopefully) fix the import. Thanks for the help testing this out.

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -49,7 +48,14 @@ class SnowflakeMicrobachelorsITK {
             wrappers {
                 timestamps()
             }
-            publishers common_publishers(allVars)
+            publishers {
+                archiveArtifacts {
+                    // job normally will generate two unique CSV files, each ending with timestamps as part of the file name
+                    pattern('*.csv')
```
brianhw: When I created the job, it provided this warning:  

  | ‘*.csv’ doesn’t match anything, but ‘analytics-tools/snowflake/*.csv’ does. Perhaps that’s what you mean?

justinhynes: Ah, yeah. A misunderstanding from my end. Looks like I made a bad assumption from where the archive step action would occur since the bash script changes directory into the analytics-tools/snowflake directory when executed. 

I'll push an update and thanks again for your help and patience!

PR #908, opened 2020-02-26
Load vertica schema to SF using metadata.

brianhw:
Use LoadVerticaSchemaFromS3WithMetadataToSnowflakeTask instead of LoadVerticaSchemaFromS3ToSnowflakeTask. 
Since this uses metadata in S3 instead of querying Vertica, we no longer need VERTICA_CREDENTIALS or EXCLUDE. 

Just tested (again) using http://jenkins.analytics.edx.org/job/load-vertica-schema-to-snowflake-financial_reporting/108.


brianhw:
It has been around since the loading of the Vertica reference into Snowflake.  I didn't switch the daily runs to use it at the time, but now that we got a failure that could have been avoided (and indeed was fixed) by using it, I thought I'd create this PR to put it in place.   There is similar code for  read replicas that was implemented and could also be switched to, but I wanted to keep the two improvements separate.

PR #910, opened 2020-03-02
Add automatic make upgrade job for edx-val.

feanil:


feanil:
This repo was behind when we hit it for doprecation work so I'm just adding this automation here.

feanil:
Will merge once https://github.com/edx/edx-val/pull/203 lands which should make the upgrade clean.

feanil:
Job is running now.  https://build.testeng.edx.org/job/edx-val-upgrade-python-requirements/2/console

PR #912, opened 2020-03-02
create view for dbt jobs

estute:
The current dbt view on analytics jenkins was manually created. Just codifying this into dsl.

### DIFF
--- a/dataeng/jobs/createJobs.groovy
+++ b/dataeng/jobs/createJobs.groovy
@@ -237,3 +237,12 @@ listView('Backups') {
     }
     columns DEFAULT_VIEW.call()
 }
+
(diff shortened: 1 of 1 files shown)


dataeng/jobs/createJobs.groovy
```diff
@@ -237,3 +237,14 @@ listView('Backups') {
     }
     columns DEFAULT_VIEW.call()
 }
+
+listView('dbt') {
+    jobs {
+        name('dbt-docs')
+        name('dbt-manual')
+        name('dbt-source-freshness')
```
brianhw: nit: you could use a regex for these, and that might generalize well.

PR #920, opened 2020-03-17
Run `make upgrade` for `opaque-keys` with Python 3.5, not 2.7

timmc-edx:


timmc-edx:
Not going to remove python 2.7 for now.

PR #921, opened 2020-03-17
Add an optional config argument for installing system packages.

dianakhuang:
The issue with `make upgrade` for `openedxstats` is that we're seeing the issue `Error: pg_config executable not found.` when trying to install `psycopg2`.

This executable is provided by the `libpq-dev` Ubuntu package, which needs to be installed via Ubuntu package manager.


dianakhuang:
Based on our conversation, we are planning on installing this on all jenkins workers because this strategy runs into security issues.

PR #924, opened 2020-03-20
Schedule MB ITK job to run everyday at 9am

tuchfarber:
The existing `common_triggers` schedules a cron job given an environment config, but we don't have one in this job, so we're just scheduling manually.


tuchfarber:
Updated schedule because epic requested Monday-Friday only.

PR #928, opened 2020-03-24
update upgrade jobs

robrap:
Before merge:
- [x] Squash commits

This PR has two independent changes:
1. Move testeng-ci to devops.
2. Update the cronValues.

---------------

Switched cronValue from:
  `@weekly`
to:
  `* H(4-11) * * H(0-4)`

This will choose a random time between midnight-7:59am ET,
Sunday-Thursday, which will have this run mostly off hours (depending on
time zone), and new PRs should be available in the morning during the
week with sufficient time to merge and deploy.

Notes:
- The H symbol can be thought of as a random value over a range, but
it actually is a hash of the job name, not a random function, so that
the value remains stable for any given project.
- Times are in UTC.

For help text for Jenkins cron settings, see:
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -11,11 +11,15 @@ package testeng
 //     emails: List of emails that should be notified when job fails, e.g.: ['email1', 'email2']
 // ]
 
+// Cron values for: midnight-7:59am ET, Sunday-Thursday (times are in UTC)
+// See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly
+def cronOffHoursBusinessWeek = '* H(4-11) * * H(0-4)'
+
 Map apiDocTools = [
     org: 'edx',
     repoName: 'api-doc-tools',
     pythonVersion: '3.5',
-    cronValue: '@weekly',
+    cronValue: cronOffHoursBusinessWeek,
     githubUserReviewers: [''],
     githubTeamReviewers: ['teaching-and-learning'],
     emails: ['ned@edx.org']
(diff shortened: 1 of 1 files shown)


feanil:
The cron is going to execute in UTC, so you probably want to add 4 to the hour of day.

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -15,7 +17,7 @@ Map apiDocTools = [
     org: 'edx',
     repoName: 'api-doc-tools',
     pythonVersion: '3.5',
-    cronValue: '@weekly',
+    cronValue: '* H(4-11) * * H(0-4)',
```
feanil: You can extract this variable to a string above this map and then just refer to that string in all the places.
https://groovy-lang.org/syntax.html#all-strings

```
weekly = '* H(4-11) * * H(0-4)'
```

robrap: Wise.  Done.

robrap:
Although Travis CI is not reporting back to Github, the two builds both passed:
* https://travis-ci.org/github/edx/jenkins-job-dsl/builds/666505979
* https://travis-ci.org/github/edx/jenkins-job-dsl/builds/666505926
I am going to merge.

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -11,11 +11,15 @@ package testeng
 //     emails: List of emails that should be notified when job fails, e.g.: ['email1', 'email2']
 // ]
 
+// Cron values for: midnight-7:59am ET, Sunday-Thursday (times are in UTC)
+// See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly
+def cronOffHoursBusinessWeek = '* H(4-11) * * H(0-4)'
```
waheedahmed: I believe the cron value should be `H H(4-11) * * H(0-4)`, currently it's running the job every minute for that random hour.
Screen Shot 2020-03-25 at 2 08 11 PM

waheedahmed: Made a fix for this https://github.com/edx/jenkins-job-dsl/pull/933.

robrap: Thanks again @waheedahmed.

PR #933, opened 2020-03-25
Fix python upgrade cron value.

waheedahmed:
Currently it's running every minute for a random single hour.
Screen Shot 2020-03-25 at 2 08 11 PM


robrap:
Thanks @waheedahmed!

PR #939, opened 2020-03-27
WIP: Update ITK job to use S3 credentials

bmedx:


dataeng/resources/snowflake-microbachelors-send-coaching-data-itk.sh
```diff
@@ -5,6 +5,9 @@ set -ex
 cd $WORKSPACE/analytics-tools/snowflake
 pip install -r requirements/microbachelors.txt
 
+# download the SFTP credentials from S3
+aws s3 cp $SFTP_CREDENTIALS_BUCKET $WORKSPACE/
```
justinhynes: Do we have to specify the filename anywhere... that we are trying to copy `itk_sftp.json` from s3 to the workspace? Or is the file/path built into the value `$SFTP_CREDENTIALS_BUCKET`? Sorry, this looks to be syntax for the aws cli which I'm not familiar with.

bmedx: Yeah the naming convention is weird for that variable, but seems to match what we've got for other S3 files. The value coming from secure will be SFTP_CREDENTIALS_BUCKET: "s3://edx-analytics-secure/itk_sftp.json"

pwnage101: Nit: does it need _BUCKET as a suffix, or can we lop that off here and in analytics-secure?

PR #946, opened 2020-04-06
MICROBA-184 | Setup OpsGenie alerts for coaching Jenkins job

justinhynes:
[MICROBA-184]
- Update coaching Jenkins job to send mails if the job fails to new Aperture mailing group/OpsGenie team
- Noticed a warning in the Jenkins configuration for building periodically settings. Adjusted cron schedule to follow the guidance
- Update default value of `$NOTIFY` for the warehouse-transforms MicroBachelors job to alert us first instead of the DE team

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
+++ b/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
@@ -23,6 +23,7 @@ class SnowflakeMicrobachelorsITK {
             parameters {
                 stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
                 stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY_APERTURE'), 'Space separated list of emails to send notifications to.')
             }
             environmentVariables {
                 env('KEY_PATH', allVars.get('KEY_PATH'))
@@ -48,7 +49,7 @@ class SnowflakeMicrobachelorsITK {
                 }
             }
             triggers {
-                cron('0 6 * * 1-5')
+                cron('H 6 * * 1-5')
             }
             wrappers {
                 timestamps()
@@ -61,6 +62,7 @@ class SnowflakeMicrobachelorsITK {
                     defaultExcludes()
                 }
             }
+            publishers common_publishers(allVars)
             steps {
                 virtualenv {
                     pythonName('PYTHON_3.7')
--- a/dataeng/jobs/analytics/WarehouseTransforms.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransforms.groovy
@@ -22,7 +22,12 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('TEST_SOURCES_FIRST', env_config.get('TEST_SOURCES_FIRST', 'true'), 'Set to \'true\' to perform source testing first. All other values test sources post-run.')
-                    stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                    if (environment.equalsIgnoreCase("microbachelors")) {
+                        // send alerts for the microbachelors data-mart warehouse-transforms job to aperture team first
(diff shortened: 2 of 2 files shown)


dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -48,7 +49,7 @@ class SnowflakeMicrobachelorsITK {
                 }
             }
             triggers {
-                cron('0 6 * * 1-5')
+                cron('H 6 * * 1-5')
```
justinhynes, line 52: Unrelated change but I noticed it when I was looking at the configuration in the UI:
image

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -55,4 +60,3 @@ class WarehouseTransforms{
         }
     }
 }
-
```
justinhynes, line 58: This looks strange. There were two blank lines at the end of this file. I removed one, but it looks like there are none now? Either way, if it matters, there should still be one blank line hah.

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -23,6 +23,7 @@ class SnowflakeMicrobachelorsITK {
             parameters {
                 stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
                 stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY_APERTURE'), 'Space separated list of emails to send notifications to.')
```
brianhw, line 26: Sorry to be a little confused here.   Is there a reason to specify the Aperture pager email address as an environment variable on the Jenkins server, rather than as a configuration parameter in the job.   This has an allVars.get('NOTIFY'), so it should get the value from the job settings in analytics-secure.   But that should be enough, in my view, rather than wiring a particular transient team name up through Jenkins and the configuration repo.   (And just clear the default here, I guess.)   And if you include publishers, would you need this line at all anymore?

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -22,7 +22,12 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('TEST_SOURCES_FIRST', env_config.get('TEST_SOURCES_FIRST', 'true'), 'Set to \'true\' to perform source testing first. All other values test sources post-run.')
-                    stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                    if (environment.equalsIgnoreCase("microbachelors")) {
+                        // send alerts for the microbachelors data-mart warehouse-transforms job to aperture team first
+                        stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY_APERTURE'), 'Space separated list of emails to send notifications to.')
+                    } else {
+                        stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                    }
```
brianhw, line 30: Similarly, I would change the logic here to check env_config.get('NOTIFY', allVars.get('NOTIFY', '$PAGER_NOTIFY)), and then put the aperture pager in as the value of NOTIFY in the definition of the microbachelors environment.

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
brianhw: I understand your motivation entirely.  I'm just concerned whether following the example of PAGER_NOTIFY is a good way to go for DE and non-DE groups alike.  And I'm checking with DE on this.   But so far we have Enterprise jobs having their NOTIFY value set to "enterprise-integrations@edx.org", and program enrollment report failures being sent to "program-analytics@edx.opsgenie.net", using NOTIFY values in analytics-secure.  So there is already precedence.   

If there is a concern about cut-and-paste, then we can define a value similar to the value of ANALYTICS_EXPORTER_NOTIFY_LIST, which is also defined in a common.yaml file in analytics-secure and shared across several exporter-related jobs.    The downside of this approach is that it has to be fetched as allVars.get('ANALYTICS_EXPORTER_NOTIFY_LIST'), so it has to be done in the groovy code.   This is fine for exporter, where the notifications are the same independent of environment, but it doesn't work for per-environment configuration.    It means having to use your approach below of moving per-environment configuration up into the job definition.  

Either way, I'd prefer not to use the Jenkins environment variable approach.   And hope to confirm that with DE tomorrow morning.

brianhw: Yes, the problem is compounded by the fact that you don't have visibility into the analytics-secure repo.

justinhynes:
Closing this PR. There was a bit of rework on this with DE's help and the changes are covered in https://github.com/edx/jenkins-job-dsl/pull/948

PR #947, opened 2020-04-07
Always run python and quality checks under Django 2.2 as well

timmc-edx:
(BOM-1509)


timmc-edx:
Ah right, need to actually list them. :-)

I wasn't clear on a11y, since we haven't had Django 2.2 configured for those yet even as an optional run. (Nor for JS.)

PR #948, opened 2020-04-07
DESUPPORT-222: Add aperture notify to MB jobs.

brianhw:
This is intended to replace #946 .

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransforms.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransforms.groovy
@@ -22,7 +22,7 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('TEST_SOURCES_FIRST', env_config.get('TEST_SOURCES_FIRST', 'true'), 'Set to \'true\' to perform source testing first. All other values test sources post-run.')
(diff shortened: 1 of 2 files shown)


dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -22,7 +22,7 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('TEST_SOURCES_FIRST', env_config.get('TEST_SOURCES_FIRST', 'true'), 'Set to \'true\' to perform source testing first. All other values test sources post-run.')
-                    stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                    stringParam('NOTIFY', env_config.get('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY')), 'Space separated list of emails to send notifications to.')
```
pwnage101, line 25: Just to be clear, this change together with https://github.com/edx-ops/analytics-secure/pull/366/files#diff-f03bd4a3367b65a270bcb3d8ee4ac191R32 results in the microbachelors warehouse-transforms job no longer alerting Data Engineering.  is that expected?

brianhw: That was the requested behavior.  We have other workflows, such as enterprise and program reporting, that use NOTIFY to override notification of DE, so I assumed this would be in the same category.    Do you think it should be different?

PR #957, opened 2020-04-23
Add stitch-snowflake-lag-monitor job

bmedx:


dataeng/jobs/analytics/StitchSnowflakeLagMonitor.groovy
```diff
@@ -0,0 +1,63 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+
+
+
+class StitchSnowflakeLagMonitor {
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("stitch-snowflake-lag-monitor") {
```
doctoryes, line 16: Does the `SCHEMA` need to be added to the name of this job - in order to support other jobs monitoring other schemas?

brianhw: Ah, as in an "environment" for each schema?   That seems not unreasonable.

bmedx: I had not expected this to be a full implementation since we were trying to solve a production issue, but since it's already taken too long to be useful for that I'll update everything to make it actually reusable.

PR #980, opened 2020-05-12
INFRA-144 | flag file/docker shim conditional

christopappas:


devops/resources/run-ansible.sh
```diff
@@ -44,5 +44,13 @@ if [[ -n "${BECOME_USER}" ]]; then
   ANSIBLE_BECOME=" --become --become-user=${BECOME_USER} "
 fi
 
+#SOME_THING_THAT_CHECKS_FOR_FLAG_FILE
+# only run docker shim command if 1) flag file exists 2) the shim args exist
+if [[ SOME_CONDITIONAL_HERE && -n "${ANSIBLE_MODULE_ARGS_SHIM}" ]]; then
```
coryleeio, line 49: NIT If the flag file exists but the shim commands do not, go ahead and blow up so it will be obvious we forgot to define the shim

devops/resources/run-ansible.sh
```diff
@@ -44,5 +44,17 @@ if [[ -n "${BECOME_USER}" ]]; then
   ANSIBLE_BECOME=" --become --become-user=${BECOME_USER} "
 fi
 
+# Test if docker shim flag file is present
+shim_enabled=true
 ansible ${ANSIBLE_PATTERN} ${ANSIBLE_INVENTORY} -u ${ANSIBLE_SSH_USER} ${ANSIBLE_BECOME} -m ${ANSIBLE_MODULE_NAME} \
--a "${ANSIBLE_MODULE_ARGS}"
+-a 'test -f /edx/etc/docker_shim_enabled' || echo "Docker shim is not enabled."; shim_enabled=false
```
coryleeio, line 50: Is there a way to do this in a bash if condition or does it need to be passed in as module args to ansible?

coryleeio: Oh - answered my question, it's on the box we are reaching out to, got it.

coryleeio: Can you update the comment here to give an indication that it's the remote box being reached out to

PR #986, opened 2020-05-15
Estute/deng 210

estute:
As part of the cutover process from the old DE jenkins server to the new one, I have created a new seed job. This should be run on the new server as jobs are ported. Additionally, there are three new views added to the old seed job, which will help us organize which jobs have been ported, which still need to be, and which ones are custom jobs, not codified in DSL.

### DIFF
--- a/dataeng/jobs/createJobsNew.groovy
+++ b/dataeng/jobs/createJobsNew.groovy
@@ -1,0 +1,130 @@
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.DEFAULT_VIEW
+import org.yaml.snakeyaml.Yaml
+import org.yaml.snakeyaml.error.YAMLException
+
+def globals = binding.variables
+String commonVarsDir = globals.get('COMMON_VARS_DIR')
+String commonVarsFilePath = commonVarsDir + 'common.yaml'
+Map commonConfigMap = [:]
+
+try {
+    out.println('Parsing secret YAML file')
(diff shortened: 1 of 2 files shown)


dataeng/jobs/createJobsNew.groovy
```diff
@@ -0,0 +1,130 @@
...
+    Yaml yaml = new Yaml()
+    commonConfigMap = yaml.load(commonConfigContents)
+    out.println('Successfully parsed secret YAML file')
+
+}  catch (YAMLException e) {
+    throw new IllegalArgumentException("Unable to parse ${commonVarsFilePath}: ${e.message}")
+}
+
+def taskMap = [
+    // Add jobs here as they are ported from the old analytics Jenkins server.
```
pwnage101, line 22: Let me know if this is out of scope, but i was wondering if we could take this opportunity to make this new seed job more like a conventional seed job that only seeds one file at a time.  Would that require a major overhaul of how we write our job files?

pwnage101: The reason I ask is because your subteam was spun off for quality of life improvements, and being able to simply test a job DSL without having to create a temp commit to comment out this taskMap would increase my quality of life.

estute: I would like to do this, but we should think it through. For the sake of what I am doing right now (trying to get a box up asap to run warehouse transform ci), I would consider it out of scope, but please put in a ticket so we can circle back to address it.

PR #992, opened 2020-05-21
begin de jenkins cutover process

estute:


estute:
@bmedx the snowflake job creates an 'Expire 1 password' job, which I can test against myself

estute:
ah, what the heck. I'll add emr-cost-reporter, as it is fairly low-stakes

bmedx:
:shipit:

PR #994, opened 2020-05-27
DE-222: Add second job to refresh snowpipe for segment events.

brianhw:


brianhw:
Yeah, I was hoping that DELAY was in minutes, but I see that it's in seconds.    The last two alter pipe refresh runs I made took 2h 29m and 2h 28m respectively.   So I assume this should be upped to at least 150m*60s/m = 9000 seconds.

pwnage101:
Hm, It also seems that the job is hard-coded to run hourly: https://github.com/edx-ops/analytics-secure/blob/master/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS.yaml#L8
Should we split cron into the local job config in this PR?

brianhw:
Actually, I now think this should wait a week to turn it on, so that the copy history that pulls the last seven days of copies isn’t overwhelmed by the data load.    Not sure if the job is doing what we want.    If I understand correctly, the current job is pulling history over the last week, and outputting success if there were seven events in that week time period.   That's not particularly helpful.  But bigger thing it's doing is looking for duplicates.  Did this ever happen much?

brianhw:
Actually, the job is hardcoded to run daily.  With an "H" parameter that would stagger the two jobs' start times within that hour.

brianhw:
Turns out the DELAY kicks in *after* the pipe refresh call returns, so it doesn't matter whether that takes two minutes or two-plus hours -- the script is going to wait.  See http://jenkins.analytics.edx.org/view/all/job/test-refresh-snowpipe-segment-event-loader/2/console.

PR #996, opened 2020-05-28
Make tests and seeding optional for warehouse transforms jobs

bmedx:


dataeng/resources/warehouse-transforms.sh
```diff
@@ -11,10 +11,14 @@ cd $WORKSPACE/warehouse-transforms/projects/reporting
...
-dbt seed --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
+
+if [ $SKIP_SEED = 'true' ]
+then
+  dbt seed --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
+fi
 
 # Source testing *before* model-building can be enabled/disabled with this envvar.
-if [ $TEST_SOURCES_FIRST = 'true' ]
+if [ $TEST_SOURCES_FIRST = 'true' ] && [ $SKIP_TESTS != 'true' ]
```
doctoryes, line 21: Does anything use `TEST_SOURCES_FIRST` after this change? Can/should it be removed in this PR?

bmedx: Yes one job uses it, and it seems like a handy thing to keep around.

PR #1001, opened 2020-06-05
Add k8s Apps to sandbox job (SRE-59)

nadeemshahzad:


devops/jobs/CreateSandbox.groovy
```diff
@@ -262,6 +275,9 @@ class CreateSandbox {
                 booleanParam("run_oauth",true,"")
 
                 stringParam("nginx_users",'[{"name": "{{ COMMON_HTPASSWD_USER }}","password": "{{ COMMON_HTPASSWD_PASS }}","state":"present"}]',"")
+
+                booleanParam("argocd",false,"This will install the ArgoCD in minikube in the sandbox")
```
adzuci, line 279: If you want to you can remove "the" in front of "ArgoCD"

devops/jobs/CreateSandbox.groovy
```diff
@@ -94,6 +94,18 @@ class CreateSandbox {
                         relativeTargetDirectory('configuration-internal')
                     }
                 }
+                git {
```
adzuci, line 97: @coryleeio pointed out that `configuration-internal` above is actually `edx-internal` so we shouldn't be checking it out twice.

Cory:
```
I also think that configuration-internal and configuration-secure, might actually be edx-internal and
edx-secure at a different level of abstraction and that adding edx-internal there may not be necessary, 
you have to remember that code is in configuration so it’s open source so relying on edx-internal directly 
isn’t appropriate
```

@nadeemshahzad could your logic point at the `configuration-internal` folder instead?

adzuci (changes requested):
I think some of this may not be needed.

nadeemshahzad:
Closing as we don't want to install ArgoCD inside minikube on Sandox.

devops/jobs/CreateSandbox.groovy
adzuci: @coryleeio I had misunderstood this and thought that `configuration-internal` was `edx-internal`, but it is in fact https://github.com/edx/sandbox-internal so we either need to:

1: Copy the four argocd django application folders out of edx-internal and into sandbox-internal.
2: Clone edx-internal onto the box.

adzuci: @nadeemshahzad would like to try #1 first since there is less of a risk of leaking remote config data.

adzuci (approved):
:shipit:

devops/jobs/CreateSandbox.groovy
adzuci: That sounds good to me.

PR #1004, opened 2020-06-08
Dry-run ownership reconciliation script (using spreadsheet export)

timmc-edx:
- Seems to be working and multi-org
- Needs Google integration
- Suggests deleting a bunch of repos that my token doesn't have permission to see, but that probably exist

This is a first pass for ARCHBOM-1207.


ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+
+        groupkey = lambda a: (a['action'], a['why'])
+        for ((action_name, why), group) in groupby(sorted(actions, key=groupkey), groupkey):
+            print()
+            print("Action: %s -- %s" % (action_name, why))
+            for action in group:
+                data = action.copy()
+                del data['action']
+                del data['name']
+                del data['why']
```
timmc-edx, line 187: There has to be a better way to disassociate multiple keys from a copy of dictionary, right? In Clojure this would be `(dissoc data :action :name :why)` but I can't figure out a nice equivalent for Python.

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+
+import click
+import csv
+import github3
+from github3.exceptions import NotFoundError
+from itertools import groupby
+import re
+
+def bail(message):
+    raise KnownError(message)
```
timmc-edx, line 17: I can probably get rid of `bail` as a layer of indirection. I think it was originally going to be a little fancier.

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+            'why': why,
+            **(more_data or {})
+        })
+        names_only_ss.remove(ss_name)
+
+    # First, if a repo is only in the spreadsheet, check if it was
+    # moved or archived
+    for ss_name in names_only_ss.copy():
+        try:
+            found_repo = hub.repository(*ss_name.split('/', 2))
```
timmc-edx, line 113: Is this too opaque? I can't tell.

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+
+    - name: Lowercased full name of repository
+    - row_id: Spreadsheet row number
+
+    Raises an exception if any entries are not in provided orgs.
+    """
+    if spreadsheet_csv:
+        with open(spreadsheet_csv, 'r') as csvf:
+            reader = csv.reader(csvf)
+            # Figure out which columns are which
```
timmc-edx, line 221: I'm sure there's a better way to do this, but we'll be switching to reading from Google's Spreadsheet API instead, so whatever.

ownership/sync-repos.py
feanil: Agree, let's get rid of this indirection.

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+def bail(message):
+    raise KnownError(message)
+
+class KnownError(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+        self.message = message
+
+@click.command()
+@click.option(
```
feanil, line 25: When we insert this as a jenkins job step, we'll want to be able to pass in the other settings as environment variables.  can you set it so that all these options can also be passed in via the environment: https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables

ownership/sync-repos.py
feanil: Are you passing in the org and repo name to hub?  I think a comment would be good.

ownership/sync-repos.py
feanil: Why does this need to be done?

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+        if not row_id:
+            continue
+        if row_id in seen_row_ids:
+            bail("Possible bug in script: Was going to suggest multiple "
+                 "actions for row %s: %s"
+                 % (row_id, [r for r in actions if r.get('row_id') == row_id]))
+        seen_row_ids.add(row_id)
+
+    # For now, just make recommendations, don't make changes.
+    if actions:
```
feanil, line 176: Should this be split out from the run-compare?  It seems like there is a nice seam here where the compare outputs the diff + actions and than what we do with them can be de-coupled? Might be a premature optimization though so fine to leave it as is too.

ownership/sync-repos.py
feanil: The dict reader is good for that and then you don't have to care so much about column indices but agree that you don't need to fix it since this is temporary.

ownership/sync-repos.py
```diff
@@ -0,0 +1,255 @@
...
+        run_compare(hub, orgs, repos_gh, repos_ss)
+    except KnownError as e:
+        print(e.message)
+
+
+def run_compare(hub, orgs, repos_gh, repos_ss):
+    # List of actions to recommend at end of run.
+    #
+    # This list will *grow* as the comparison progresses.
+    actions = []
```
feanil, line 58: Can you also document that each action is a dict and what the keys in that dict mean here?

ownership/sync-repos.py
timmc-edx: OK. I'll just name the parts and pass them in explicitly.

ownership/sync-repos.py
timmc-edx: There's extra data associated with some action types, but not all. This removes the keys that are always present, since they're printed explicitly, and then `data` can be listed out if there's anything left. Sample output:

```
Action: delete_row -- Moved to a Github org not in the scan list
- edx/cookiecutter-django-app {'row_id': 26, 'new_location': 'edx-unsupported/cookiecutter-django-app'}
- edx/edx-oauth2-provider {'row_id': 25, 'new_location': 'edx-unsupported/edx-oauth2-provider'}

Action: delete_row -- Repo has been archived
- edx/eslint-config-edx {'row_id': 244}

Action: move_repo_unsupported -- Repo has been archived but is still in a main org rather than edx-unsupported
- edx/edx-ora
```

If each action type were handled by a different code branch, I would pull out relevant attributes explicitly for display, but this was a simpler way. This is probably worth refining if called for, since I bet we'll want a `--dry-run` flag even once this can make changes.

Incidentally, I did figure out a way of doing this as a one-liner, but I'm not sure if it would make people sad:

```python
data = {k: v for k, v in action.items() if k not in {'action', 'name', 'why'}}
```

ownership/sync-repos.py
timmc-edx: Agreed, this really shouldn't be in `run_compare`, but rather `actions` should be returned to `main`, and then passed to another function that actually makes changes. Easy enough to make the split right now, though.

timmc-edx:
@feanil Ready for re-review. Also probably going to rename sync-repos.py to sync_repos.py before squashing, to match naming conventions.

ownership/sync_repos.py
```diff
@@ -0,0 +1,286 @@
...
+    if actions:
+        print("Recommendations:")
+
+        groupkey = lambda a: (a['action'], a['why'])
+        for ((action_name, why), group) in groupby(sorted(actions, key=groupkey), groupkey):
+            print()
+            print("Action: %s -- %s" % (action_name, why))
+            for action in group:
+                data = action.copy()
+                del data['action']
```
feanil, line 213: I think a comment here on why you're doing these deletes would be useful.  I think you can just take the explanation from your github comment https://github.com/edx/jenkins-job-dsl/pull/1004#discussion_r437057700

FWIW, the onliner was pretty easy to read for me so I would be fine with that as well along with the explanation of why.

PR #1010, opened 2020-06-15
run terminate cluster from correct directory

estute:


dataeng/jobs/analytics/TerminateCluster.groovy
```diff
@@ -16,7 +16,7 @@ class TerminateCluster {
             steps {
                 virtualenv {
                     nature("shell")
-                    command('EXTRA_VARS="name=$CLUSTER_NAME" make terminate.emr')
+                    command('EXTRA_VARS="name=$CLUSTER_NAME" cd edx-analytics-configuration && make terminate.emr')
```
HassanJaveed84, line 19: Can we use $WORKSPACE here ?

HassanJaveed84: Also, the repo gets checked out in analytics-configuration folder.

estute: good catch

PR #1014, opened 2020-06-17
Added python 3 interpreter for analytics exporter jobs

mraarif:


morenol:
It looks like we need to use this name instead: PYTHON_3.5

Jeremy fixed that in other places yesterday

https://github.com/edx/jenkins-job-dsl/pull/1012

mraarif:
this change needs to wait for python 3 support in `edx-analytics-exporter` as discussed [here](https://github.com/edx/edx-analytics-exporter/pull/48)

PR #1017, opened 2020-06-17
Change Schema Builder to use the PyPI version

bmedx:


dataeng/resources/snowflake-schema-builder.sh
```diff
@@ -3,7 +3,7 @@ set -ex
 
 # Setup
 cd $WORKSPACE/warehouse-transforms
-pip install -r tools/dbt_schema_builder/requirements.txt
+pip install dbt-schema-builder==0.1.2
```
estute, line 6: should we pin dbt-schema-builder in warehouse-transforms instead of in the dsl script?

bmedx: My goal was to separate schema builder from warehouse-transforms entirely and make w-t simply a dbt repository that just happens to be populated quite a bit by schema builder. When I remove schema builder from w-t I was planning up moving / updating that requirements file for people who run dbt locally, but leaving schema builder out of it. 

I'm open to suggestions if you can see a reason why we would want that in there, though.

estute: I see your point. I was thinking perhaps to keep it as a separate `.in` file in the w-t repo, so that it's not a runtime dependency of our dbt code, but a utility. Similar to how we have something like `docs.in` or `quality.in`. I feel like pinning dependencies in this repo can lead to us not updating it.

Could we run into issues in which dbt-schema-builder and w-t are not using the same underlying version of dbt? Not sure if that would cause issues. Having them all in one repo and make upgrade would take care of that.

bmedx: That seems counter-intuitive to me since we won't have any runnable code in w-t. This is the only place that actually runs schema builder, so I could remove the pinned version and just grab the latest release. 

I see the issue about differing versions between w-t and schema builder, we've only avoided them because of the weird way they were entangled before, but since it's only two places where we have the dbt version, and those things should be logically separate, it makes sense to just update the doc about bumping dbt versions to make it clear which two places to update, rather than continue a dependency that we probably shouldn't have had in the first place.

estute: Ok, you have convinced me. Let's just make sure to update the docs for updating dbt.

PR #1024, opened 2020-06-18
Update dbt jobs to pull requirements from new location

bmedx:
Must merge with: https://github.com/edx/warehouse-transforms/pull/741

### DIFF
--- a/dataeng/resources/dbt-docs.sh
+++ b/dataeng/resources/dbt-docs.sh
@@ -5,7 +5,7 @@ set -ex
 cd $WORKSPACE/warehouse-transforms
 
 # To install right version of dbt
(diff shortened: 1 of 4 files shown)


estute:
Should we be using the Makefile target for installing dependencies?

bmedx:
> Should we be using the Makefile target for installing dependencies?

I'm not sure. The Makefile is down in the reporting project so it seems kind of odd, and I'm really not super sure what the scope of that Makefile is intended to be. I read it as "human users of the reporting project", but I'm open to other possibilities.

bmedx:
@pwnage101 do you have thoughts on the makefile question?

PR #1030, opened 2020-06-29
Run Python 3.8 edx-platform tests by default ARCHBOM-1317

jmbowman:
Run the Python 3.8 tests by default in edx-platform now that they're passing.


jmbowman:
Run the Python 3.8 tests by default in edx-platform now that they're passing.

PR #1032, opened 2020-06-29
Doc improvements for sandbox creation

timmc-edx:
- Names can in fact contain hyphens
- Make it clearer how to experiment with private sandbox configuration value changes (make repo names greppable)

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -105,9 +105,10 @@ class CreateSandbox {
                          One reason you might want to override this field is if you are building a sandbox for review or a specific task. \
                          If setting this, you probably want to also set name_tag below. \
                          For example, if you are building a sandbox for pull request 1234 put in 'pr1234' which will setup the sandbox <i>pr1234.sandbox.edx.org</i>.<br /> \
-                         <b>Do not use underscores here.</b> <br /> \
+                         <b>Do not use hyphens or underscores here.</b> <br /> \
                          If you are building a sandbox for yourself, you may leave this blank <b>unless</b> your GitHub \
-                         username contains underscores. <br />")
+                         username contains underscores (which are invalid for domain names in URLs) \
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -105,9 +105,9 @@ class CreateSandbox {
                          One reason you might want to override this field is if you are building a sandbox for review or a specific task. \
                          If setting this, you probably want to also set name_tag below. \
                          For example, if you are building a sandbox for pull request 1234 put in 'pr1234' which will setup the sandbox <i>pr1234.sandbox.edx.org</i>.<br /> \
-                         <b>Do not use hyphens or underscores here.</b> <br /> \
```
adzuci, line 108: Though I know we can use hyphens for some builds I am curious if the variable is sometimes passed into something that expects it to not have hyphens.  That said, I like hyphens so I support this change.

jdmulloy: Are we sure about this? I think it may get used for dns which means you can't use underscores. Do we know why this was documented in the first place and if the issue has been resolved?

timmc-edx: My Github username has a hyphen in it, which is how I know that hyphens aren't a problem here -- I can use the defaults and still get a working sandbox. Underscores are probably still a problem, because URLs can't have underscores in the host component, but I've left that warning in there.

timmc-edx: My sandboxes converge, but is it possible that something optional is breaking that I don't notice.

Looks like this was actually added relatively recently: https://github.com/edx/jenkins-job-dsl/pull/765 I'll ask in there to see if it's something that's still reproducible.

timmc-edx: I don't know how to reproduce the conditions Kyle mentions, so I'm reverting this part but explaining it a little.

PR #1034, opened 2020-06-30
Jwang/de 269

w-joy5015:
Added job for collecting snowflake metrics (DENG-269).

Specifically, this Jenkins job automates the collect_queue_depth.py script (in analytics-tools/snowflake) to run every 5 minutes.

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
+++ b/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
@@ -17,7 +17,7 @@ This job runs every 5 minutes using the Snowflake task automation user.
             logRotator common_log_rotator(allVars)
             parameters secure_scm_parameters(allVars)
             parameters {
-                stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                stringParam('ANALYTICS_TOOLS_URL', allVars.get('TOOLS_REPO_URL'), 'URL for the analytics tools repo.')
                 stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
             }
             environmentVariables {
--- a/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
+++ b/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
@@ -1,0 +1,73 @@
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+
+class SnowflakeQueueDepth {
+
+    public static def job = { dslFactory, allVars ->
+
+        Map SnowflakeQueueDepthConfig = [
+            NAME: "collect-queue-depth-hourly",
+            MANUAL: false,
+            SCRIPT_TO_RUN: "dataeng/resources/snowflake-collect-queue-depth.sh"
+        ]
+        List jobConfigs = [
+            SnowflakeQueueDepthConfig
+        ]
+
+        jobConfigs.each { jobConfig ->
+
+            dslFactory.job(jobConfig['NAME']) {
+
+                logRotator common_log_rotator(allVars)
+                parameters secure_scm_parameters(allVars)
+                parameters {
+                    stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                    stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
+                }
+                environmentVariables {
+                    env('KEY_PATH', allVars.get('KEY_PATH'))
+                    env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
+                    env('USER', allVars.get('USER'))
+                    env('ACCOUNT', allVars.get('ACCOUNT'))
+                }
+                multiscm secure_scm(allVars) << {
+                    git {
+                        remote {
+                            url('$ANALYTICS_TOOLS_URL')
+                            branch('$ANALYTICS_TOOLS_BRANCH')
+                            credentials('1')
+                        }
+                        extensions {
+                            relativeTargetDirectory('analytics-tools')
+                            pruneBranches()
+                            cleanAfterCheckout()
+                        }
+                    }
+                }
+                if (!jobConfig['MANUAL']) {
+                    triggers common_triggers(allVars)
+                }
+                wrappers {
+                    timestamps()
+                }
+                publishers common_publishers(allVars)
+                steps {
+                        virtualenv {
+                        pythonName('PYTHON_3.7')
+                        nature("shell")
+                        systemSitePackages(false)
+                        command(
+                            dslFactory.readFileFromWorkspace(jobConfig['SCRIPT_TO_RUN'])
+                        )
+                    }
+                }
+            }
+        }
+    }
+}
--- a/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
+++ b/dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
@@ -8,11 +8,13 @@ import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
 
 
 class SnowflakeQueueDepth {
-
+"""
+This job runs every 10 minutes using the Snowflake task automation user.
+"""
     public static def job = { dslFactory, allVars ->
 
         Map SnowflakeQueueDepthConfig = [
-            NAME: "collect-queue-depth-hourly",
+            NAME: "collect-queue-depth",
             MANUAL: false,
             SCRIPT_TO_RUN: "dataeng/resources/snowflake-collect-queue-depth.sh"
         ]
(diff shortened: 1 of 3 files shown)


dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,73 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+
+class SnowflakeQueueDepth {
+
+    public static def job = { dslFactory, allVars ->
```
w-joy5015, line 12: Where is allVars coming in from, and how do I define it? This should be where I define my environment variables, right?

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+                            credentials('1')
+                        }
+                        extensions {
+                            relativeTargetDirectory('analytics-tools')
+                            pruneBranches()
+                            cleanAfterCheckout()
+                        }
+                    }
+                }
+                if (!jobConfig['MANUAL']) {
```
pwnage101, line 55: leftover cruft?

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+                    triggers common_triggers(allVars)
```
pwnage101, line 56: as discussed, i'd be down to just have a literal `cron()` defined here.

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+                    timestamps()
+                }
+                publishers common_publishers(allVars)
+                steps {
+                        virtualenv {
+                        pythonName('PYTHON_3.7')
+                        nature("shell")
+                        systemSitePackages(false)
+                        command(
+                            dslFactory.readFileFromWorkspace(jobConfig['SCRIPT_TO_RUN'])
```
pwnage101, line 68: Is this what some of our other DSL jobs do?  I think we can just put the string literal path to the script in here.

dataeng/resources/snowflake-collect-queue-depth.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Setup
+cd $WORKSPACE/analytics-tools/snowflake
+make requirements
+
+python collect_queue_depth.py \
+    --key_path $KEY_PATH \
+    --passphrase_path $PASSPHRASE_PATH \
+    --automation_user $USER \
+    --account $ACCOUNT
```
pwnage101, line 12: I'm noticing all your files don't end with newlines...it only matters to me inasmuch as github adds a red dash on the last line every time.  I know IDEs can be a bit finnicky about this, maybe you can find a setting to automatically add a newline?

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+            MANUAL: false,
+            SCRIPT_TO_RUN: "dataeng/resources/snowflake-collect-queue-depth.sh"
+        ]
+        List jobConfigs = [
+            SnowflakeQueueDepthConfig
+        ]
+
+        jobConfigs.each { jobConfig ->
+
+            dslFactory.job(jobConfig['NAME']) {
```
pwnage101, line 27: I'd like to see the name as a string literal here.

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+    public static def job = { dslFactory, allVars ->
+
+        Map SnowflakeQueueDepthConfig = [
+            NAME: "collect-queue-depth",
+            MANUAL: false,
+            SCRIPT_TO_RUN: "dataeng/resources/snowflake-collect-queue-depth.sh"
+        ]
+        List jobConfigs = [
+            SnowflakeQueueDepthConfig
+        ]
```
pwnage101, line 23: I think this whole section can be deleted provided that you moved the strings into the code below.

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
```
pwnage101, line 4: if you don't use common_triggers anymore, this line can be removed.

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,75 @@
...
+                parameters secure_scm_parameters(allVars)
+                parameters {
+                    stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                    stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
+                }
+                environmentVariables {
+                    env('KEY_PATH', allVars.get('KEY_PATH'))
+                    env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
+                    env('USER', allVars.get('USER'))
+                    env('ACCOUNT', allVars.get('ACCOUNT'))
```
pwnage101, line 39: I'd prefer SNOWFLAKE_USER and SNOWFLAKE_ACCOUNT, there are way too many services flying around for which we have users for me to keep track of.

pwnage101:
Also, we typically change the PR title to something more descriptive, like "Add job for snowflake metrics (DENG-269)"

dataeng/resources/snowflake-collect-queue-depth.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup
+cd $WORKSPACE/analytics-tools/snowflake
+make requirements
+
+python collect_queue_depth.py \
+    --key_path $KEY_PATH \
```
pwnage101, line 9: On the google hangout we were talking about changing this to:
```suggestion
    --key_path $WORKSPACE/analytics-secure/$KEY_PATH \
```

pwnage101: but to be honest I would rather it just be completely literal:
```suggestion
    --key_path $WORKSPACE/analytics-secure/snowflake/rsa_key_snowflake_task_automation_user.p8
```

the path to the secret is not itself a secret.

pwnage101: Likewise, the all of these things are not secret keys and therefore can be moved into the job DSL file:

* PASSPHRASE_PATH
* SNOWFLAKE_USER
* SNOWFLAKE_ACCOUNT


On a related note, ANALYTICS_TOOLS_URL and ANALYTICS_TOOLS_BRANCH definitions could be moved into common.yaml: https://github.com/edx-ops/analytics-secure/blob/master/job-configs/common.yaml

after all those changes, your job config file would almost be completely nullified!

w-joy5015: Ok, I can make that change if it's preferred for readability, but that's the main reason, correct?

pwnage101: readability is the main reason.

dataeng/resources/snowflake-collect-queue-depth.sh
```diff
@@ -6,7 +6,7 @@ cd $WORKSPACE/analytics-tools/snowflake
...
 
 python collect_queue_depth.py \
-    --key_path $KEY_PATH \
-    --passphrase_path $PASSPHRASE_PATH \
-    --automation_user $SNOWFLAKE_USER \
-    --account $SNOWFLAKE_ACCOUNT
+    --key_path $WORKSPACE/analytics-secure/$KEY_PATH \
+    --passphrase_path $WORKSPACE/analytics-secure/$PASSPHRASE_PATH \
+    --automation_user $WORKSPACE/analytics-secure/$SNOWFLAKE_USER \
+    --account $WORKSPACE/analytics-secure/$SNOWFLAKE_ACCOUNT
```
pwnage101, line 12: You'll want to keep these two lines as they were.
```suggestion
    --automation_user $SNOWFLAKE_USER \
    --account $SNOWFLAKE_ACCOUNT
```

w-joy5015: Yeah sorry, I didn't see some of your comments appear until after I pushed that last commit, so I'm going to move some of those things you mentioned above to the job DSL file in my next commit

dataeng/jobs/analytics/SnowflakeQueueDepth.groovy
```diff
@@ -0,0 +1,62 @@
...
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('ANALYTICS_TOOLS_URL', allVars.get('TOOLS_REPO_URL'), 'URL for the analytics tools repo.')
+                stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
+            }
+            environmentVariables {
+                env('KEY_PATH', allVars.get('KEY_PATH'))
+                env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
+                env('SNOWFLAKE_USER', allVars.get('SNOWFLAKE_USER'))
+                env('SNOWFLAKE_ACCOUNT', allVars.get('SNOWFLAKE_ACCOUNT'))
```
pwnage101, line 27: These two parameters are no longer used, so it's effectively dead code.  Instead of deleting the lines, I would actually recommend changing them to 
```suggestion
                env('SNOWFLAKE_USER', 'SNOWFLAKE_TASK_AUTOMATION_USER')
                env('SNOWFLAKE_ACCOUNT', 'edx.us-east-1')
```
and change the bash script to use these environment variables like before.  The reason is that we still want to retain the flexibility of being able to manually override these values while running the job.

PR #1036, opened 2020-07-02
BOM-1865 : Adjust "make upgrade" schedule for Arch-owned repos

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-1865

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -16,10 +16,14 @@ package testeng
 // Cron value for once per week, sometime during: midnight-7:59am ET, Monday-Friday (times are in UTC)
 // See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly
 def cronOffHoursBusinessWeekday = 'H H(4-11) * * H(1-5)'
+// Cron value for once per week, sometime during: 5am-12:59pm Pakistan Time, Monday-Friday (times are in UTC)
+def cronOffHoursBusinessWeekdayLahore = 'H H(0-7) * * H(1-5)'
 
 // Cron value for daily, sometime during: midnight-7:59am ET, Monday-Friday (times are in UTC)
 // See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly
 def cronOffHoursBusinessDaily = 'H H(4-11) * * 1-5'
+// Cron value for daily, sometime during: 5am-12:59pm Pakistan Time, Monday-Friday (times are in UTC)
+def cronOffHoursBusinessDailyLahore = 'H H(0-7) * * 1-5'
 
 Map apiDocTools = [
     org: 'edx',
@@ -37,7 +41,7 @@ Map bokchoy = [
     repoName: 'bok-choy',
     targetBranch: "master",
     pythonVersion: '3.5',
-    cronValue: cronOffHoursBusinessWeekday,
+    cronValue: cronOffHoursBusinessWeekdayLahore,
     githubUserReviewers: [],
     githubTeamReviewers: ['arbi-bom'],
     emails: ['arbi-bom@edx.org'],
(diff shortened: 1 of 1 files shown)


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -222,7 +226,7 @@ Map edxEnterprise = [
     repoName: 'edx-enterprise',
     targetBranch: "master",
     pythonVersion: '3.5',
-    cronValue: cronOffHoursBusinessWeekday,
+    cronValue: cronOffHoursBusinessWeekdayLahore,
```
robrap, line 229: Maybe @jmbowman needs to speak to these owned by other teams. I’m not sure what was agreed upon. Feels like either you should be added to githubTeamReviewers, or you shouldn’t be adjusting the schedule, but this isn’t really my decision.

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -16,10 +16,14 @@ package testeng
 // Cron value for once per week, sometime during: midnight-7:59am ET, Monday-Friday (times are in UTC)
 // See https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly
 def cronOffHoursBusinessWeekday = 'H H(4-11) * * H(1-5)'
+// Cron value for once per week, sometime during: 5am-12:59pm Pakistan Time, Monday-Friday (times are in UTC)
+def cronOffHoursBusinessWeekdayLahore = 'H H(0-7) * * H(1-5)'
```
robrap, line 20: I personally liked when these would appear as new work in the morning. Are most people on arbi-bom not starting until after 1pm your time? If that is your adjusted schedule, this is fine. Otherwise, you can bump that up a couple of hours. I found it nicer than having it appear as an interruption in the middle of the day.
In any case, this schedule should be all about your convenience. So, whatever you think best is fine.

testeng/jobs/upgradePythonRequirements.groovy
mzulqarnain1: Yes, you are right, in case of this repo, reviewer team is `business-enterprise-team` but email were being sent to `arch-bom` previously, so I changed that to arbi-bom, imo I should revert this change, but still @jmbowman can guide better.

testeng/jobs/upgradePythonRequirements.groovy
mzulqarnain1: Yes, you are right, most of us usually start work around 12-1 PM and these hours are agreed upon by the team.

robrap: If so, you could switch to H(0-5) [or H(0-6)] from H(0-7) so you have the PRs waiting when you start. But, again, totally your decision.

mzulqarnain1: Yes, that sounds good, would that be fine, If I reduce the window to 5 hours?

mzulqarnain1: Changed as suggested.

testeng/jobs/upgradePythonRequirements.groovy
jmbowman: I suspect that email was a copy/paste error; @georgebabey , what email address should receive notifications when an edx-enterprise "make upgrade" job fails to get far enough to even create a PR?

mzulqarnain1: I've reverted cronValue for this repo.

mzulqarnain1:
Hey @jmbowman Can I merge this?

testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -84,7 +88,7 @@ Map cookiecutterDjangoApp = [
     repoName: 'cookiecutter-django-app',
     targetBranch: "master",
     pythonVersion: '3.5',
-    cronValue: cronOffHoursBusinessWeekday,
+    cronValue: cronOffHoursBusinessWeekdayLahore,
```
jmbowman, line 91: This repo has been archived since it was moved into edx-cookiecutters, it can be removed from the list.

mzulqarnain1: Np. I've removed this app from the list.

mzulqarnain1: It's already replaced with `edx-cookiecutters`, just had to rebase the PR.

PR #1045, opened 2020-07-11
Disable jobs that load the BigQuery Warehouse

macdiesel:
Disable the loading of the big query warehouse.


HassanJaveed84:
Just realized, we would also want to disable load-json-events-to-s3 job.

PR #1049, opened 2020-07-14
Added timeout to insights jobs.

HassanJaveed84:


HassanJaveed84:
Ready for review again.

brianhw:
Just so I understand, this now adds two things -- one is to mark a build as failed if it times out, and a second is to terminate the cluster if a build fails.   Is there a reason we would want the failed-build termination to only apply when the build timed out, and not all cases where a build is failing?   This presumably means we cannot abort a build and look at a cluster.   This may be a good thing, on average, but just so we know.....

HassanJaveed84:
The post build task to terminate the cluster will only run if the build times out. It looks for text `Build timed out` before running the post build script. You can see it in action here http://jenkins.analytics.edx.org/job/module-engagement-production/665/console at the very end.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -162,6 +162,23 @@ This text may reference other parameters in the task as shell variables, e.g.  $
...
+                likelyStuck()
+                failBuild()
+            }
+        }
+    }
+
+    public static def terminate_cluster_post_build = { allVars ->
+        return {
+            postBuildTask {
+                task('Build timed out', 'cd $WORKSPACE/analytics-configuration && EXTRA_VARS="name=$CLUSTER_NAME" make terminate.emr', true)
```
brianhw, line 177: Oh, so the "Build timed out" is not a doc string, but a pattern that needs to match somewhere in the console log?   I didn't realize that.   I was going to suggest that maybe a comment here describing it would be helpful, but then noticed that there are no comments anywhere.

HassanJaveed84: Added a comment.

jmbowman:
All Travis jobs on this repo have been failing since this was merged, and it looks like they failed on this PR also.  Any chance of reverting until that can be resolved?  It looks like codeNarc is unhappier than usual about at least one of the changes (hard to tell at a glance which one).

jmbowman:
Previously:
```
[ant:codenarc] Summary: TotalFiles=4 FilesWithViolations=4 P1=0 P2=89 P3=62
[ant:codenarc] 
[ant:codenarc] File: org/edx/jenkins/dsl/AnalyticsConstants.groovy
```

After this PR:
```
[ant:codenarc] Summary: TotalFiles=4 FilesWithViolations=4 P1=0 P2=94 P3=64
[ant:codenarc] 
[ant:codenarc] File: org/edx/jenkins/dsl/AnalyticsConstants.groovy
```

It used to proceed after running codeNarc on that file, now it fails the build there.

HassanJaveed84:
@jmbowman reverted.

PR #1052, opened 2020-07-15
Added custom parameter for target branch

UsamaSadiq:
**Issue:** [BOM-1770](https://openedx.atlassian.net/browse/BOM-1770)

### Description
- Add the custom parameter to run the `upgrade job` for the specific branch on a repo.

#### Testing
- Tested the changes by manually running the seed job on Jenkins and verified the parameter option in all the jobs.

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -4,7 +4,7 @@ package testeng
 // Map exampleConfig = [
 //     org: Github organization,
 //     repoName: Github repository,
-//     targetBranch: the default branch of the repository. Ideally, this should be 'main'
+//     defaultBranch: the default branch of the repository. Ideally, this should be 'main'
 //     pythonVersion: The version of python under which to run pip-compile.  If multiple versions are supported by the repo, use the oldest one
 //     cronValue: How often to run the job,
 //     githubUserReviewers: List of Github users that should be tagged on PR's, e.g.: ['user1', 'user2']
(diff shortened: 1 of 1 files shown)


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -28,7 +28,6 @@ def cronOffHoursBusinessDailyLahore = 'H H(0-5) * * 1-5'
 Map apiDocTools = [
     org: 'edx',
     repoName: 'api-doc-tools',
-    targetBranch: "master",
```
jmbowman, line 31: These were all added in https://github.com/edx/jenkins-job-dsl/pull/1025 as part of preparation to support renaming the default branch in many repos to something like "main" instead of "master".  Could we retain the ability to specify this (maybe renamed to "defaultBranch"), but override it with the new parameter if provided?  The parameter should default to empty, since different repos will soon have different default branch names that we want to fall back to.

UsamaSadiq: My understanding was that the automated job will always run for the `master` branch of a repo so it should be a common default value for all the repos which can be renamed later to `main` if required. 
If different repos are going to have different default branches for the job then I'll add a check in the script as you suggested.

jmbowman: dbt-schema-builder has already switched to `main` as the default branch: https://github.com/edx/dbt-schema-builder , and the plan is to switch more one-by-one (although there isn't a schedule for doing so yet).  This came out of the "Juneteenth" mini-hackathon: https://openedx.atlassian.net/wiki/spaces/ENG/pages/1553695423/How+To+Make+Engineering+More+Inclusive , and is related to a similar GitHub effort: https://www.cnet.com/news/microsofts-github-is-removing-coding-terms-like-master-and-slave/ .

I added a `github.default_branch` field to the repo health dashboard to track the status of this across our repos: https://docs.google.com/spreadsheets/d/1VCxNVq-niT-uv5BFmsYPF21r6I2-IQ-GJbidF0zUPBc/edit#gid=921158295

UsamaSadiq:
Tested the change in the seed job and ran a successful upgrade job https://github.com/edx/xqueue/pull/756 on a branch. After verifiying the changes in the created PR, now finally merging the PR.

PR #1060, opened 2020-07-21
Jenkins job for hourly Snowflake credit metric collection script

w-joy5015:
This PR configures a job in Jenkins to run the Snowflake credit metric collection script (in analytics-tools) at the start of every hour. This is part of a larger project to collect health metrics on Snowflake and create a data-visualization dashboard.

This PR also attempts to consolidate repetitious code since it is closely related to the SnowflakeQueueDepth.groovy job. Both these jobs have almost the same bash script and the same variables in analytics-secure.

DENG-268


dataeng/resources/snowflake-collect-metrics.sh
```diff
@@ -5,7 +5,7 @@ set -ex
 cd $WORKSPACE/analytics-tools/snowflake
 make requirements
 
-python collect_queue_depth.py \
+python $METRIC_COLLECTION_SCRIPT \
```
pwnage101, line 8: I'd say that in the industry how we might typically abstract this is by writing only one script (one entrypoint) but allowing different metrics to be selected as part of an argument, like `--metric queue_depth`.  Works either way so I'm not going to block.  I've blocked enough this week!

dataeng/jobs/createJobsNew.groovy
```diff
@@ -31,7 +32,7 @@ def taskMap = [
     EMR_COST_REPORTER_JOB: EmrCostReporterJob,
     EXPIRE_VERTICA_PASSWORD_JOB: ExpireVerticaPasswordJob,
     SNOWFLAKE_EXPIRE_PASSWORDS_JOB: SnowflakeExpirePasswordsJob,
-    SNOWFLAKE_QUEUE_DEPTH_JOB: SnowflakeQueueDepthJob,
+    SNOWFLAKE_COLLECT_METRICS_JOB: SnowflakeQueueDepthJob, SnowflakeCreditUsageJob,
```
pwnage101, line 35: Actually, how do you know this works?  does this map a key to a list?  I would actually expect this to break the syntax of the map.  Even if it does create a list value, see this later line:

```
task.value(this, commonConfigMap + extraVars)
```

value is treated as a callable, but this wouldn't work if value is a list.

w-joy5015: I see, hmm... I was trying to condense the repetitive code by setting up a way to have these (and future) metric collection scripts point to the same file in analytics-secure. I could still try to introduce some if...else logic to the `for (task in taskMap)` loop like: if the data type is a list, iterate through the values in the list.

Or I could just do it the usual way and not rock the boat too much...

dataeng/resources/snowflake-collect-metrics.sh
w-joy5015: Just to clarify, this would mean I have 1 .groovy script for all the metric collection jobs and then each job has its own bash script, right?

dataeng/jobs/createJobsNew.groovy
w-joy5015: Basically this: 
`if (task.value instanceof List) {
        task.value.forEach(this, commonConfigMap + extraVars)
    } else {
        task.value(this, commonConfigMap + extraVars)
    }`

dataeng/resources/snowflake-collect-metrics.sh
estute: Just to clarify after our meeting: if you refactor the analytics-tools code to provide one interface for all snowflake metric gathering, you could just have one bash file in which the only thing that is parameterized would be the argument passed to your script. Something like:
```
python gather_snowflake_metrics.py --metric $METRIC
```

dataeng/jobs/createJobsNew.groovy
```diff
@@ -50,8 +51,11 @@ for (task in taskMap) {
     catch (Exception e) {
         out.println("File $extraVarsFileName does not exist.")
     }
-
-    task.value(this, commonConfigMap + extraVars)
+    if (task.value instanceof List) {
+        task.value.forEach(this, commonConfigMap + extraVars)
+    } else {
+        task.value(this, commonConfigMap + extraVars)
+    }
```
estute, line 58: I think you should continue to follow the pattern here of mapping the key to just one Job object. In our meeting, we discussed refactoring the job dsl to iterate over a config map to generate n-many Jenkins jobs from one DSL file

dataeng/jobs/analytics/SnowflakeMetricCollection.groovy
```diff
@@ -0,0 +1,75 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+
+class SnowflakeQueueDepth {
+
+    public static def job = { dslFactory, allVars ->
+
+        Map SnowflakeWarehouseCreditConfig = [
+            NAME: "collect_credit_metrics"
```
pwnage101, line 14: These job names might be a little vague out of context.  I wouldn't be surprised we have over 100 jobs already---we need the job name to set itself apart a little.  What about:

* snowflake-metrics-collect-cedits
* snowflake-metrics-collect-queue-depth

Just adding "snowflake" might be sufficient.  Also, looking at other job names, it seems we typically use dashes instead of underscores.

dataeng/jobs/analytics/SnowflakeMetricCollection.groovy
```diff
@@ -0,0 +1,75 @@
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+
+class SnowflakeQueueDepth {
```
pwnage101, line 9: class name should be updated to SnowflakeCollectMetrics

dataeng/jobs/createJobsNew.groovy
```diff
@@ -52,7 +52,6 @@ for (task in taskMap) {
     catch (Exception e) {
         out.println("File $extraVarsFileName does not exist.")
     }
-
```
pwnage101, line 55: If changes in this code were reverted, try not to make the diff show that it was touched.

PR #1061, opened 2020-07-24
Add edx-toggles to upgrade jobs

timmc-edx:


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -327,6 +327,18 @@ Map edxSphinxTheme = [
     alwaysNotify: false
 ]
 
+Map edxToggles = [
+    org: 'edx',
+    repoName: 'edx-toggles',
+    targetBranch: "master",
+    pythonVersion: '3.5',
+    cronValue: cronOffHoursBusinessWeekday,
```
robrap, line 335: Missing Lahore at end. Can you check other options again from something configured correctly?

timmc-edx: Ah, good catch! I copied this from edxRepoHealth, which specifies this same combination -- but probably should not!

timmc-edx: `edxRepoHealth` and `pytestRepoHealth` both have this mismatch. Additionally, `notifier` specifies `arch-bom` instead of `arbi-bom`. Should they all be `arbi-bom` and Lahore hours?

timmc-edx: I'll go ahead and fix those first two.

PR #1075, opened 2020-07-29
Reconfigure sqoop loads to support stitch validation

pwnage101:
* Pass the Sqoop start time all the way down to the validation job.
* Reduce frequency of loading into Snowflake.
* Cut down the number of tables being loaded into Snowflake.
* Refactor job configs so that they are closer to the code.

DENG-359

---

analytics-secure changes: https://github.com/edx-ops/analytics-secure/pull/424


dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
+                    // validation.
+                    conditionalSteps {
+                        condition { fileExists('${WORKSPACE}/build_skipped') }
+                        steps { setBuildResult('ABORTED') }
+                    }
+                }
+                publishers common_publishers(allVars)
+                publishers {
+                    downstreamParameterized {
+                        trigger("validate-stitch-against-sqoop" )
```
pwnage101, line 79: This needs to change once https://openedx.atlassian.net/browse/DENG-360 is complete.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
+                    stringParam('WAREHOUSE', 'LOADING_SQOOP', 'The Snowflake warehouse to use.')
+                    stringParam('ROLE', 'PIPELINE_ETL_LOADER_ROLE', 'The Snowflake role to assume.')
+                    stringParam('SNOWFLAKE_DATABASE', 'PROD', 'The destination database in Snowflake.')
+                    stringParam('SCHEMA', jobConfig.get('SCHEMA'), 'The destination raw schema in Snowflake.')
+                    stringParam('SCRATCH_SCHEMA', jobConfig.get('SCRATCH_SCHEMA'), 'Scratch schema name - temporary loading location in Snowflake.')
+                    stringParam('DATABASE', jobConfig.get('DATABASE'), 'Name of MySQL database to copy from.')
+                    stringParam('INCLUDE', jobConfig.get('INCLUDE', ''), 'List of regular expressions matching table names that should be copied.')
+                    stringParam('EXCLUDE', jobConfig.get('EXCLUDE', ''), 'List of regular expressions matching table names that should not be copied.')
+                    stringParam('SQOOP_START_TIME', '', 'The start time of the upstream sqoop load (in ISO timestamp format).')
```
pwnage101, line 58: Note that I expanded several of these defaults into string literals.

dataeng/resources/snowflake-replica-import.sh
```diff
@@ -2,18 +2,34 @@
...
+     --role $ROLE \
+     --sf-database $SNOWFLAKE_DATABASE \
+     --schema $SCHEMA \
+     --scratch-schema $SCRATCH_SCHEMA \
+     --run-id $BUILD_ID \
+     --database $DATABASE \
+     --overwrite \
+     ${INCLUDE} \
+     ${EXCLUDE} \
+     ${EXTRA_ARGS}
```
pwnage101, line 32: None of this changed---I just indented the whole block.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
+                SCRATCH_SCHEMA: 'CREDENTIALS_SCRATCH_LOADING',
+                DATABASE: 'credentials',
+                INCLUDE: '--include ["core_user", "django_session", "social_auth_usersocialauth"]',
+            ],
+            LMS: [
+                CLUSTER_NAME: 'ImportLMSReadReplicaSnowflakeFromS3',
+                SCHEMA: 'LMS_RAW',
+                SCRATCH_SCHEMA: 'LMS_SCRATCH_LOADING',
+                DATABASE: 'wwc',
+                INCLUDE: '--include ["assessment_assessmentd", "assessment_assessmentfeedbackd", "assessment_assessmentpartd", "auth_userprofiled", "certificates_generatedcertificated", "course_goals_coursegoald", "course_overviews_courseoverviewtabd", "edxval_videoimaged", "enterprise_enterprisecustomerd", "enterprise_historicalenterprisecustomerd", "entitlements_courseentitlementd", "problem_builder_answerd", "proctoring_proctoredexamstudentattemptd", "social_auth_usersocialauthd", "student_courseenrollmentd", "student_loginfailuresd", "student_sociallinkd", "submissions_scoresummaryd", "user_api_usercoursetagd", "wiki_articled", "wiki_articlerevisiond", "workflow_assessmentworkflowd"]',
```
brianhw, line 24: Curious if there is a reason why these table names have a "d" at the end.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
-                    stringParam('SCHEMA', db_config.get('SCHEMA', allVars.get('SCHEMA')), 'Schema')
-                    stringParam('SCRATCH_SCHEMA', db_config.get('SCRATCH_SCHEMA', allVars.get('SCRATCH_SCHEMA')), 'Scratch schema name - temporary loading location.')
-                    stringParam('DATABASE', db_config.get('DATABASE'), 'Name of MySQL database to copy from.')
-                    stringParam('INCLUDE', db_config.get('INCLUDE'), 'List of regular expressions matching table names that should be copied.')
-                    stringParam('EXCLUDE', db_config.get('EXCLUDE'), 'List of regular expressions matching table names that should not be copied.')
+                    stringParam('RUN_DATE', 'today', 'Run date for the job. A string that can be parsed by the GNU coreutils "date" utility.')
+                    stringParam('OVERWRITE', '--overwrite', 'Set to: --overwrite if you want to enable overwrite.')
+                    stringParam('SNOWFLAKE_CREDENTIALS', allVars.get('SNOWFLAKE_CREDENTIALS'), 'The path to the Snowflake credentials file.')
```
brianhw, line 49: If you're moving everything else in here, is there a reason not to move SNOWFLAKE_CREDENTIALS as well?   Isn't it  just an S3 path?

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
-                publishers common_publishers(allVars)
                 steps {
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/snowflake-replica-import.sh'))
                 }
+                postBuildSteps {
+                    // If we skipped today's run, set the current build to ABORTED so that we don't trigger stitch
+                    // validation.
+                    conditionalSteps {
+                        condition { fileExists('${WORKSPACE}/build_skipped') }
+                        steps { setBuildResult('ABORTED') }
```
brianhw, line 73: Is there another result that you could use?  Like "WARNING" or something?   Thinking of what we get on exporter runs.

dataeng/resources/snowflake-replica-import.sh
```diff
@@ -2,18 +2,34 @@
...
- ${EXCLUDE} \
- ${EXTRA_ARGS}
+# Only invoke this script on Mondays, otherwise skip it.  Since this job is
+# triggered by an upstream for which we can't change the cron schedule (due to
+# vertica not yet being deprecated), we must skip it manually in this shell
+# script.
+
+DAY_OF_WEEK=$(date +%u)
+MONDAY=1
+if [[ ${DAY_OF_WEEK} -eq ${MONDAY} ]]; then
```
brianhw, line 12: I would suggest providing some flag that can override this logic, so that if we need to run this manually on a different day, we can do so.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,91 @@
...
+                    stringParam('SNOWFLAKE_DATABASE', 'PROD', 'The destination database in Snowflake.')
+                    stringParam('SCHEMA', jobConfig.get('SCHEMA'), 'The destination raw schema in Snowflake.')
+                    stringParam('SCRATCH_SCHEMA', jobConfig.get('SCRATCH_SCHEMA'), 'Scratch schema name - temporary loading location in Snowflake.')
+                    stringParam('DATABASE', jobConfig.get('DATABASE'), 'Name of MySQL database to copy from.')
+                    stringParam('INCLUDE', jobConfig.get('INCLUDE', ''), 'List of regular expressions matching table names that should be copied.')
+                    stringParam('EXCLUDE', jobConfig.get('EXCLUDE', ''), 'List of regular expressions matching table names that should not be copied.')
+                    stringParam('SQOOP_START_TIME', '', 'The start time of the upstream sqoop load (in ISO timestamp format).')
+                }
+                environmentVariables {
+                    env('APP_NAME', appName)
```
doctoryes, line 61: I'm still confused about where `APP_NAME` gets set?

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
pwnage101: hah!  must have been a typo in my macro.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
pwnage101: It is just an S3 path, But I just wasn't sure if it was too-much-information for this open source repo.

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
pwnage101: Here are the options: 'SUCCESS', 'UNSTABLE', 'FAILURE', 'ABORTED' or 'CYCLE'.

ABORED seemed like the most appropriate because that exactly what's happening: the build script is aborting the build.  Do you think it should be UNSTABLE?

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
pwnage101: Here's we're setting the environment variable APP_NAME to whatever the value of the `appName` variable is.  Higher up we set the `appName` variable on line 42.

dataeng/resources/snowflake-replica-import.sh
```diff
@@ -2,18 +2,34 @@
...
- ${EXCLUDE} \
- ${EXTRA_ARGS}
+# Only invoke this script on Mondays, or when forced to run via the FORCE
+# parameter, otherwise skip it.  Since this job is triggered by an upstream for
+# which we can't change the cron schedule (due to vertica not yet being
+# deprecated), we must skip it manually in this shell script.
+
+DAY_OF_WEEK=$(date +%u)
+MONDAY=1
+if [[ ${DAY_OF_WEEK} -eq ${MONDAY} -o ${FORCE} = "true" ]]; then
```
HassanJaveed84, line 12: Non-blocker: Can we use conditional build step instead of putting this in the script ?
https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.step.MultiJobStepContext.conditionalSteps

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
brianhw: I guess we can go ahead with ABORTED here.  I was thinking of UNSTABLE, yes, but either is going to be ugly.  It's a pity that we have to use the build result to communicate this, but I guess there are only a limited number of environment variables that one can check in a condition in a downstream trigger.

dataeng/resources/snowflake-replica-import.sh
brianhw, line 12: thanks for adding the force.

pwnage101: I briefly evaluated that option, but the only condition function they provided was `time()` which was not sufficient:

```
// Only runs the build steps during a certain period of the day.
time(int earliestHours, int earliestMinutes, int latestHours, int latestMinutes, boolean useBuildTime) 
```

dataeng/jobs/analytics/SnowflakeReplicaImportFromS3.groovy
```diff
@@ -1,37 +1,92 @@
...
+                    // If we skipped today's run, set the current build to ABORTED so that we don't trigger stitch
+                    // validation.
+                    conditionalSteps {
+                        condition { fileExists('${WORKSPACE}/build_skipped') }
+                        steps { setBuildResult('ABORTED') }
+                    }
+                }
+                publishers common_publishers(allVars)
+                publishers {
+                    downstreamParameterized {
```
brianhw, line 79: Here too:  wondering if we can make use of https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.flexiblePublish to wrap this call to downstreamParametrized, and then we can just use the same condition above `condition { fileExists('${WORKSPACE}/build_skipped') }` here instead of in a post-build step.   Then we don't have to mess with the status.

pwnage101: This is essentially a solution to avoid the "aborted" state, which I essentially see as "the job aborted itself" so I'm not completely opposed to it.  For example, it can help us identify in the UI which builds in a list of builds actually did anything.  I'm hesitant to implement this on the basis of that, but you're right that we could consolidate the the conditional logic here in the publishers section.

PR #1084, opened 2020-08-05
Only trigger email notifications when $NOTIFY param is non-empty.

HassanJaveed84:


pwnage101:
I have used the NOTIFY parameter manually in the past to do a test run of a job without alerting the oncall, but I'm sure we can add that on a per-job basis.  Nevertheless, this PR does not remove that parameter so LGTM.

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -164,7 +164,7 @@ This text may reference other parameters in the task as shell variables, e.g.  $
 
     public static def common_publishers = { allVars ->
         return {
-            mailer('$NOTIFY')
+            mailer(allVars.get('NOTIFY','$PAGER_NOTIFY'))
```
HassanJaveed84, line 167: @pwnage101 @brianhw this change ignores the NOTIFY parameter altogether, note the lack of `$` with NOTIFY. 
We could still do test runs of jobs by overriding the RECIPIENT section of the email notification post build action, not ideal. 
Open to suggestions

HassanJaveed84: I don't think we can check if the `NOTIFY` parameter is defined in this context.

brianhw: So this isn't quite doing what we want.  Another way to go would be to make sure that $NOTIFY is always set, and that it's default is set to $PAGER_NOTIFY.

HassanJaveed84: What do we want ? :) 
We have a default, $PAGER_NOTIFY, and we can override by defining `NOTIFY` in the extra vars files. 

What is missing ?

brianhw: Also, do we (not) have cases where the NOTIFY value is not in allVars, but rather in a more environment-specific one?   I'm thinking of the microbachelor DBT builds in particular.

HassanJaveed84: Okay, just saw them. I think you're referring to the ones here: https://github.com/edx-ops/analytics-secure/blob/master/job-configs/WAREHOUSE_TRANSFORMS_JOB_EXTRA_VARS.yaml

We could have it like
```
mailer(env.get('NOTIFY', allVars.get('NOTIFY', '$PAGER_NOTIFY')))
```

estute: as @brianhw mentioned in chat, we could use the flexible publish plugin, but this should work just fine. We already do it in some other jobs

estute:
ugh, that may be right. ok, so you could add this to the `common_publishers` function:
```
sendMailTo = System.getenv("NOTIFY") ?: allVars.get('PAGER_NOTIFY')
```

here is an example of how that works
```
groovy:000> pager_var = "PAGE"
===> PAGE
groovy:000> notify_var = System.getenv("NOTIFY")
===> null
groovy:000> result = notify_var ?: pager_var
===> PAGE
groovy:000> result
===> PAGE

```

HassanJaveed84:
@brianhw that's what I suggested in inline comments. 
```
mailer(env.get('NOTIFY', allVars.get('NOTIFY', '$PAGER_NOTIFY')))
```
This should work but requires a few other changes.

HassanJaveed84:
@estute I tried using what you suggested. But I'm not sure if you can access it in the publisher context.
```
No such property: system for class: javaposse.jobdsl.dsl.helpers.publisher.PublisherContext

File: src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy (hunk 1 above)
Comment (brianhw) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:167]: So `mailer(env.get('NOTIFY', allVars.get('NOTIFY', '$PAGER_NOTIFY')))` would be passing two arguments to mailer().   I'm not clear what gets into the $NOTIFY, and what is currently being loaded into vars.get('NOTIFY').   And if nothing is in allvars.get('NOTIFY'), then this would be adding $PAGER_NOTIFY, regardless of what is in $NOTIFY?

Comment (HassanJaveed84) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:167]: Taking this to slack.

File: src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy [hunk 2]
```diff
@@ -164,7 +164,18 @@ This text may reference other parameters in the task as shell variables, e.g.  $
 
     public static def common_publishers = { allVars ->
         return {
-            mailer('$NOTIFY')
+            flexiblePublish {
+                conditionalAction {
+                    condition {
+                        not {
+                            stringsMatch('$NOTIFY','', false)
```
pwnage101, line 171: Can you add a comment explaining what the `false` does?

pwnage101:
So just to be clear, if there's no NOTIFY parameter defined, `$NOTIFY` will evaluate to empty string?

HassanJaveed84:
No, in that case, the job will throw an error during condition evaluation and mark the build as failure. I don't see a way around it.
```
Exception caught evaluating condition: [org.jenkinsci.plugins.tokenmacro.MacroEvaluationException: Unrecognized macro 'NOTIFY' in '$NOTIFY'], action = [Fail the build]

pwnage101:
Oh, okay any time common_publishers is used, it will require a NOTIFY parameter to be defined (or just use common_parameters)?  Can you comment that common_publishers depends on common_parameters?

brianhw:
I don't think we want it to be dependent on common_parameters.   They are common to a lot of EMR jobs, but there's lots of other times we don't want all those parameters.   So we do require that NOTIFY be defined as a stringParam, but it can be explicitly defined.

HassanJaveed84:
Yes, that is how it should work. I've tested these conditions manually.

PR #1090, opened 2020-08-06
Add prospectus checkbox and versionbox to the list of options for thi…

reillz10:
…ngs to build in the sandbox

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -203,6 +203,9 @@ class CreateSandbox {
                              "video_pipeline and video_encode_worker must be selected for video pipeline to work")
                 stringParam("video_encode_worker_version","master","")
 
+                booleanParam("prospectus",false,"Enable Prospectus, the edx catalog")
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -203,6 +203,9 @@ class CreateSandbox {
                              "video_pipeline and video_encode_worker must be selected for video pipeline to work")
                 stringParam("video_encode_worker_version","master","")
 
+                booleanParam("prospectus",false,"Enable Prospectus, the edx cataolog")
```
srwang, line 206: catalog

devops/jobs/CreateSandbox.groovy
```diff
@@ -203,6 +203,9 @@ class CreateSandbox {
...
+                stringParam("prospectus_version","master","")
```
srwang, line 207: out of curiosity what is that third argument?

reillz10: a description of what the param is for. Other services didn't have one so 🤷

PR #1095, opened 2020-08-10
Fix two issues with replica jobs

pwnage101:
1. In one job the downstream properties file gets clobbered on accident,
   so avoid that by appending instead of truncating.

2. Regarding the --exclude parameter, somewhere along the
   groovy->jenkins->bash->luigi code path we are apparently missing
   backslash escape charaters for double quotes.  This commit adds them
   back.

DENG-359

Testing
---

The old shell argument that worked:
```bash
--exclude '[\"student_passwordhistory\",\".*courseware_studentmodule.*\",\"oauth.*\",\"shoppingcart_orderitem\",\"certificates_certificateinvalidation\",\"certificates_certificategenerationhistory\",\"enterprise_enterprisecustomerreportingconfiguration\",\"social_auth_partial\"]'
```

The new shell argument that is throwing JSON parse errors:
```bash
--exclude '["student_passwordhistory",".*courseware_studentmodule.*","oauth.*","shoppingcart_orderitem","certificates_certificateinvalidation","certificates_certificategenerationhistory","enterprise_enterprisecustomerreportingconfiguration","social_auth_partial"]'
```

It appears that the string literal that we need to pass into Luigi params needs the backslashes.  Using the jenkins /script endpoint, I was able to test how groovy evaluates the string literal:

```groovy
// Baseline test, the current behavior:
String test = '[\"student_passwordhistory\"]'
Result: ["student_passwordhistory"]

// Test with only ONE backslash:
String test = '[\"student_passwordhistory\"]'
Result: ["student_passwordhistory"]

// Test with two backslashes results in the desired behavior of inserting a
// literal backslash in the string:
String test = '[\\"student_passwordhistory\\"]'
Result: [\"student_passwordhistory\"]
```

In conclusion, my fix is to double-backslash all the double quotes in the --exclude statement.  Here's what the parameter looks like after I seeded the job:

image

### DIFF
--- a/dataeng/resources/read-replica-export.sh
+++ b/dataeng/resources/read-replica-export.sh
@@ -1,19 +1,22 @@
 #!/usr/bin/env bash
 
+set -ex
+
 env
 
-# Truncate downstream properties file:
-printf '' > "${WORKSPACE}/downstream.properties"
+# Cleanup downstream properties file first:
+DOWNSTREAM_PROPERTIES_FILE="${WORKSPACE}/downstream.properties"
+rm "${DOWNSTREAM_PROPERTIES_FILE}" || true
 
 # Keep track of the start time to feed into downstream validation scripts.
-echo "SQOOP_START_TIME=$(date --utc --iso=minutes)" >> "${WORKSPACE}/downstream.properties"
+echo "SQOOP_START_TIME=$(date --utc --iso=minutes)" >> "${DOWNSTREAM_PROPERTIES_FILE}"
 
 # Interpolate the RUN_DATE now so that the downstream job is guaranteed to use
 # the same exact date as this job. Otherwise, if this job runs over a date
 # boundary, the downstream job would re-interpolate the value of 'yesterday' on
 # a different date.
 INTERPOLATED_RUN_DATE="$(date +%Y-%m-%d -d "$TO_DATE")"
-echo "RUN_DATE=${INTERPOLATED_RUN_DATE}" > "${WORKSPACE}/downstream.properties"
+echo "RUN_DATE=${INTERPOLATED_RUN_DATE}" >> "${DOWNSTREAM_PROPERTIES_FILE}"
 
 ${WORKSPACE}/analytics-configuration/automation/run-automated-task.sh \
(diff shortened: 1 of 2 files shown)


dataeng/resources/read-replica-export.sh
```diff
@@ -1,19 +1,22 @@
...
-echo "SQOOP_START_TIME=$(date --utc --iso=minutes)" >> "${WORKSPACE}/downstream.properties"
+echo "SQOOP_START_TIME=$(date --utc --iso=minutes)" >> "${DOWNSTREAM_PROPERTIES_FILE}"
 
 # Interpolate the RUN_DATE now so that the downstream job is guaranteed to use
 # the same exact date as this job. Otherwise, if this job runs over a date
 # boundary, the downstream job would re-interpolate the value of 'yesterday' on
 # a different date.
 INTERPOLATED_RUN_DATE="$(date +%Y-%m-%d -d "$TO_DATE")"
-echo "RUN_DATE=${INTERPOLATED_RUN_DATE}" > "${WORKSPACE}/downstream.properties"
+echo "RUN_DATE=${INTERPOLATED_RUN_DATE}" >> "${DOWNSTREAM_PROPERTIES_FILE}"
```
pwnage101, line 19: Only this was the bugfix; other changes to this file were opportunistic.

brianhw:
Oh, and just for the record, Luigi does take standard JSON, not a weird dialect.  The hard part is that we have a command line in between, and the backslashes are needed to get the argument through the command-line processing.   Still our code-base, but it's not a particularly obscure issue.

pwnage101:
> Oh, and just for the record, Luigi does take standard JSON, not a weird dialect. The hard part is that we have a command line in between, and the backslashes are needed to get the argument through the command-line processing. Still our code-base, but it's not a particularly obscure issue.

Thanks, I'll update the commit message.

pwnage101:
> For the other issue, I don't understand the problem with truncation not working, but a hard delete seems fine.

That's just so the code is consistent with the other place where I do the exact same thing: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/resources/snowflake-replica-import.sh#L10

PR #1098, opened 2020-08-12
Fix issues with excluded models being run

bmedx:
This ticket intended for the given model tag to not be run or tested:
https://openedx.atlassian.net/servicedesk/customer/portal/16/DESUPPORT-414

We only excluded it from tests:
https://github.com/edx/jenkins-job-dsl/pull/1067/files

This PR fixes that to also exclude from runs.

### DIFF
--- a/dataeng/resources/warehouse-transforms.sh
+++ b/dataeng/resources/warehouse-transforms.sh
@@ -24,31 +24,31 @@ then
     dbt test --models source:* --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
 fi
 
+# By default, exclude no models from testing.
+exclude_models=""
+exclude_param=""
+
+if [ "$MODEL_TAG" = 'daily' ]
+then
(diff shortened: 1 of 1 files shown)


dataeng/resources/warehouse-transforms.sh
```diff
@@ -24,33 +24,33 @@ then
...
-    then
-        # For 'daily' models, exclude these two models from finrep that inexplicably are tested when testing those models.
-        exclude_models="$exclude_models finrep_map_organization_course_courserun finrep_royalty_order_dimension tag:daily-exclude"
-    fi
-
-    if [ "$exclude_models" != "" ]
-    then
-        # If models were excluded, add the leading exclude parameter.
-        exclude_param="--exclude $exclude_models"
-    fi
```
brianhw, line 52: I think this needs to be restored, so that the TEST_SOURCES_FIRST exclusion above still gets into the exclude_param as well as the exclude_models.   It could of course be restored in the same if statement (i.e. on line 52).    But the code below only uses exclude_param.

bmedx: I think this is fixed now.

PR #1100, opened 2020-08-12
add prospectus to jenkins job variable list

reillz10:


reillz10:
I realized that I don't need to do this.

PR #1108, opened 2020-08-19
Updated the cleanup job to enable running scripts against all edx repos

mraarif:
Relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1974).

### DIFF
--- a/testeng/resources/cleanup-python-code-create-pr.sh
+++ b/testeng/resources/cleanup-python-code-create-pr.sh
@@ -1,34 +1,62 @@
 #!/bin/bash
 set -e
 
-rm -rf code_cleanup_venv
-virtualenv --python=python$PYTHON_VERSION code_cleanup_venv -q
-source code_cleanup_venv/bin/activate
+IFS=',' read -ra SCRIPTSTORUN <<<"$SCRIPTS"
 
-echo "Upgrading pip..."
-pip install pip==20.0.2
+PACKAGESTOINSTALL=$(echo $PACKAGES | tr , " ")
 
-echo "Getting current sha..."
-cd $REPO_NAME
-export CURRENT_SHA=$(git rev-parse HEAD)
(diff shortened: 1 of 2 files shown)


testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -1,34 +1,49 @@
...
+  rm -rf "${repo}"-code_cleanup_venv
+  virtualenv --python=python"$PYTHON_VERSION" "${repo}"-code_cleanup_venv -q
+  source "${repo}"-code_cleanup_venv/bin/activate
+
+  echo "Upgrading pip..."
+  if pip install pip==20.0.2; then true; else RESULT=$? && continue; fi
+
+  echo "Getting current sha..."
+  cd "${repo}"
+  export CURRENT_SHA=$(git rev-parse HEAD)
```
jmbowman, line 27: I learned from the recent repo health dashboard trouble that this can fail if the repo is empty (such as a new one with no initial commit yet), since `HEAD` doesn't exist yet.

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -1,34 +1,49 @@
...
+for repo in "${REPOS[@]}"; do
 
-pip install $PACKAGESTOINSTALL
+  rm -rf "${repo}"
 
-IFS=',' read -ra SCRIPTSTORUN <<< "$SCRIPTS"
-for element in "${SCRIPTSTORUN[@]}"; do
-  `echo "$element"`
-done
+  if git clone https://github.com/edx/"${repo}".git; then true; else RESULT=$? && continue; fi
```
jmbowman, line 16: Could we also build a list of which repos failed, and give it at the end?  For a nontrivial number of repos, it can be pretty hard to inspect the log and figured out how many and which repos failed.

PR #1119, opened 2020-09-08
run de update users job in the right directory

estute:


brianhw (approved):
If it works, LGTM.  👍

PR #1121, opened 2020-09-10
Add Program console version parameter so we can build Program Console on Sandbox

schenedx:
This is the change to set the $program_console_version into the ansible environments.
This way, we can build program_console on sandbox at the same time as program_manager.
The plan is to remove program_manager items after this is merged.

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -189,6 +189,8 @@ class CreateSandbox {
                             "Key for Organization to be created in Registrar. Must match key in Discovery catalog. Ignore this setting if Registrar is disabled.")
                 stringParam("program_manager_version","master",
                             "The repository version of the frontend-app-program-manager")
+                stringParam("program_console_version","master",
+                            "The repository version of the frontend-app-program-console")
 
                 booleanParam("learner_portal",false,"Learner Portal")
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -189,6 +189,8 @@ class CreateSandbox {
                             "Key for Organization to be created in Registrar. Must match key in Discovery catalog. Ignore this setting if Registrar is disabled.")
                 stringParam("program_manager_version","master",
                             "The repository version of the frontend-app-program-manager")
+                stringParam("program_console_version","master",
+                            "The repository version of the frontend-app-program-console")
```
matthugs, line 193: can we also add the boolean option of whether or not to build this mfe at all? a la https://github.com/edx/configuration/pull/6001/files#diff-32b2548c4c6e77f9c5a605464f58b6a4R273

matthugs (approved):

devops/jobs/CreateSandbox.groovy
schenedx: It's kinda outside of the scope of this PR.
Also, I think we should build the program-console MFE as soon as we decide to have registrar.
It doesn't take more disk space and it provides necessary UI for registrar APIs

PR #1127, opened 2020-09-15
Point snowflake collections job to the loading warehouse

macdiesel:
I also updated the cron to run every 10 minutes instead of 5.

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
+++ b/dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
@@ -16,7 +16,7 @@ class SnowflakeCollectMetrics {
         ]
         Map SnowflakeQueueDepthConfig = [
             NAME: 'snowflake-collect-queue-depth',
-            CRON: '*/5 * * * *'
+            CRON: '*/10 * * * *'
         ]
         List jobConfigs = [
             SnowflakeWarehouseCreditConfig,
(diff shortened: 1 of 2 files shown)


macdiesel:
Companion PR: https://github.com/edx/analytics-tools/pull/65

dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
```diff
@@ -37,6 +37,7 @@ class SnowflakeCollectMetrics {
                 environmentVariables {
                     env('SNOWFLAKE_USER', 'SNOWFLAKE_TASK_AUTOMATION_USER')
                     env('SNOWFLAKE_ACCOUNT', 'edx.us-east-1')
+                    env('SNOWFLAKE_WAREHOUSE', 'LOADING')
```
estute, line 40: I feel like this will make the job of tracking cost/usage confusing. Should we create or use another warehouse here?

estute (commented):
see inline

dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
macdiesel: At some point yes, I will backlog a ticket to do so. For now I don't want to keep spinning up more warehouses that aren't doing much. For now we consolidate and cut costs. Once we have more things happening for monitoring we'll add a monitoring specific warehouse.

PR #1133, opened 2020-09-18
Update the email optin job to use the fixed mgmt command

pwnage101:
I just merged this edx-platform PR:

https://github.com/edx/edx-platform/pull/25011

which updates the email_opt_in_list.py management command to port the
changes in our analytics fork back upstream.  Not taking action will
result in a change in behavior of the analytics-email-optin job, which
is what we're trying to avoid.

I would normally use release tags created by the edxapp gocd pipeline,
but the pipeline is borked so I used a hash instead.

Testing
---

Successful. http://jenkins.analytics.edx.org/job/analytics-email-optin-worker/9619/console


brianhw (approved):
👍

PR #1138, opened 2020-10-02
Bump analytics-exporter-worker concurrency by 6

pwnage101:


brianhw (approved):
I thought the suggestion from Fred was to bump it up directly to 12 (as noted in the SRE ticket).   But one's better than nothing....

PR #1149, opened 2020-10-19
Correct twice monthly schedule of jenkins jobs

matthugs:
inspired by https://stackoverflow.com/a/33788028


coryleeio (approved):

matthugs:
fyi @tuchfarber since this will fix the schedule to do what it says on the tin for a repo they own

PR #1150, opened 2020-10-21
1421: set up Prospectus bihourly sandbox build

srwang:


rgraber (commented):
We can probably save some time if we turn off the things we don't need (forums, test data, etc.)

rgraber (approved):
LGTM

devops/jobs/CreateSandboxCI.groovy
```diff
@@ -21,7 +21,7 @@ import static org.edx.jenkins.dsl.DevopsConstants.common_read_permissions
 class CreateSandboxCI {
   public static def jobs = { dslFactory, extraVars ->
 
-    ["Hourly", "Daily", "MastersWeekly"].each { type ->
+    ["Hourly", "Daily", "MastersWeekly", "ProspectusBiHourly"].each { type ->
```
matthugs, line 24: I know it deviates from an established pattern, but I bend over backwards whenever possible to avoid words like "bimonthly" because they induce the moment of panic of "do I actually know whether that's half as often or twice as often"? I can't be the only one. Can we use `ProspectusEveryTwoHours` instead?

devops/jobs/CreateSandboxCI.groovy
```diff
@@ -106,6 +106,20 @@ class CreateSandboxCI {
                   booleanParam("journals",false)
                   booleanParam("video_pipeline",false)
                   booleanParam("video_encode_worker",false)
+                } else if (type == 'ProspectusBiHourly') {
+                  predefinedProp('dns_name','prospectus-bihourly')
+                  predefinedProp('name_tag','prospectus-bihourly')
+
+                  predefinedProp('server_type', 'full_edx_installation_from_scratch')
+                  booleanParam('reconfigure',true)
+                  booleanParam('recreate', true)
```
matthugs, line 115: the first and last here are the default. The middle won't matter if you take the default for the first.

Defaults are here: https://github.com/edx/jenkins-job-dsl/blob/029b9c6a920f1971be69b4f480c6da8de547a185/devops/jobs/CreateSandbox.groovy
To see what reconfigure does see: https://github.com/edx/configuration/blob/25850cf1de16066da714551c302fb00c8d418b24/util/jenkins/ansible-provision.sh#L489 + like 10 or so lines. (The setting has been kind of redundant for a while, since it's basically always superseded by server_type.)

devops/jobs/CreateSandboxCI.groovy
```diff
@@ -106,6 +106,20 @@ class CreateSandboxCI {
...
+                  predefinedProp('edxapp', false)
+                  predefinedProp('testcourses', false)
+                  predefinedProp('performance_course', false)
+                  predefinedProp('demo_test_course', false)
+                  predefinedProp('edx_demo_course', false)
+                  predefinedProp('forum', false)
```
matthugs, line 121: do you wanna' turn off ecommerce (& its worker) too?

matthugs (approved):
well that was easier than expected, wasn't it?

PR #1151, opened 2020-10-21
Add ability to enable/disable Opsgenie heartbeats for jobs.

doctoryes:
https://openedx.atlassian.net/browse/DENG-547

Uses the values defined here: https://github.com/edx-ops/analytics-secure/pull/463

### DIFF
--- a/dataeng/jobs/analytics/AnswerDistribution.groovy
+++ b/dataeng/jobs/analytics/AnswerDistribution.groovy
@@ -24,11 +24,10 @@ class AnswerDistribution {
                 wrappers common_wrappers(allVars)
                 publishers common_publishers(allVars)
                 steps {
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-enable-heartbeat.sh'))
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/answer-distribution.sh'))
-                    if (env_config.get('OPSGENIE_HEARTBEAT_NAME') && allVars.get('OPSGENIE_HEARTBEAT_KEY')){
-                        shell("curl -X GET 'https://api.opsgenie.com/v2/heartbeats/" + env_config.get('OPSGENIE_HEARTBEAT_NAME') + "/ping' -H 'Authorization: GenieKey " + allVars.get('OPSGENIE_HEARTBEAT_KEY') + "'")
-                    }
-               }
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-disable-heartbeat.sh'))
+                }
             }
         }
     }
--- a/dataeng/resources/opsgenie-enable-heartbeat.sh
+++ b/dataeng/resources/opsgenie-enable-heartbeat.sh
@@ -1,0 +1,31 @@
+# Enables an Opsgenie Heartbeat, while also setting the expected duration.
+#
+# Assumes that the heartbeat is already created in the Opsgenie settings here:
+# https://edx.app.opsgenie.com/settings/heartbeat
+#
+# Required environment variables:
+# OPSGENIE_HEARTBEAT_NAME: Name of the already-created Opsgenie heartbeat.
+# OPSGENIE_HEARTBEAT_CONFIG_KEY: API key provided by Opsgenie which is authorized to modify heartbeat configuration.
+# OPSGENIE_HEARTBEAT_DURATION_NUM: Specifies how often a heartbeat message should be expected.
+# OPSGENIE_HEARTBEAT_DURATION_UNIT: Interval specified as minutes, hours or days.
+
+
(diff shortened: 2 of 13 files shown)


dataeng/jobs/analytics/AnswerDistribution.groovy
```diff
@@ -24,11 +24,10 @@ class AnswerDistribution {
                 wrappers common_wrappers(allVars)
                 publishers common_publishers(allVars)
                 steps {
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-enable-heartbeat.sh'))
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/answer-distribution.sh'))
-                    if (env_config.get('OPSGENIE_HEARTBEAT_NAME') && allVars.get('OPSGENIE_HEARTBEAT_KEY')){
-                        shell("curl -X GET 'https://api.opsgenie.com/v2/heartbeats/" + env_config.get('OPSGENIE_HEARTBEAT_NAME') + "/ping' -H 'Authorization: GenieKey " + allVars.get('OPSGENIE_HEARTBEAT_KEY') + "'")
-                    }
-               }
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-disable-heartbeat.sh'))
```
doctoryes, line 29: Will these multiple `shell()` commands work? Will Jenkins simply glom these scripts together?

pwnage101: jenkins will invoke them one after the other as completely separate scripts, so if you want to persist anything between then you'll have to write to disk (under $WORKSPACE).

pwnage101: In this case, the enable and disable scripts are only making network calls so it should be fine.

dataeng/resources/opsgenie-enable-heartbeat.sh
```diff
@@ -0,0 +1,24 @@
+# Enables an Opsgenie Heartbeat, while also setting the expected duration.
```
pwnage101, line 1: Doesn't this only enable the heartbeat?  Don't you still have to ping it?

doctoryes: There will be no `/ping` issued. Instead, we'll simply use the `/disable` as the "ping" when the job finishes.

pwnage101: I assume you've tested this or considered whether it would basically work, and I haven't read the docs so I'll approve, but I'm still curious: is the enable/PATCH call is an implicit ping?

pwnage101 (approved):

dataeng/resources/opsgenie-enable-heartbeat.sh
doctoryes: @pwnage101 Turns out that I *did* need to ping it once after enabling it. So the PATCH call was *not* an implicit ping - fixed now.

pwnage101 (approved):

bmedx (approved):
Nice solution!

dataeng/resources/opsgenie-enable-heartbeat.sh
```diff
@@ -0,0 +1,31 @@
...
+# Required environment variables:
+# OPSGENIE_HEARTBEAT_NAME: Name of the already-created Opsgenie heartbeat.
+# OPSGENIE_HEARTBEAT_CONFIG_KEY: API key provided by Opsgenie which is authorized to modify heartbeat configuration.
+# OPSGENIE_HEARTBEAT_DURATION_NUM: Specifies how often a heartbeat message should be expected.
+# OPSGENIE_HEARTBEAT_DURATION_UNIT: Interval specified as minutes, hours or days.
+
+
+OPSGENIE_HEARTBEAT_API_URL="https://api.opsgenie.com/v2/heartbeats"
+
+if [ -n "$OPSGENIE_HEARTBEAT_NAME" ] && \
```
bmedx, line 15: Actually one question- if this script is getting called, shouldn't all of these exist? If they don't should we throw an error instead of ignoring it?

pwnage101: Look at the proposed job config for answer-distribution, we don't always configure these variables for every environment.  Ideally I think we would switch on these at a higher level (i.e. the job DSL level), like we do with setting the cron trigger, but I'm okay with either solution.

doctoryes: I guess my guiding principle here was that this heartbeat monitoring shouldn't interfere with the job triggering on error. Because of that, I also chose *not* to check the `curl` return codes either. A loud failure on a misconfiguration would arguably be better than silently ignoring that condition, as the code is currently written. But as we're doing this stuck-job checking for a limited number of jobs - and we're not on jenkins-new here - I'm going forward with this for now.

PR #1154, opened 2020-10-23
WS-1421 (fix): set prospectus sandbox to reconfigure, not recreate

srwang:
Recreate terminates the existing instance, whereas reconfigure re-runs ansible against the existing container which is still up.


rgraber (approved):
sure, let's try!

PR #1158, opened 2020-10-23
Remove the snowflake-collect-queue-depth job creation.

doctoryes:
Because it generates frequent & unnecessary alarms.

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
+++ b/dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
@@ -14,13 +14,8 @@ class SnowflakeCollectMetrics {
             NAME: 'snowflake-collect-credit-metrics',
             CRON: '0 * * * *'
         ]
-        Map SnowflakeQueueDepthConfig = [
-            NAME: 'snowflake-collect-queue-depth',
-            CRON: '*/10 * * * *'
-        ]
         List jobConfigs = [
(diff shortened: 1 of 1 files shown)


dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
```diff
@@ -14,13 +14,8 @@ class SnowflakeCollectMetrics {
...
             CRON: '0 * * * *'
         ]
-        Map SnowflakeQueueDepthConfig = [
-            NAME: 'snowflake-collect-queue-depth',
-            CRON: '*/10 * * * *'
-        ]
         List jobConfigs = [
-            SnowflakeWarehouseCreditConfig,
-            SnowflakeQueueDepthConfig
+            SnowflakeWarehouseCreditConfig
```
brianhw, line 18: nit: don't we normally leave commas at the end?

brianhw (approved):
I don't remember the details of the failures of this job, but I do have a memory of a job being discussed because it was noisy and unreliable.   If this is the job, then I'm fine with turning it off.

dataeng/jobs/analytics/SnowflakeCollectMetrics.groovy
doctoryes: As we'll probably be moving this remaining job over to the data health telemetry project eventually, I'm leaving this as is.

PR #1159, opened 2020-10-24
Add env command to heartbeat shell script invocation.

doctoryes:
Without this shebang, the environment variables aren't carried into the shell scripts, which is currently causing these scripts to do nothing because of the environment var conditionals.


brianhw (approved):
LGTM.  👍

PR #1164, opened 2020-10-30
Add python reqs upgrade job for edx-when

jmyatt:


testeng/jobs/upgradePythonRequirements.groovy
```diff
@@ -461,6 +462,19 @@ Map edxVal= [
     githubTeamReviewers: ['sustaining-vulcans'],
     emails: ['sustaining-vulcans@edx.org'],
     alwaysNotify: false
+]
+
+Map edxWhen= [
+    org: 'edx',
+    repoName: 'edx-when',
+    targetBranch: "master",
+    pythonVersion: '3.5',
```
mikix, line 471: Is edx-platform still on 3.5?

mikix (approved):
Sure?

testeng/jobs/upgradePythonRequirements.groovy
jmyatt: good point - no, 3.8

PR #1175, opened 2020-11-09
Make the upgrade job less error prone.

feanil:
It's easy to forget to add the name of the map object you create to the
list at the bottom of the file.  Instead of having that layer of
abstraction, simply put the maps directly in the list that's referenced
by the DSL.


robrap (approved):

jmbowman:
I'm a little concerned that this is a lot of diff churn to change the most common mistake from "my new job didn't get created because I forgot to add it to the list" to "the seed job failed because I forgot to add a comma after my list".  Thoughts?  Not particularly opposed, just not sure how much it will help either.

feanil:
My thinking is that right now, the place where you have to make the fix is not very local to where you might have made the change in the file.  In the new mistake of not getting the syntax quite right, it will at least be more local to where you made your change and so at least more likely to be caught in code-review.

Ideally, the testing on this repo would compile all the DSL as a part of the PR testing and catch the error automatically but that could be a future enhancement that could solve the syntax problem but would not catch the current problem.

PR #1180, opened 2020-11-12
Disable the video-timeline job while DENG-604 is in progress

pwnage101:
DENG-604

### DIFF
--- a/dataeng/jobs/analytics/VideoTimeline.groovy
+++ b/dataeng/jobs/analytics/VideoTimeline.groovy
@@ -14,7 +14,8 @@ class VideoTimeline {
             dslFactory.job("video-timeline-$environment") {
                 logRotator common_log_rotator(allVars, env_config)
                 multiscm common_multiscm(allVars)
-                triggers common_triggers(allVars, env_config)
+                // Disable this job while DENG-604 is in progress.
+                //triggers common_triggers(allVars, env_config)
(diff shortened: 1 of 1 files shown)


dataeng/jobs/analytics/VideoTimeline.groovy
```diff
@@ -14,7 +14,8 @@ class VideoTimeline {
             dslFactory.job("video-timeline-$environment") {
                 logRotator common_log_rotator(allVars, env_config)
                 multiscm common_multiscm(allVars)
-                triggers common_triggers(allVars, env_config)
+                // Disable this job while DENG-604 is in progress.
+                //triggers common_triggers(allVars, env_config)
```
doctoryes, line 18: Nit: I'd prefer a `disabled(true)` as it's more direct.

doctoryes (approved):

dataeng/jobs/analytics/VideoTimeline.groovy
pwnage101: The reason against doing it that way is that later I would need to re-enable the job manually to be able to create a manual build.  If I do it this way (just disabling the triggers), I do not need to reconfigure the job to be able to manually run it.

doctoryes: Good reason - TIL.

PR #1190, opened 2020-11-19
MICROBA-374 Send data to ITK only if MB data-mart build is successful

oliviaruizknott:
[MICROBA-374](https://openedx.atlassian.net/browse/MICROBA-374)

Here’s the documentation for `buildResult`, but it’s rather sparse!: https://jenkinsci.github.io/job-dsl-plugin/#path/job-triggers-buildResult

**Questions** (so many! I don’t know what I’m doing in here!)
- The docs say it "Requires BuildResultTrigger v0.17+"—https://plugins.jenkins.io/buildresult-trigger/#documentation. Uhh, do I need to install that?
- See questions in the comments on the diff

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
+++ b/dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
@@ -51,7 +51,10 @@ class SnowflakeMicrobachelorsITK {
                 }
             }
             triggers {
-                cron('H 6 * * 1-5')
+                buildResult('H 6 * * 1-5') {
+                    combinedJobs()
+                    triggerInfo('warehouse-transforms-microbachelors', BuildResult.SUCCESS)
+                }
             }
             wrappers {
                 timestamps()


dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -51,7 +51,10 @@ class SnowflakeMicrobachelorsITK {
                 }
             }
             triggers {
-                cron('H 6 * * 1-5')
+                buildResult('H 6 * * 1-5') {
+                    combinedJobs()
```
oliviaruizknott, line 55: Do I need this if there’s only one job?

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
```diff
@@ -51,7 +51,10 @@ class SnowflakeMicrobachelorsITK {
...
+                    triggerInfo('warehouse-transforms-microbachelors', BuildResult.SUCCESS)
```
oliviaruizknott, line 56: Does this groovy script know about warehouse-transforms-microbachelors? That’s the job name in Jenkins but that name is generated from some parameterization, I think.

Do I have to indicate BuildResult.SUCCESS or is that the default?

dataeng/jobs/analytics/SnowflakeMicrobachelorsITK.groovy
tuchfarber: I don't think we need this if it's only 1 job

PR #1191, opened 2020-11-19
Improve the UX of the event-export-incremental jobs

pwnage101:
* Add support for notifying the person who triggered a build if it fails.
* Fix an error in the help text for a job parameter.

DESUPPORT-596

### DIFF
--- a/src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
+++ b/src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
@@ -159,13 +159,15 @@ This text may reference other parameters in the task as shell variables, e.g.  $
         return {
             sshAgent('1')
             timestamps()
+            buildUserVars()
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -159,13 +159,15 @@ This text may reference other parameters in the task as shell variables, e.g.  $
         return {
             sshAgent('1')
             timestamps()
+            buildUserVars()
```
pwnage101, line 162: Since I'm enabling this plugin in common_wrappers, it will effectively enable the below feature (to automatically notify the builder) on all manually built jobs, not just the event-export-incremental ones.  I can't think of any reason this would be bad, and the code seemed simpler this way.

brianhw (approved):
Looks like a good improvement!  👍

doctoryes (approved):

PR #1193, opened 2020-11-20
Move Opsgenie heartbeat disabling to a post-build step.

doctoryes:
Although the Opsgenie heartbeat detection of potentially stuck jobs now works, the alarm is always triggered when the job fails. Previously to this PR, the heartbeat disabling was done in a separate build step. When the main build step fails, the heartbeat-disabling step is never run. This change will disable the heartbeat on all main task outcomes - success/failure/etc.


dataeng/jobs/analytics/AnswerDistribution.groovy
```diff
@@ -33,10 +32,15 @@ class AnswerDistribution {
                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
                     }
                 }
+                publishers {
+                    postBuildTask {
+                        task('Finished', 'if [ -n "$OPSGENIE_HEARTBEAT_NAME" ] && [ -n "$OPSGENIE_HEARTBEAT_CONFIG_KEY" ] then curl -X POST "https://api.opsgenie.com/v2/heartbeats/$OPSGENIE_HEARTBEAT_NAME/disable" --header "Authorization: GenieKey $OPSGENIE_HEARTBEAT_CONFIG_KEY" fi', true)
```
doctoryes, line 37: I don't like that this code is duplicated in each job file. Is there a way to include a separate shell file here? I couldn't find one.

pwnage101: or, preferrably use more functional semantics by defining it with parameters: `opsgenie_heartbeat_publisher(name, config_key)`, then define the function:
```groovy
    public static def opsgenie_heartbeat_publisher = { name, config_key ->
        return {
            postBuildTask {
                ...
            }
        }
    }
```

pwnage101 (approved):

adzuci (approved):

doctoryes:
@pwnage101 I had to keep the common function without parameters to keep the API config key hidden as an environment variable. Re-review?

src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -203,4 +203,14 @@ This text may reference other parameters in the task as shell variables, e.g.  $
             }
         }
     }
+
+    public static def opsgenie_heartbeat_publisher = {
+        // No args because the params need to be environment variables to remain hidden.
```
pwnage101, line 208: Nit: can you update the comment to also explicitly say which environment variables are required.

dataeng/jobs/analytics/VideoTimeline.groovy
```diff
@@ -30,11 +30,12 @@ class VideoTimeline {
                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
                     }
                 }
+                publishers common_publishers(allVars)
                 steps {
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-enable-heartbeat.sh'))
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/video-timeline.sh'))
-                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-disable-heartbeat.sh'))
                 }
+                publishers opsgenie_heartbeat_publisher
```
pwnage101, line 38: hm, I *think* it's still a function so it requires empty `()`

pwnage101 (approved):
otherwise LGTM!

dataeng/jobs/analytics/AnswerDistribution.groovy
```diff
@@ -33,11 +33,12 @@ class AnswerDistribution {
                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
                     }
                 }
+                publishers common_publishers(allVars)
                 steps {
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-enable-heartbeat.sh'))
                     shell(dslFactory.readFileFromWorkspace('dataeng/resources/answer-distribution.sh'))
-                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/opsgenie-disable-heartbeat.sh'))
                 }
+                publishers opsgenie_heartbeat_publisher
```
pwnage101, line 41: another Nit: put the publishers together.  They are grouped in the UI after the build steps, so I think it's reasonable to mirror that here.

pwnage101 (approved):

doctoryes:
@pwnage101 @estute Should I be worried about the Gradle exception which happened during this PR's CI?

pwnage101:
To be completely honest, I've been in the habit of ignoring these errors.  Unfortunately, I've never derived anything useful from them, but I also have no idea what they do.

doctoryes:
As always, I appreciate your honesty @pwnage101 😸 .

estute:
@doctoryes looks like it failed the quality step. See: 
```* What went wrong:
Execution failed for task ':codenarcMain'.
> CodeNarc rule violations were found.
```
and 
```
Caused by: : Exceeded maximum number of priority 2 violations: (p1=0; p2=93; p3=63)
```
I don't think anyone really looks at this stuff, so it's not a big deal. If you want to see what caused the failure, you can run:
```
./gradlew codenarcSrc
```

PR #1195, opened 2020-11-23
Remove broken feature to email build user

pwnage101:
Before this commit, when $NOTIFY is non-empty but $BUILD_USER_EMAIL is
empty, then the mailer plugin attempts to send emails to the string
literal "$BUILD_USER_EMAIL@edx.org" which results in a bounce.

I'm at my wit's end here.  I've evaluated several different combinations
of jenkins plugins and features, groovy scripts, etc.  So, I give up!


bmedx (approved):
Thanks for sorting this out

PR #1206, opened 2020-12-02
Disable Jenkins jobs for vertica upsize

brianhw:
for DENG-633

### DIFF
--- a/dataeng/jobs/analytics/BackupVertica.groovy
+++ b/dataeng/jobs/analytics/BackupVertica.groovy
@@ -8,6 +8,10 @@ class BackupVertica {
     public static def job = { dslFactory, allVars ->
         allVars.get('VERTICA_BACKUPS').each { backup, backup_config ->
             dslFactory.job("backup-vertica-$backup") {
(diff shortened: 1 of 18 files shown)


doctoryes (commented):
Question: How do you know you've disabled *all* the jobs that read from/write to Vertica here?

brianhw:
Fair question.   First, this only addresses those jobs that are seeded, so it doesn't address jobs that may be running that are outside of dsl.   I will be confirming that, but it's outside the scope of this PR.   Second, this started with those jobs that Troy identified last July, and included jobs that he missed in a first pass but were captured in code review.  Third, I then surveyed the remaining jobs to see if I knew of any that were added, or were otherwise missed last time.   (Two jobs have been removed since July, but I identified no jobs that needed to be added.)

doctoryes (approved):
I am satisfied with this answer. 😺

brianhw:
I ran reseeding from this branch, and when done, reseeded from master.   So this branch is no longer necessary.

PR #1211, opened 2020-12-07
Revert "ISRE-627 | Creating virtualenv without shining panda for run-ansible.sh"

christopappas:
Reverts edx/jenkins-job-dsl#1209


christopappas:
This breaks some, but not all, builds for reasons yet TBD. reverting until i can understand it better

PR #1212, opened 2020-12-08
Null and out and disable org-prefix upgrade jobs until they can be deleted

timmc-edx:
This is a follow-on to 19801196a/#1207.

### DIFF
--- a/testeng/jobs/upgradePythonRequirements.groovy
+++ b/testeng/jobs/upgradePythonRequirements.groovy
@@ -625,6 +625,12 @@ seenRepoNames = []
 /* Iterate over the job configurations */
 jobConfigs.each { jobConfig ->
 
+    // Null out and disable the version of these jobs that had the org
+    // in the name. After these are deleted, this block can be removed.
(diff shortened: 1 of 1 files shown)


feanil (approved):
So I suspect that the org names were in here because for a while we did have forks of repos in the edx-solutions org that also needed to be kept up-to-date on requirements.  It sounds like there aren't any on this list right now so I'm fine with doing this and adding the test you added in the earlier PR but just for future context, we have had times when we forked a repo and the code was in different places(older version) but we wanted to keep requirements up-to-date to get external security updates.

timmc-edx:
They actually weren't in there as of a few weeks ago. I noticed that the `org` field in the configurations wasn't actually being *used* anywhere, and decided it should be. In https://github.com/edx/jenkins-job-dsl/pull/1203 I used it in both the git checkout (rather than assuming it was always `edx`) and the job name.

But it makes the listing hard to read, so I partially undid that change, removing it from the job name: https://github.com/edx/jenkins-job-dsl/pull/1207 (but adding a bit of code to make sure it was never ambiguous).

PR #1220, opened 2020-12-15
split stitch validation jobs into unique instances

estute:
Currently the snowflake-validate-stitch job runs for multiple applications. A successful run using app X will close out the opsgenie alert for app Y, regardless of whether or not the issue was addressed. Similarly, a failure in app X should appear as a separate opsgenie alert from a failure for app Y.


macdiesel (approved):

macdiesel:
Thanks Stu!

PR #1224, opened 2020-12-16
reduce analytics exporter job concurrency

estute:
Jenkins has been very slow lately. I jumped on the box and often the exporter jobs are top resource users, aside from the Jenkins process.


bmedx (commented):
I'm not actually sure what the trade off is here time wise. We upped the number to deal with the master job running longer and longer, but it might be a better trade to push other things off to new Jenkins?

estute:
We upped it from 12 to 18, so this will hopefully strike the sweet spot, at least for a bit. I am also trying to get the dbt jobs onto the new Jenkins, but I just want to make sure jenkins is more usable for the next on-call.

bmedx (approved):

PR #1229, opened 2020-12-17
colorize dbt jobs

estute:
This will colorize the console logs and get rid of those messy color encoding characters


doctoryes (approved):
So many 👍🏼 s!

macdiesel (approved):

PR #1231, opened 2020-12-18
remove total events daily report

estute:
@brianhw said that this job can be removed and is no longer needed


brianhw (approved):
Yeah, I guess I still say that.  : )   👍

PR #1243, opened 2021-01-04
Ihassan/isre 710 update jenkins jobs python

syedimranhassan:


coryleeio (approved):
This looks right to me, but you will want to be ready to roll it back if some or all of them break / can't handle python3 etc, you might wish to do a smaller subset as some may have issues with python3.8. but doing them all at once and seeing what breaks is understandable, just be ready to roll it back if you do it wholesale like this.

I'm especially worried about some of the scripts maybe needing small tweaks in the configuration repo since that repo is python 2.7 for ansible. 

That said, probably the easiest way to know is to try it.

PR #1247, opened 2021-01-05
Add separate section to transfer model data to S3 via DBT macro.

doctoryes:
https://openedx.atlassian.net/browse/DENG-678

Works in concert with these PRs:
https://github.com/edx-ops/analytics-secure/pull/495
https://github.com/edx/warehouse-transforms/pull/1419

### DIFF
--- a/dataeng/resources/warehouse-transforms.sh
+++ b/dataeng/resources/warehouse-transforms.sh
@@ -61,3 +61,47 @@ then
     # Run all tests which haven't been excluded.
     dbt test --models tag:$MODEL_TAG $exclude_param --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
 fi
+
+# Each transfer is two elements in this array:
+#     - the DBT macro name
+#     - the macro arguments to use when calling the DBT macro
+PROGRAMS_REPORTING_TRANSFERS=( \
+    unload_program_reporting_learner_enrollments_report '{database: PROD, schema: PROGRAMS_REPORTING}' \
+    unload_program_reporting_program_cohort_report '{database: PROD, schema: PROGRAMS_REPORTING}' \
+)
+
+if [ "$PROGRAMS_REPORTING_TRANSFER_TO_S3" = 'true' ]
+then
+
+    # Call each DBT macro to initiate a transfer.
+    num_txfrs=${#PROGRAMS_REPORTING_TRANSFERS[@]}/2
+
+    for (( i=0; i<${num_txfrs}; i++ ));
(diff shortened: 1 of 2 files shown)


dataeng/resources/warehouse-transforms.sh
```diff
@@ -61,3 +61,47 @@ then
     # Run all tests which haven't been excluded.
     dbt test --models tag:$MODEL_TAG $exclude_param --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
 fi
+
+# Each transfer is two elements in this array:
+#     - the DBT macro name
+#     - the macro arguments to use when calling the DBT macro
+PROGRAMS_REPORTING_TRANSFERS=( \
```
macdiesel, line 68: I'm wondering if instead of putting this data here and having people edit config in yet another repo (4 all together for this PR I believe) could we put this configuration in to the DBT layer? If we could figure something out there it would allow us to remove at least 1 repos worth of PRs while creating a new s3 transfer.

macdiesel: That would also enable us to DRY this code up some more. If we did so we could have a single jenkins job that runs all the s3 exports and pushes each functional area or project to an sub job to be executed. 

Also I'm wondering about execution schedule... Should this be in here at all? Given that different teams and functions may want to execute these exports on different schedules and different times we might be better served by moving this functionality to prefect?

bmedx (approved):
I agree with that @macdiesel is saying, but since we're having multiple failures right now due to not having these PRs merged maybe we can get this running and then iterate?

dataeng/resources/warehouse-transforms.sh
doctoryes: Re: editing config here:
Agreed that this is not ideal. We should minimize the number of places to edit config as much as possible. That means we'd move this config to warehouse-transfers somehow (maybe a seed? a macro included in other macros?) so that it could be editable by other teams (as opposed to analytics-secure).

Re: a single Jenkins job & execution schedule:
I separated the transfers into two batches because the transferred models were generated by two different Jenkins jobs - w-t-daily and w-t-enterprise (which is triggered after w-t-daily). I chose to keep the transfers happening after the models being transferred were built. My assumption was that newly-built models should be transferred immediately after being built. I guess we could have a separate "w-t-xfer-models" Jenkins job that gets triggered by multiple upstream jobs if we wanted?

doctoryes:
Further work has been ticketed here:

https://openedx.atlassian.net/browse/DENG-707
https://openedx.atlassian.net/browse/DENG-713

PR #1259, opened 2021-01-14
Remove event-type-distribution job from seed

bmedx:


doctoryes (commented):
There's also a stray reference here:
https://github.com/edx/jenkins-job-dsl/blob/b0e064fcd820036a6abf075b33592621c30e4acc/dataeng/jobs/createJobsNew.groovy#L93

bmedx:
@doctoryes should be fixed

doctoryes (approved):

PR #1276, opened 2021-01-22
Fix name of model_transfers job.

doctoryes:
There's many a slip betwixt cup and lip...


brianhw (approved):
Argh -- sorry to have missed that. 👍

PR #1285, opened 2021-02-01
update Python version to run pip-compile under for Registrar

MichaelRoytman:
Update the version of Python that `pip-compile` runs under for the Registrar Python Requirements Updates jobs that Jenkins runs. Registrar currently supports Python 3.8 and runs also Travis under Python 3.8.

A new [version](https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-600154) of the `newrelic` Python agent does not support Python 3.5.


mattcarter (approved):
approved

matthugs (approved):

PR #1287, opened 2021-02-02
chore: Delete extraneous upgrade-requirements jobs

timmc-edx:
I renamed the upgrade jobs to include the org name, but the names ended
up much harder to read, so I renamed them back again. However, this left
a lot of extra jobs around, since seeding on this server doesn't clean up
old jobs.

I then set all the old jobs to be disabled, but they're still around and
confusing people.

This script will list all the jobs I want to delete, and then I'll add
a job.delete() once it looks good and has gotten several approvals.

Process:

- [x] Iterate without review by branch-seeding the script in print-only mode until the results look correct
- [x] Get multiple approval reviews
- [ ] Merge
- [ ] Add a commit to a new branch adding a `job.delete()` call after the println
- [ ] "Seed" from that branch
- [ ] Delete that branch


timmc-edx:
Looking good: https://build.testeng.edx.org/job/manually-seed-one-job/909/console

Prints 54 upgrade jobs, and unreferenced job list contains 108 upgrade jobs -- double the number printed, as expected.

testeng/jobs/manageJenkinsJobs.groovy
```diff
@@ -0,0 +1,13 @@
+import jenkins.model.*
+
+// Ad-hoc job management script.
```
coryleeio, line 3: NIT: 
I would call this Janitor or cleanup or something of the sort, just to use existing nomenclature, I would not type manage if I was trying to find this

coryleeio (approved):

adzuci (approved):

timmc-edx:
Deletion performed, everything looks good:

- https://github.com/edx/jenkins-job-dsl/commit/0217d492aed7846fb88671b491a95bb14b3387a4
- https://build.testeng.edx.org/job/manually-seed-one-job/912/console

PR #1288, opened 2021-02-02
Stop using shiningpanda for user retirement jobs

pwnage101:
Shiningpanda was creating an impossible situation for python 3.5 because
it seems to be hard-coded to upgrade pip, but the latest pip (21) has
dropped python 3.5 support.  Also, shiningpanda is no longer supported
upstream, supposedly, so the rest of the edx org has moved away from it.

Additionally, I have changed the python version for these jobs from 3.5
to 3.8.

---
Manually tested in https://build.testeng.edx.org/view/user-retirement-jobs/job/user-retirement-collector/39665/console

### DIFF
--- a/platform/resources/user-retirement-collector.sh
+++ b/platform/resources/user-retirement-collector.sh
@@ -8,6 +8,13 @@ set -ex
 # just stores a filename in the environment (rather than the content).
 env
 
+# Create and activate a virtualenv.  In case we ever change the concurrency
+# setting on the jenkins worker, it would be safest to keep the builds from
+# clobbering each other's virtualenvs.
+VENV="venv-${BUILD_NUMBER}"
+virtualenv --python=python3.8 --clear "${VENV}"
+source "${VENV}/bin/activate"
+
 # Make sure that when we try to write unicode to the console, it
 # correctly encodes to UTF-8 rather than exiting with a UnicodeEncode
 # error.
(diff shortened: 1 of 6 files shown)


platform/resources/user-retirement-collector.sh
```diff
@@ -8,6 +8,10 @@ set -ex
 # just stores a filename in the environment (rather than the content).
 env
 
+# create and activate a virtualenv
+virtualenv --python=python3.8 venv
```
bmedx, line 12: Is there a chance this venv will already exist from a previous run? I'm not where this sits in the process, but I was under the impression that workers will be re-used more than once. Not sure if this will fail if `venv` already exists.

doctoryes (approved):
This is great - 👍🏼 . A question:
- Should a follow-on ticket be made to completely remove shiningpanda from all analytics jobs? It seems like work we can and should do.

Hmmm - just occurred to me: These tests happened on the user-retirement-specific AMI which I built last week. I now wonder if that work can be reverted and if these user retirement jobs can be run on a regular Jenkins worker - which would be simpler. The work:
https://github.com/edx-ops/testeng-secure/pull/267
https://github.com/edx/configuration/pull/6265
https://github.com/edx/configuration/pull/6266
https://github.com/edx/configuration/pull/6267
https://github.com/edx/configuration/pull/6274

platform/resources/user-retirement-collector.sh
pwnage101: whoops, I think I missed adding a `--clear` here

pwnage101: I addressed this by both adding `--clear` (a safety measure) and also by creating a unique venv directory name for each build, as discussed on slack with @doctoryes .  The idea is that we don't need them to be unique today, but in the future somebody could flip the switch to make workers allow concurrent jobs and this would protect against that.  I'm on the fence between adding complexity vs. ignoring a possible time bomb so I just picked one at random (the former).

PR #1295, opened 2021-02-08
Added automated/reporting project (WT-CI job)

jazibhumayun:
Added support for automated/reporting project.

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransformsCI.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsCI.groovy
@@ -10,6 +10,7 @@ import static org.edx.jenkins.dsl.JenkinsPublicConstants.GHPRB_CANCEL_BUILDS_ON_
 class WarehouseTransformsCI{
     public static def job = { dslFactory, allVars ->
         dslFactory.job("warehouse-transforms-ci"){
+            authorization common_authorization(allVars)
             logRotator common_log_rotator(allVars)
             parameters secure_scm_parameters(allVars)
             parameters {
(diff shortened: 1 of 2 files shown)


dataeng/resources/warehouse-transforms-ci.sh
```diff
@@ -16,11 +16,13 @@ aws s3 cp s3://edx-dbt-docs/manifest.json ${WORKSPACE}/manifest
 cd $WORKSPACE/warehouse-transforms
 
 # Pull the origin master code to latest branch which will be used to compare the diff
-git pull origin master:latest
+git pull -f origin master:latest
 
 # Put back the head at PR commit
 git checkout ${ghprbActualCommit}
 
+git rebase latest
```
jazibhumayun, line 24: It is better to do re-basing from master before merging it, so that merge conflicts if any can be resolved. If Warehouse Transform user missed to re-base from master the following git diff command might catch the files that are already tested and merged into master and this job will take extra time to build/test those models as well. That is why I have added this git re-base in the job as well, this will help us to optimize we will only run the modified models as a result the PR in question

dataeng/resources/warehouse-transforms-ci.sh
```diff
@@ -69,9 +71,21 @@ fi
 if [ "$isApplications" == "true" ]
```
jazibhumayun, line 71: I added these lines manually in the shell script of jenkins dsl, it ran fine over the weekend and I think we are good to merge this.

dataeng/jobs/analytics/WarehouseTransformsCI.groovy
```diff
@@ -10,6 +10,7 @@ import static org.edx.jenkins.dsl.JenkinsPublicConstants.GHPRB_CANCEL_BUILDS_ON_
 class WarehouseTransformsCI{
     public static def job = { dslFactory, allVars ->
         dslFactory.job("warehouse-transforms-ci"){
+            authorization common_authorization(allVars)
```
jazibhumayun, line 13: Adding job View permissions for DS&A members

bmedx (approved):
lgtm

PR #1301, opened 2021-02-10
py38 for edx-platform venv

jazibhumayun:
switching to py38 for edx-platform venv - analytics-email-optin-master

### DIFF
--- a/dataeng/resources/setup-platform-venv-py3.sh
+++ b/dataeng/resources/setup-platform-venv-py3.sh
@@ -3,12 +3,18 @@
 # This version of setup-platform-venv is specifically designed for newer
 # python3 versions of the platform.
 
+
+# Create and activate a virtualenv in shell script
+PLATFORM_VENV="platform_venv"
+virtualenv --python=python3.8 --clear "${PLATFORM_VENV}"
+source "${PLATFORM_VENV}/bin/activate"
+
 # Install requirements
 pushd edx-platform
 # This is the same pip version we currently pin in our devstack edxapp container:
-pip install pip==20.0.2
+pip install pip==20.2.3
 make requirements
 popd
 
 # Save virtualenv location
-echo "PLATFORM_VENV=${VIRTUAL_ENV}" > platform_venv
+echo "PLATFORM_VENV=${WORKSPACE}/${PLATFORM_VENV}" > platform_venv
--- a/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
+++ b/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
@@ -136,13 +136,8 @@ class AnalyticsEmailOptin {
             publishers common_publishers(allVars)
 
             steps {
-                virtualenv {
-                    pythonName('PYTHON_3.7')
-                    nature("shell")
-                    command(
-                        dslFactory.readFileFromWorkspace("dataeng/resources/setup-platform-venv-py3.sh")
-                    )
-                }
+                // This will create python 3.8 venv inside shell script instead of using shiningpanda
+                shell(readFileFromWorkspace('platform/resources/setup-platform-venv-py3.sh'))
                 virtualenv {
                     // The exporter itself still runs python 2.
                     nature("shell")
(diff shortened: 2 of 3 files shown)


dataeng/resources/setup-platform-venv-py3.sh
```diff
@@ -3,12 +3,16 @@
...
 # Install requirements
 pushd edx-platform
 # This is the same pip version we currently pin in our devstack edxapp container:
-pip install pip==20.0.2
+pip install pip==20.2.3
 make requirements
 popd
 
-# Save virtualenv location
-echo "PLATFORM_VENV=${VIRTUAL_ENV}" > platform_venv
```
jazibhumayun, line 14: This will be excluded right ? @HassanJaveed84

HassanJaveed84: We still need to as the worker jobs need to know the location of platform's virtualenv. See comment above.

dataeng/resources/setup-platform-venv-py3.sh
```diff
@@ -3,12 +3,16 @@
 # This version of setup-platform-venv is specifically designed for newer
 # python3 versions of the platform.
 
+
+# Create and activate a virtualenv in shell script
+VENV="venv-${BUILD_NUMBER}"
```
HassanJaveed84, line 8: We can be more explicit with the naming here. Something like
`PLATFORM_VENV="platform_venv"`

I don't think we need a separate venv per build, unless we also remove the ones from prior runs.
Also, there is only one instance of this job running at any point, so I think just using it with `--clear` should work.

We need to echo it to the file as it is needed for the workers. And I believe we would need an absolute path as
that's how  `VIRTUAL_ENV` behaved.

`echo "PLATFORM_VENV=${WORKSPACE}/${PLATFORM_VENV}" > platform_venv`

dataeng/resources/setup-platform-venv-py3.sh
pwnage101: I expect you'll still need this:
```
echo "PLATFORM_VENV=${VENV}" > platform_venv
```
the file `platform_venv` is used in downstream jobs, e.g. https://github.com/edx/jenkins-job-dsl/blob/7dda40199dcf01170a471ba552fc8cd2816297b1/dataeng/resources/setup-exporter.sh#L40

pwnage101 (changes requested):

HassanJaveed84:
We also need to copy over these changes to the two jobs defined here
https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/AnalyticsExporter.groovy

dataeng/resources/setup-platform-venv-py3.sh
jazibhumayun: ok right,
should be echo "PLATFORM_VENV=${WORKSPACE}/${**VIRTUAL_VENV**}" > platform_venv ??

jazibhumayun: Previously it is exporting the path from `VIRTUAL_ENV` not sure where that is coming from but is the current push correct or should it be `echo "PLATFORM_VENV=${WORKSPACE}/${PLATFORM_VENV}" > platform_venv`

@HassanJaveed84

jazibhumayun:
> We also need to copy over these changes to the two jobs defined here
> https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/AnalyticsExporter.groovy

Yes, sure will do

dataeng/resources/setup-platform-venv-py3.sh
```diff
@@ -3,12 +3,18 @@
...
 pushd edx-platform
 # This is the same pip version we currently pin in our devstack edxapp container:
-pip install pip==20.0.2
+pip install pip==20.2.3
 make requirements
 popd
 
 # Save virtualenv location
-echo "PLATFORM_VENV=${VIRTUAL_ENV}" > platform_venv
+echo "PLATFORM_VENV=${WORKSPACE}/${VIRTUAL_ENV}" > platform_venv
```
HassanJaveed84, line 20: this would be `${WORKSPACE}/${PLATFORM_VENV}`

dataeng/resources/setup-platform-venv-py3.sh
HassanJaveed84: Yes, it should be PLATFORM_VENV, VIRTUAL_ENV was coming from shiningpanda I believe.

HassanJaveed84 (approved):

PR #1303, opened 2021-02-10
Add heartbeats for the WarehouseTransforms jobs

bmedx:


brianhw (commented):
Looking at existing uses of this, I find an OPSGENIE key mentioned:
```
                wrappers {
                    credentialsBinding {
                    }
                }
```
Is that not necessary here?

bmedx:
> Looking at existing uses of this, I find an OPSGENIE key mentioned:
> 
> ```
>                 wrappers {
>                     credentialsBinding {
>                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
>                     }
>                 }
> ```
> 
> Is that not necessary here?

Good catch, added.

macdiesel (approved):

PR #1304, opened 2021-02-11
fix vendor monitoring job

nadeemshahzad:


syedimranhassan (approved):

engagement/jobs/WebsiteVendorMonitoring.groovy
```diff
@@ -46,8 +46,9 @@ class WebsiteVendorMonitoring {
             multiscm {
                 git {
                     remote {
-                        url('https://github.com/edx/prospectus.git')
+                        url('git@github.com:edx/prospectus.git')
                         branch('master')
+                        credentials(gitCredentialId)
```
adzuci, line 51: Is the credential above correct / where is it set?

adzuci (approved):

engagement/jobs/WebsiteVendorMonitoring.groovy
nadeemshahzad: I have defined the var in edx-internal https://github.com/edx/edx-internal/pull/4420

PR #1305, opened 2021-02-11
raw_to_source and notification settings -  wt ci job dsl

jazibhumayun:


dataeng/resources/warehouse-transforms-ci.sh
```diff
@@ -43,6 +43,9 @@ pip install -r requirements.txt
 if [ "$isReporting" == "true" ]
 then
 
+    # Overide the commit author in case of reporting
+    NOTIFY=${ghprbActualCommitAuthorEmail}
```
jazibhumayun, line 47: I am trying to overwrite NOTIFY email address to Commit author email authors, commit author emails do appear but NOTIFY variable is not overwritten, Is is it possible if we have some variable defined in DSL and can be overwritten inside shell script?

jazibhumayun: @bmedx

bmedx:
I'm not sure about this one, I think @estute or @HassanJaveed84 might know better than I do.

PR #1311, opened 2021-02-11
Adding UI hooks for Sandbox config of needed enterprise repositories

edx-abolger:
Adding additional groovy UI flags that are needed to enable the Sandbox to support the Enterprise use case.


jlajoie (approved):
👍

PR #1316, opened 2021-02-16
Change txfr script to retrieve macro calls from DBT and execute them.

doctoryes:
https://openedx.atlassian.net/browse/DENG-707

Using the macro added in this PR: https://github.com/edx/warehouse-transforms/pull/1713, modify the DBT model transfer script to call the macro to retrieve the information about which DBT models to transfer to S3 via a DBT macro using Snowflake stage and a `COPY..INTO` command.

This new pattern will allow for more models to be added to the transfers more easily.

### DIFF
--- a/dataeng/resources/model-transfers.sh
+++ b/dataeng/resources/model-transfers.sh
@@ -9,49 +9,40 @@ pip install -r requirements.txt
 
 cd $WORKSPACE/warehouse-transforms/projects/$DBT_PROJECT
 
-
-# Each transfer is two elements in this array:
-#     - the DBT macro name
-#     - the macro arguments to use when calling the DBT macro
-DAILY_TRANSFERS=( \
-    unload_program_reporting_learner_enrollments_report '{database: PROD, schema: PROGRAMS_REPORTING}' \
-    unload_program_reporting_program_cohort_report '{database: PROD, schema: PROGRAMS_REPORTING}' \
-)
-
+# Choose the marts from which to transfer DBT models based on Jenkins job parameter.
 if [ "$MODELS_TO_TRANSFER" = 'daily' ]
 then
-
-    # Call each DBT macro to initiate a transfer.
-    num_txfrs=${#DAILY_TRANSFERS[@]}/2
-
-    for (( i=0; i<${num_txfrs}; i++ ));
-    do
-        dbt run-operation ${DAILY_TRANSFERS[$i*2]} --args "${DAILY_TRANSFERS[$i*2+1]}" --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
-    done
-
+    MART_NAME=programs_reporting
+elif [ "$MODELS_TO_TRANSFER" = 'enterprise' ]
+    MART_NAME=enterprise
 fi
 
-# TODO: Adding single_file argument is a short term solution. We have created
-# https://openedx.atlassian.net/browse/ENT-4087 to find a better way.
-ENTERPRISE_TRANSFERS=( \
-    enterprise_copy_to_s3 '{prefix: ENTERPRISE, table: ENT_BASE_ENTERPRISE_USER}' \
-    enterprise_copy_to_s3 '{prefix: ENTERPRISE, table: FACT_CUSTOMER_COURSE_DAILY_ROLLUP_ADMIN_DASH}' \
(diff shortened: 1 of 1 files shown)


dataeng/resources/model-transfers.sh
```diff
@@ -9,49 +9,31 @@ pip install -r requirements.txt
...
-        dbt run-operation ${ENTERPRISE_TRANSFERS[$i*2]} --args "${ENTERPRISE_TRANSFERS[$i*2+1]}" --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
-    done
-
-fi
+ARGS="{mart: ${MART_NAME} }"
+
+# Currently, only the programs_reporting mart's transfers are performed in the daily transfer.
+#
+# Call DBT to query the list of macros to call and the macro parameters to perform all transfers.
+dbt run-operation get_all_s3_transfers --args "${ARGS}" > dbt_get_data_output_tmp.txt || exit
```
pwnage101, line 25: This feels like an attack vector, given that it's "download a list of arbitrary commands and run them" vs. "download a list of command arguments and run them using the expected commands".  Would it still align with your vision to implement the latter?

pwnage101: Hm i see now that there's multiple different possible commands that are being sent, e.g. `enterprise_copy_to_s3` and `unload_program_reporting_learner_enrollments_report` and `unload_program_reporting_program_cohort_report`, so I can understand the convenience of the simpler approach of executing what's sent.

An allow-list seems clunky.  Have you given this any thought from a security perspective and have any insights that can address my concern?

dataeng/resources/model-transfers.sh
```diff
@@ -9,49 +9,31 @@ pip install -r requirements.txt
...
+
+# Parse the output to extract the list of macros/parameters from the JSON and convert to DBT commands.
+commands=$( sed -n '/^\[/p' dbt_get_data_output_tmp.txt | jq '.[] | ["dbt run-operation " + .[0] + " --args " + "\"" + .[1] + "\""]' )
```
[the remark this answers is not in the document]
doctoryes: I will.

dataeng/resources/model-transfers.sh
doctoryes: Yes - I made this solution flexible wrt the macro executed for the sake of flexibility. I see the concern around security here. Some thoughts:

- Arbitrary commands are not available to the caller. Instead only the macros within warehouse-transforms are available to execute. All those macros are in a private GH repository and are reviewed by our usual code review process.
- An external/internal network service is *not* being called to provide the macro names/parameters - it's the same DBT that other Jenkins jobs execute as well. So it's almost as if the new DBT macro is simply reading the configuration from a config file in warehouse-transforms and providing it.
- If the security of the warehouse-transforms repo used on a Jenkins worker is compromised, then even a `dbt run` command is no longer safe, as those commands execute SQL and can be used to R/W/delete data as well.
- So, in an indirect sense, there's already an allow-list implemented, which are all the macros in warehouse-transforms - because whatever is returned by the DBT macro command is appended to a `dbt run-operation` prefix.

All this being said: yes, a call is made which writes a file to disk, which is read and appended to a DBT macro call. So it's different than the usual "execute the commands in this committed file". If this difference is unacceptable, I'll attempt to devise another way to do this.

pwnage101: That all makes sense.  With all of that perspective it seems the attacker would need enough privileges which would make other attacks more practical.

dataeng/resources/model-transfers.sh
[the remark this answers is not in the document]
doctoryes: Added.

doctoryes:
@pwnage101 Re-review?

pwnage101:
Drat, whatever you told me on the call yesterday I interpreted as the opposite of this most recent change.  Essentially, now this not only fetches dbt macro names, but entire shell statements.  Unfortunately, I think it's less secure than it was before.  I'm basically okay with how it was before.

For some reason I interpreted from you that there was a way to make it so that you could make jenkins call a single dbt macro at the top level of the bash script, and that dbt macro in-turn calls a hard-coded list of other macros, the names of which would otherwise be passed via JSON back to this jenkins job.

doctoryes:
Closing in favor of https://github.com/edx/jenkins-job-dsl/pull/1319

PR #1318, opened 2021-02-18
Re-order the generated date for branch and PR names.

doctoryes:
Change the current date ordering of:
`18_02_2021_14_34_45`
to:
`2021_02_18_14_34_45`
which is in order of least granular component to most granular component. It could be only me - but I find this ordering easier to read.


bmedx (approved):
Been meaning to do this forever 👍🏼

PR #1324, opened 2021-02-22
enable concurrent Builds

jazibhumayun:


pwnage101 (approved):

estute (approved):
In terms of the DSL, yes, this is correct. The only thing I would like to know is if there are any implications on the Snowflake infrastructure/validity of the tests. I tend to think not, as you create a separate schema for each run, but I just wanted to double check.

PR #1327, opened 2021-02-24
COOL_OFF_DAYS parameter set to 14

jazibhumayun:
DESUPPORT-763
Reference thread: https://edx-internal.slack.com/archives/CDATH3P6V/p1613676995154500


doctoryes (approved):
👍🏼 You can now use this job to re-seed it:

https://build.testeng.edx.org/job/manually-seed-one-job/

using the filename changed here, which is:

platform/jobs/RetirementJobs.groovy

PR #1331, opened 2021-03-01
Bump the oep2-report job to python3.8

UsamaSadiq:


christopappas (changes requested):
It would be cleaner to add in the venv calls directly in the script before the pip commands if possible: https://github.com/edx/jenkins-job-dsl/blob/master/testeng/resources/create-oep-report.sh#L1

You may have seen me add to the `String script = ` in backupsJenkins.groovy, but I did that because the pattern was already there. I think we should avoid adding more bits like that 👍

christopappas (approved):

PR #1334, opened 2021-03-04
Source the correct file to get the platform venv

pwnage101:
`platform_venv` is now a directory, not a file.  The correct filename is `platform_venv_path`.

### DIFF
--- a/dataeng/resources/run-course-exporter.sh
+++ b/dataeng/resources/run-course-exporter.sh
@@ -10,7 +10,7 @@ pip install mysql-connector-python -e .
 popd
 
 # Get name of other (platform) virtual environment
-source platform_venv
+source platform_venv_path
 
 # Configuration paths in analytics-secure
 SECURE_ROOT=${WORKSPACE}/analytics-secure/analytics-exporter


bmedx (approved):

dataeng/resources/run-course-exporter.sh
```diff
@@ -10,7 +10,7 @@ pip install mysql-connector-python -e .
 popd
 
 # Get name of other (platform) virtual environment
-source platform_venv
+source platform_venv_path
```
doctoryes, line 13: I thought the course exporter uses Python 2 or "legacy" still? As defined here?:

https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/resources/setup-platform-venv-legacy.sh#L23-L22

This change implies it uses this setup instead?:

https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/resources/setup-platform-venv-py3.sh#L21

pwnage101: There are 2 virtualenvs at play.  The platform one is python 3, and the exporter one is python 2.  This line is sourcing the file containing the path to the python 3 platform venv, but this script `run-course-exporter.sh` is running inside a python 2 venv.  Is that confusing enough?

pwnage101: The setup-platform-venv-*.sh files are for creating the platform virtualenv, and this `run-course-exporter.sh` file creates the exporter py2 venv using shiningpanda (i.e. virtualenv build step).

PR #1335, opened 2021-03-05
WT CI Manual Job

jazibhumayun:
If some tests fails on your Warehouse Transform CI job and re-running the whole job is time consuming, this job provides short way of running CI only on failed tests

Warehouse Transform Manual CI job will be used to run only specific models that failed, specify GITHUB_PR_ID, WAREHOUSE_TRANSFORMS_BRANCH, marks that project as true and specify DBT_TEST_OPTIONS_<PROJECT-NAME>

Document: https://openedx.atlassian.net/wiki/spaces/DE/pages/2464776611/CI+Warehouse+Transform+Manual+Job
Ticket: https://openedx.atlassian.net/browse/DENG-788

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransformsCIManual.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsCIManual.groovy
@@ -1,0 +1,103 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_authorization
+import static org.edx.jenkins.dsl.JenkinsPublicConstants.GHPRB_CANCEL_BUILDS_ON_UPDATE
+
+class WarehouseTransformsCIManual{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("warehouse-transforms-ci-manual"){
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('WAREHOUSE_TRANSFORMS_URL', allVars.get('WAREHOUSE_TRANSFORMS_URL'), 'URL for the warehouse-transforms repository.')
+                stringParam('WAREHOUSE_TRANSFORMS_BRANCH', '', 'Must specify branch of warehouse-transforms repository to use.')
+                stringParam('GITHUB_PR_ID', '', 'Must specify Github Pull Request Id')
+                stringParam('DBT_TARGET', allVars.get('DBT_TARGET'), 'DBT target from profiles.yml in analytics-secure.')
+                stringParam('DBT_PROFILE', allVars.get('DBT_PROFILE'), 'DBT profile from profiles.yml in analytics-secure.')
+                stringParam('DBT_PROJECT_PATH', allVars.get('DBT_PROJECT_PATH'), 'Path in warehouse-transforms to use as the dbt project, relative to "projects" (usually automated/applications or reporting).')
+                stringParam('RUN_DBT_TESTS_ONLY', allVars.get('RUN_DBT_TESTS_ONLY'), 'Mark it as false if you want to run dbt models')
+                stringParam('DBT_RUN_OPTIONS', allVars.get('DBT_RUN_OPTIONS'), 'Additional options to dbt run, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_RUN_EXCLUDE', allVars.get('DBT_RUN_EXCLUDE'), 'Additional options to dbt run, such as --exclude. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_OPTIONS', allVars.get('DBT_TEST_OPTIONS'), 'Additional options to dbt test, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_EXCLUDE', allVars.get('DBT_TEST_EXCLUDE'), 'Additional options to dbt test, such as "--exclude <model>". Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DB_NAME', allVars.get('DB_NAME'), 'Database name used to create output schema of dbt run/tests')
+                stringParam('NOTIFY', allVars.get('NOTIFY'), 'Space separated list of emails to send notifications to.')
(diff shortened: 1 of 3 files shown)


pwnage101:
Is it possible to make the job insensitive to new or renamed dbt projects?  For example, our current dbt-manual takes a project path from the root of the repo via the DBT_PROJECT_PATH parameter jenkins-new.analytics.edx.org/job/dbt-manual/build instead of exploding the parameters to cover every possible dbt project which makes the list of parameters very long.   The benefit of the exploded approach taken in this PR currently is that it allows simultaneous testing of models in multiple projects, but I think that's a rare use case.  The un-exploded approach also allows us to set a default project of `reporting` so DS&A don't have to take manual steps to enable the project for testing.

pwnage101:
Also, we are adopting conventional commits now: https://open-edx-proposals.readthedocs.io/en/latest/oep-0051-bp-conventional-commits.html So you should update your commit name and message!

jazibhumayun:
> Is it possible to make the job insensitive to new or renamed dbt projects? For example, our current dbt-manual takes a project path from the root of the repo via the DBT_PROJECT_PATH parameter http://jenkins-new.analytics.edx.org/job/dbt-manual/build instead of exploding the parameters to cover every possible dbt project which makes the list of parameters very long. The benefit of the exploded approach taken in this PR currently is that it allows simultaneous testing of models in multiple projects, but I think that's a rare use case. The un-exploded approach also allows us to set a default project of `reporting` so DS&A don't have to take manual steps to enable the project for testing.

Initially I also thought to use DBT_PROJECT_PATH parameter like its is used in dbt-manual job but the reason for parameter per project is exactly what you mentioned, "simultaneous testing for multiple projects" it might happen some team member wants to use this job for running more than one project in a given run but it makes the handling of this manual difficult.
I am fine with both approaches, whichever we decides to proceed with. @pwnage101

jazibhumayun:
> Also, we are adopting conventional commits now: https://open-edx-proposals.readthedocs.io/en/latest/oep-0051-bp-conventional-commits.html So you should update your commit name and message!

sure will do, when I will squash commit next time!

pwnage101:
If somebody wants to run multiple projects, can't they just create multiple manual builds?  I also strongly suspect people will only be changing one project at a time; do we know of cases where that is not true?

It also seems best to make this job take the same parameters as dbt-manual which some members of DS&A are already familiar with and so they will not need to re-learn.

jazibhumayun:
@pwnage101 This is up for second review, I have changed it to run one project in one run just like dbt-manual job.

jazibhumayun:
I have seen some PRs touching more than one projects at a time but those are not very common and people will usually be using the regular CI job. I have made it similar to dbt-manual so it will be easier to use for DS&A, I myself think having bunch of parameters for every project was making this job complex.

dataeng/jobs/analytics/WarehouseTransformsCIManual.groovy
```diff
@@ -0,0 +1,104 @@
...
+                stringParam('DBT_PROJECT_PATH', allVars.get('DBT_PROJECT_PATH'), 'Path in warehouse-transforms to use as the dbt project, relative to "projects" (usually automated/applications or reporting).')
+                stringParam('RUN_DBT_TESTS_ONLY', allVars.get('RUN_DBT_TESTS_ONLY'), 'Mark it as false if you want to run dbt models')
+                stringParam('DBT_RUN_OPTIONS', allVars.get('DBT_RUN_OPTIONS'), 'Additional options to dbt run, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_RUN_EXCLUDE', allVars.get('DBT_RUN_EXCLUDE'), 'Additional options to dbt run, such as --exclude. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_OPTIONS', allVars.get('DBT_TEST_OPTIONS'), 'Additional options to dbt test, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_EXCLUDE', allVars.get('DBT_TEST_EXCLUDE'), 'Additional options to dbt test, such as --exclude. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
```
pwnage101, line 28: Can you clarify in the description if the user is expected to provide the `--<option>` flag?  Or just a hint, like:
```groovy
                 stringParam('DBT_TEST_EXCLUDE', allVars.get('DBT_TEST_EXCLUDE'), 'Additional options to dbt test, such as "--exclude <model>". Details here: https://docs.getdbt.com/docs/model-selection-syntax')
```
Alternatively, I've always wondered why we require parameters with CLI arguments in them.  Can't we just accept them as model selectors, and add the flag in the bash script later:

```bash
# If DBT_TEST_EXCLUDE is provided, prepend "--exclude "
DBT_TEST_EXCLUDE="${DBT_TEST_EXCLUDE:+--exclude }$DBT_TEST_EXCLUDE"
```

pwnage101 (approved):
Just a nit about our bad pattern of requiring parameter values to contain boilerplate code.

dataeng/jobs/analytics/WarehouseTransformsCIManual.groovy
jazibhumayun: Yes we can prepend it, but I feel it should be common across all jobs or at-least manual jobs

jazibhumayun: Ok I just saw the dbt-manual job does not have DBT_TEST_OPTIONS or DBT_TEST_EXCLUDES parameters anymore :D
Question: What should I do now prepend the --exclude  or keep like as is ?

pwnage101: I think this would be a good opportunity to cleanup the other two jobs to be consistent:

* dataeng/jobs/analytics/DBTManual.groovy
* dataeng/jobs/analytics/WarehouseTransformsCI.groovy

I think the `DBT_*_OPTIONS` can left alone, only `DBT_*_EXCLUDE` needs to be updated.  However, if you do decide to update `DBT_*_OPTIONS`, you should rename it to `DBT_*_MODELS` because it will no longer serve the purpose of passing arbitrary options.

I wouldn't normally be this pedantic, but I think this improves the UX of the various jobs, and my understanding is that data ops is tackling some quality-of-life improvements.  One time, confusion around how to format parameter values in jenkins jobs has caused partner support many lost hours of passing incorrectly formatted values to the event export jobs, and that's just one example.

pwnage101 (commented):
I also want to emphasize that my previous comment about DBT_RUN_EXCLUDE is non-blocking.  If you need to move forward with this, feel free to merge and capture that recommendation in a separate ticket.

One thing that must be done still, however, is to add a body to your commit message.

jazibhumayun:
> I also want to emphasize that my previous comment about DBT_RUN_EXCLUDE is non-blocking (updating the description like I mentioned was my only requirement). If you need to move forward with this, feel free to merge and capture that recommendation in a separate ticket.
> 
> One thing that must be done still, however, is to add a body to your commit message.

@pwnage101 
I discussed about this PR in yesterday's scrum and decided to proceed with the current approach of parameters. I have created a following ticket to refactor this job, dbt-manual and WarehouseTransformCI jobs to have consistent approach being used.
[https://openedx.atlassian.net/browse/DENG-848](https://openedx.atlassian.net/browse/DENG-848)

pwnage101 (approved):

PR #1336, opened 2021-03-06
Merges to Master WT CI

jazibhumayun:
DENG-774
I have created two jobs
1- First job keeps polling after 10m interval warehouse-transform for any new merges to master.
2- Second job which is the actual worker job to run dbt CI, it is a down stream of first polling job
Why we need additional polling job? because the actual dbt CI job has multiple repositories to clone to get required files for snowflake python scripts, secrets and dsl so their master merges won't caught as changes we needed separate polling job

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
@@ -1,0 +1,143 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_authorization
+
+
+class WarehouseTransformsMasterCI{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("warehouse-transforms-master-ci"){
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('WAREHOUSE_TRANSFORMS_URL', allVars.get('WAREHOUSE_TRANSFORMS_URL'), 'URL for the warehouse-transforms repository.')
+                stringParam('WAREHOUSE_TRANSFORMS_BRANCH', allVars.get('WAREHOUSE_TRANSFORMS_BRANCH'), 'Branch of warehouse-transforms repository to use.')
+                stringParam('DBT_TARGET', allVars.get('DBT_TARGET'), 'DBT target from profiles.yml in analytics-secure.')
+                stringParam('DBT_PROFILE', allVars.get('DBT_PROFILE'), 'DBT profile from profiles.yml in analytics-secure.')
+                stringParam('DBT_RUN_OPTIONS', allVars.get('DBT_RUN_OPTIONS'), 'Additional options to dbt run, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_RUN_EXCLUDE', allVars.get('DBT_RUN_EXCLUDE'), 'Additional options to dbt run, such as --exclude. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_OPTIONS', allVars.get('DBT_TEST_OPTIONS'), 'Additional options to dbt test, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('DBT_TEST_EXCLUDE', allVars.get('DBT_TEST_EXCLUDE'), 'Additional options to dbt test, such as --exclude. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
+                stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), 'Branch of analytics tools repo to use.')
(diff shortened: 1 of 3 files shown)


dataeng/resources/warehouse-transforms-master-ci.sh
```diff
@@ -0,0 +1,132 @@
...
+
+# Setup to run python script to create snowflake schema
+cd $WORKSPACE/analytics-tools/snowflake
+make requirements
+
+# Setup to run dbt commands
+cd $WORKSPACE/warehouse-transforms
+
+COMMIT_ID=$(git rev-parse --short HEAD)
+# Only run CI on merge commits, otherwise exit
```
jazibhumayun, line 12: Running only CI on merge commits on master branch

dataeng/resources/warehouse-transforms-master-ci.sh
```diff
@@ -0,0 +1,132 @@
...
+else
+    echo "Exiting because not a merge commit"
+    exit 0
+fi 
+
+
+# Get second last merge commit id and compares it with the HEAD to find (git diff) files changed.
+PREV_MERGE_COMMIT_ID=$(git log --merges origin/master --format='%H' --max-count=2 | sed -n 2p)
+
+git diff $PREV_MERGE_COMMIT_ID --name-only
```
jazibhumayun, line 27: Comparing difference between the HEAD and second last merge commit to find out the projects files have changed.

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
```diff
@@ -0,0 +1,152 @@
...
+
+            }
+            triggers {
+                scm('H/10 * * * *')
+            }            
+
+            wrappers {
+                colorizeOutput('xterm')
+            }
+            wrappers common_wrappers(allVars)
```
estute, line 61: I think this is missing a `steps` section or something that will trigger the downstream job

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
```diff
@@ -0,0 +1,152 @@
...
+                stringParam('JENKINS_JOB_DSL_URL', allVars.get('JENKINS_JOB_DSL_URL'), 'URL for the jenkins-job-dsl repo.')
+                stringParam('JENKINS_JOB_DSL_BRANCH', allVars.get('JENKINS_JOB_DSL_BRANCH'), 'Branch of jenkins-job-dsl repo to use.')
+                stringParam('DB_NAME', allVars.get('DB_NAME'), 'Database name used to create output schema of dbt run/tests')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+            }
+            environmentVariables {
+                env('KEY_PATH', allVars.get('KEY_PATH'))
+                env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
+                env('USER', allVars.get('USER'))
+                env('ACCOUNT', allVars.get('ACCOUNT'))
```
estute, line 37: Why will this job need any of these credentials? It shouldn't interface with Snowflake, just determine if there is a change to the master branch, right?

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
```diff
@@ -0,0 +1,152 @@
...
+            }
+            scm {
+                 github('edx/warehouse-transforms')
+            }
```
estute, line 93: do you need this section?  I think it should be covered in the `multiscm` section

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
```diff
@@ -0,0 +1,152 @@
...
+            scm {
+                 github('edx/warehouse-transforms')
+            }                      
+            multiscm secure_scm(allVars) << {
+                git {
+                    remote {
+                        url('$WAREHOUSE_TRANSFORMS_URL')
+                        branch('$WAREHOUSE_TRANSFORMS_BRANCH')
+                        credentials('1')
```
estute, line 99: We really need to rename that credential. It can be a backlog ticket

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                    extensions {
+                        relativeTargetDirectory('jenkins-job-dsl')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }                              
+            }
+            triggers {
+                upstream('warehouse-transforms-master-ci', 'SUCCESS')
```
estute, line 133: thinking this through, it looks like the upstream job will run every 10 minutes, regardless of changes to master. So, as long as nothing goes wrong, this job will also run every 10 minutes. I feel like there is value to having this job only run when the changes are detected. What do you think about moving the bit to check if master changed into the upstream job?

jazibhumayun: Actually the above job polls every 10 minutes but only runs when it detects changes in the master. so, I think its fine like this.

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
jazibhumayun: correct, removing those

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
jazibhumayun: Yeah I nominated this job as upstream job of second job, its works fine.

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
jazibhumayun: Yes this is not needed here, removing it.

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
jazibhumayun: Ok, created one.

estute: thank you

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
estute: OK, cool. Is that because the trigger is `scm()` and not `cron()`?

estute (approved):

dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
jazibhumayun: yes true

PR #1346, opened 2021-03-16
Remove matching text before post-build step for analytics jobs.

doctoryes:
Use a regex to match *any* text so that the post-build step for analytics jobs will *always* run and disable the Opsgenie heartbeat.


brianhw (approved):

pwnage101 (approved):
Although more verbose, `postBuildScripts` is more semantically correct, i think, for doing something like this (run a shell script post build regardless of the console text).  However, this works and it fits on one line so :shrug:

PR #1349, opened 2021-03-25
Merges to master re-run logic

jazibhumayun:
Re-running upon failed tests, before notifying it as failure.


bmedx (approved):
Seems ok to me!

pwnage101 (commented):
If the first job fails and the second one succeeds, isn't the first one still going to trigger an alert?

pwnage101 (approved):
Oh, I see now that common_publishers has been removed from the first job.

pwnage101:
@estute @jazibhumayun this feels like it does a similar thing to https://github.com/edx/warehouse-transforms/pull/1955

My preference is to not have similar things done two completely different ways.  Have we considered a general solution?

jazibhumayun:
@pwnage101 
You are correct we no longer need this PR to get merged
edx/warehouse-transforms#1955 this brings a python script that will be consumed by existing Merges to Master job
Here is the PR for that will be consuming the re-run flaky test script: [https://github.com/edx/jenkins-job-dsl/pull/1353](https://github.com/edx/jenkins-job-dsl/pull/1353)

PR #1351, opened 2021-03-26
slack publisher to notify build failures

jazibhumayun:
We also want to notify data-support if Merges to Master CI job has failed, if this job fails it will likely break warehouse-transforms Prod jobs so this job works as canary test to identify the job failure ahead.

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
@@ -6,6 +6,7 @@ import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
 import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
 import static org.edx.jenkins.dsl.AnalyticsConstants.common_authorization
+import static org.edx.jenkins.dsl.AnalyticsConstants.slack_publisher
 
 
 class WarehouseTransformsMasterCI{
@@ -75,6 +76,8 @@ class WarehouseTransformsMasterCI{
                 stringParam('JENKINS_JOB_DSL_BRANCH', allVars.get('JENKINS_JOB_DSL_BRANCH'), 'Branch of jenkins-job-dsl repo to use.')
                 stringParam('DB_NAME', allVars.get('DB_NAME'), 'Database name used to create output schema of dbt run/tests')
                 stringParam('NOTIFY', allVars.get('NOTIFY'), 'Space separated list of emails to send notifications to.')
+                stringParam('SLACK_NOTIFICATION_CHANNEL', allVars.get('SLACK_NOTIFICATION_CHANNEL'), 'Space separated list of slack channel name to send build failure notifications')
+                stringParam('FAILURE_MESSAGE', allVars.get('FAILURE_MESSAGE'), 'Custom message for to send along with buid failure notification')
             }
             environmentVariables {
                 env('KEY_PATH', allVars.get('KEY_PATH'))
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy
```diff
@@ -216,5 +216,29 @@ This text may reference other parameters in the task as shell variables, e.g.  $
             }
         }
     }
+    public static def slack_publisher = { allVars ->
+        return {
+            it /
+                publishers /
+                'jenkins.plugins.slack.SlackNotifier' {
```
HassanJaveed84, line 223: Are we unable to use `slackNotifier` directly in the publisher context ?
http://jenkins-new.analytics.edx.org/plugin/job-dsl/api-viewer/index.html#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.slackNotifier
I think this might work ? unless already tried and it does not.
```
publishers {
    slackNotifier {
        room('$SLACK_NOTIFICATION_CHANNEL')
        ....
    }
}

Comment (jazibhumayun) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:223]: Ok let me try this way

Comment (jazibhumayun) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:223]: Yes that is working, thanks
So one suggestion needed should I put this code in AnalyticsConstants.groovy or inside the job DSL code like here in testing PR: https://github.com/edx/jenkins-job-dsl/pull/1358/files
@HassanJaveed84

Comment (HassanJaveed84) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:223]: If we expect to be using the notification in future for other jobs, we should put it in the constants file and make it re-usable.

Comment (jazibhumayun) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy:223]: Ok I will prefer putting it in AnalyticsConstants file because there are chances of using slack notifications in future jobs.

Comment (jazibhumayun) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy]: Hey! updated as per our discussion @HassanJaveed84

Comment (HassanJaveed84) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy]: I notice that you're passing `allVars` to the `slack_publisher` method. I'd change

`room('$SLACK_NOTIFICATION_CHANNEL')` -> `room(allVars.get('SLACK_NOTIFICATION_CHANNEL'))`
  `customMessageFailure('$FAILURE_MESSAGE')` -> `customMessageFailure(allVars.get('FAILURE_MESSAGE'))`
  
  and then we can remove the parameters defined at the top.

**Review by HassanJaveed84 — APPROVED** [review]
One nit otherwise LGTM

File: src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy (hunk 1 above)
Comment (jazibhumayun) [reply src/main/groovy/org/edx/jenkins/dsl/AnalyticsConstants.groovy]: sure, fixed it

File: dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy [hunk 2]
```diff
@@ -126,6 +129,7 @@ class WarehouseTransformsMasterCI{
                 upstream('warehouse-transforms-master-ci', 'SUCCESS')
             }
             publishers common_publishers(allVars)
+            publishers slack_publisher(allVars)
```
HassanJaveed84, line 132: Curious if this still works as the method signature doesn't match.

jazibhumayun: Ohh, it worked actually but I probably need to remove this parameter.

jazibhumayun: @HassanJaveed84 
here is the fixing PR: [https://github.com/edx/jenkins-job-dsl/pull/1359/files](https://github.com/edx/jenkins-job-dsl/pull/1359/files)

PR #1353, opened 2021-04-01
feat:re-run logic, test more than once incase of failure

jazibhumayun:
Re-running upon failed tests, before notifying it as failure.
Using ci_scripts coming from this PR: https://github.com/edx/warehouse-transforms/pull/1955

Please thumb this PR that contains parameters: https://github.com/edx-ops/analytics-secure/pull/551/files

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransformsCI.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsCI.groovy
@@ -37,6 +37,8 @@ class WarehouseTransformsCI{
                 env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
                 env('USER', allVars.get('USER'))
                 env('ACCOUNT', allVars.get('ACCOUNT'))
+                env('WITH_RETRY', allVars.get('WITH_RETRY'))
+                env('NO_OF_TRIES', allVars.get('NO_OF_TRIES'))
             }
             scm {
                  github('edx/warehouse-transforms')
--- a/dataeng/resources/warehouse-transforms-ci-dbt.sh
+++ b/dataeng/resources/warehouse-transforms-ci-dbt.sh
@@ -11,4 +11,26 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
 dbt seed --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
 
 dbt run $DBT_RUN_OPTIONS $DBT_RUN_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
-dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
+# Jenkins jobs are marked as failed when any of command fails so writing the following test command with && true so it will give a chance to
+# evaluate its success or failure base on success or failure we can do further re-tries on failed tests
+dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET && true
+if [ $? -eq 1 ]
+then
+    if [ "$WITH_RETRY" == "true" ]
+    then
+        pip install -r ../../tools/ci_scripts/requirements.txt
+        if [ "$DBT_TEST_EXCLUDE" == "" ]
+        then
+            python ../../tools/ci_scripts/rerun_flaky_tests.py --project-path . --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/  --profile $DBT_PROFILE  \
+            --target $DBT_TARGET --count $NO_OF_TRIES
+        else
+            PREFIX="--exclude "
+            TEST_EXCLUSIONS=$(echo "$DBT_TEST_EXCLUDE" | sed -e "s/^$PREFIX//")
+            python ../../tools/ci_scripts/rerun_flaky_tests.py --project-path . --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/  --profile $DBT_PROFILE  \
+            --target $DBT_TARGET --exclusions $TEST_EXCLUSIONS --count $NO_OF_TRIES
+        fi
+    else
+        echo "Tests failed but retry is not enabled"
+        exit 1
+    fi
+fi
--- a/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransformsMasterCI.groovy
@@ -81,6 +81,8 @@ class WarehouseTransformsMasterCI{
                 env('PASSPHRASE_PATH', allVars.get('PASSPHRASE_PATH'))
                 env('USER', allVars.get('USER'))
                 env('ACCOUNT', allVars.get('ACCOUNT'))
+                env('WITH_RETRY', allVars.get('WITH_RETRY'))
+                env('NO_OF_TRIES', allVars.get('NO_OF_TRIES'))
             }
             multiscm secure_scm(allVars) << {
                 git {


dataeng/jobs/analytics/WarehouseTransformsCI.groovy
```diff
@@ -30,6 +30,7 @@ class WarehouseTransformsCI{
                 stringParam('JENKINS_JOB_DSL_URL', allVars.get('JENKINS_JOB_DSL_URL'), 'URL for the jenkins-job-dsl repo.')
                 stringParam('JENKINS_JOB_DSL_BRANCH', allVars.get('JENKINS_JOB_DSL_BRANCH'), 'Branch of jenkins-job-dsl repo to use.')
                 stringParam('DB_NAME', allVars.get('DB_NAME'), 'Database name used to create output schema of dbt run/tests')
+                stringParam('WITH_RETRY', allVars.get('WITH_RETRY'), 'Set this to true to retry dbt tests')
```
jazibhumayun, line 33: Had to add this parameter with default value as 'false'

estute: This could also just be an environment variable. It might help avoid confusion, but who knows we may want to have this as our CI system comes under more usage.

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,17 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
 dbt seed --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
 
 dbt run $DBT_RUN_OPTIONS $DBT_RUN_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
-dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
\ No newline at end of file
+dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET && true
```
estute, line 14: I'm confused here. If the `set` built-in is still in place, then failing commands should cause the script to exit immediately. However, the next line has a check for a nonzero status code. I'm not sure this will work.

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,17 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
...
+if [ $? -eq 1 ]
+then
+    if [ "$WITH_RETRY" == "true" ]
+    then
+        pip install -r ../../tools/ci_scripts/requirements.txt
+        NO_OF_TRIES=3
+        python ../../tools/ci_scripts/rerun_flaky_tests.py --project-path . --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/  --profile $DBT_PROFILE  \
+        --target $DBT_TARGET --exclusions "$DBT_TEST_EXCLUDE" --count $NO_OF_TRIES
```
estute, line 22: I had to change how the rerun_flaky_tests.py script consumes exclusions on the CLI. Rather than pass it the $DBT_TEST_EXCLUDE variable, can you pass it only the actual exclusions.

jazibhumayun: Ok you mean I should write --exclusions "test_name:relationships" instead of using $DBT_TEST_EXCLUDE variable.

jazibhumayun: In-fact I realized we are not using --exclusions "test_name:relationships" in m2m job so I don't need to pass it here.

dataeng/resources/warehouse-transforms-ci-dbt.sh
jazibhumayun: It works fine as per my testing, we are good on that.

dataeng/jobs/analytics/WarehouseTransformsCI.groovy
jazibhumayun: So in case of environment variable where should I define it ? Will it go to shell script or somewhere in DSL ?
If we can define in DSL then it is fine, defining inside shell script will be a problem because regular WT CI job uses the same shell script and we are not re-trying flaky tests  there

jazibhumayun: @estute Sorry can you explain it a bit more

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,17 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
...
\ No newline at end of file
+dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET && true
+if [ $? -eq 1 ]
+then
+    if [ "$WITH_RETRY" == "true" ]
+    then
+        pip install -r ../../tools/ci_scripts/requirements.txt
+        NO_OF_TRIES=3
+        python ../../tools/ci_scripts/rerun_flaky_tests.py --project-path . --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/  --profile $DBT_PROFILE  \
+        --target $DBT_TARGET --count $NO_OF_TRIES
```
estute, line 22: ^ I think you will want to include the exclusions in the invocation of this script. The script has an optional flag `--exclusions` that should consume the same exclusions as the original test run

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,17 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
...
 dbt run $DBT_RUN_OPTIONS $DBT_RUN_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
-dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
\ No newline at end of file
+dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET && true
+if [ $? -eq 1 ]
+then
+    if [ "$WITH_RETRY" == "true" ]
+    then
+        pip install -r ../../tools/ci_scripts/requirements.txt
+        NO_OF_TRIES=3
```
estute, line 20: nit: I feel like it might be better to keep all of the variables together in the `parameters` block in the DSL, so that we don't have to dig in the scripts

dataeng/jobs/analytics/WarehouseTransformsCI.groovy
estute: sure thing. You can set environment variables in an `env` block, like this: https://github.com/edx/jenkins-job-dsl/blob/95ae617e7e35b4e0e76ca535afa3d96af07f4abe/devops/jobs/RunAnsible.groovy. That way, it's not a user modifiable parameter. I suppose this doesn't matter at the end of the day, because this job will not be triggered directly by users, but rather as part of an automated flow.

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,24 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
 dbt seed --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
 
 dbt run $DBT_RUN_OPTIONS $DBT_RUN_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
-dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
\ No newline at end of file
+dbt test $DBT_TEST_OPTIONS $DBT_TEST_EXCLUDE --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET && true
```
estute, line 14: Might be a good idea to add a comment as to why we are using the `&& true`

dataeng/resources/warehouse-transforms-ci-dbt.sh
```diff
@@ -11,4 +11,24 @@ dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --prof
...
+    if [ "$WITH_RETRY" == "true" ]
+    then
+        pip install -r ../../tools/ci_scripts/requirements.txt
+        if [ "$DBT_TEST_EXCLUDE" == "" ]
+        then
+            python ../../tools/ci_scripts/rerun_flaky_tests.py --project-path . --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/  --profile $DBT_PROFILE  \
+            --target $DBT_TARGET --count $NO_OF_TRIES
+        else
+            PREFIX="--exclude "
+            TEST_EXCLUSIONS=$(echo "$DBT_TEST_EXCLUDE" | sed -e "s/^$PREFIX//")
```
estute, line 26: nit: I feel like this is more complicated than needed. In general, I prefer explicit values over manipulated ones. I understand why you did it this way, but I just wonder if there is a simpler way of passing this value. Perhaps something like:
```
DBT_EXCLUSIONS="x,y,z"
DBT_EXCLUSION_STRING="--exclude $DBT_EXCLUSIONS"
```
that way, you have both formats. Maybe we just need to change the name of the parameter in the python script

At the end of the day, it's not really that important, and you should be able to merge this, but if you have a spare moment, I feel like we can make this more clear.

estute (approved):

PR #1362, opened 2021-04-09
fix: add full-refresh flag to dbt seed commands in wt jobs

estute:


pwnage101 (approved):

brianhw (approved):
I think that's okay to do all the time, but obviously only on the seed job.  (On run, we would lose all incremental history.)

pwnage101 (approved):

PR #1364, opened 2021-04-14
feat: Turn off Who-Tests-What check until we're ready to work on it again

timmc-edx:
This check has been running on a daily basis, but the deployment pipeline has been waiting for it to finish (in `check_pr_tests_status.py`) meaning that it sometimes blocks deployment in the mornings as the check is still in the "pending" state.

We should either delete this check or exclude it from the status checks, but for now I'm just disabling the daily run on master so that we can defer that decision.

(I would have used `disabled()` but I don't know if it apples to `pipelineJob`.)

**Note:** Who-Tests-What (WTW) is a project to collect coverage information from tests so that we know which parts of the codebase are covered by which tests. This may be owned by @nedbat.

### DIFF
--- a/platform/jobs/pipelines/edxPlatformPipelineMasterWTWNightly.groovy
+++ b/platform/jobs/pipelines/edxPlatformPipelineMasterWTWNightly.groovy
@@ -33,9 +33,13 @@ jobConfigs.each { jobConfig ->
                 PYTHON_VERSION: "${jobConfig.pythonVersion}"
             )
 
-            triggers {
-                cron('@daily')
-            }
+            // Turn off job until we're ready to work on it again; at
+            // that time make sure that it's not a blocking check for
+            // deploys if this gets reenabled. (Needs to be excluded
+            // from check_pr_tests_status.py call.)
+            // triggers {
+            //     cron('@daily')
+            // }
 
             cpsScm {
                 scm {


platform/jobs/pipelines/edxPlatformPipelineMasterWTWNightly.groovy
```diff
@@ -33,9 +33,13 @@ jobConfigs.each { jobConfig ->
                 PYTHON_VERSION: "${jobConfig.pythonVersion}"
             )
 
-            triggers {
-                cron('@daily')
-            }
+            // Turn off job until we're ready to work on it again; at
```
robrap, line 36: 1. "We should either delete this check or exclude it from the status checks, but for now I'm just disabling the daily run on master so that we can defer that decision." feels like useful context to add here or to a ticket.
2. I'm not clear on the downside of deleting this altogether. If we want it - I feel like we should ticket it and reference the ticket from the comment. The ticket can include additional comments/thoughts about fixing or permanently deleting.

timmc-edx: Downside is messing up git blame history. My plan here is to disable for now to get it out of the way and then notify the owner; they can either delete or fix.

robrap: I was assuming we were the owner because we own edx-platform.  I don't actually understand what the "who-tests-what" check is. :)

timmc-edx: I think Jeremy said that Ned would be the person to talk to, and that Cale had been leading this.

timmc-edx: WTW is a project to collect coverage information from tests so that we know which parts of the codebase are covered by which tests. I think. Something like that.

robrap (approved):
@timmc-edx Based on the conversation, I think you can merge to fix, and have @nedbat follow-up then by updating the comment, or deleting the code.  Thanks!

platform/jobs/pipelines/edxPlatformPipelineMasterWTWNightly.groovy
nedbat: WTW would collect data about which tests ran each line of product code.  It's expensive, and never got to a working state.  I don't know who might pick it up next, but it probably will not be for a long time, if ever.  I would support deleting this.  There's also a dependency in edx-platform that we could remove:
```
https://github.com/nedbat/coverage_pytest_plugin.git
```

robrap: Is there a Confluence page or anything where we could point to the clean-up PRs?

nedbat: I'm not finding any relevant wiki pages.

nedbat:
BTW, how was this a "feat" commit?  Turning off a CI job that no one was using would be "build", I would think.

timmc-edx:
One repo's `feat` is another repo's `build`?

If this were a change to a CI config in edx-platform I'd call it a `build` commit, but this repo is entirely about builds and tests, so I think of it as a feature change for the automation itself.

nedbat:
It's an interesting question: can we assume that the commit messages are read in the context of the repo?

timmc-edx:
I don't think that commit messages are particularly useful without that context. (Even in that context, it can be hard to understand without the names of the changed files as well.)

PR #1366, opened 2021-04-18
Use rename strategy for EnrollmentByBirthYearToMysqlTask.

HassanJaveed84:


pwnage101:
Can you change your commit message to be compliant with conventional commits?  In particular, I'm curious if this is a "fix" and what the reason for the change is.

PR #1367, opened 2021-04-27
REV-2184: add frontend-app-learning for sandboxes

dianekaplan:
For REV-2184: adding frontend-app-learning for sandboxes, following Step 2 from here:
https://openedx.atlassian.net/wiki/spaces/FEDX/pages/2124677169/How+To+Add+an+MFE+to+a+Sandbox

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -225,6 +225,9 @@ class CreateSandbox {
                 booleanParam("payment",false,"Enable Payment MFE")
                 stringParam("payment_version","master","The repository version of the frontend-app-payment")
 
+                booleanParam("learning",false,"Enable Learning MFE")
+                stringParam("learning_version","master","The branch of the frontend-app-learning repository")
+
                 choiceParam("server_type",
                             [
                             "full_edx_installation_from_scratch",


devops/jobs/CreateSandbox.groovy
```diff
@@ -225,6 +225,9 @@ class CreateSandbox {
                 booleanParam("payment",false,"Enable Payment MFE")
                 stringParam("payment_version","master","The repository version of the frontend-app-payment")
 
+                booleanParam("learning",false,"Enable Learning MFE")
```
adzuci, line 228: I think this can also accept a commitish, do you know if that's the case?  If not we should probably update the help text and name of all of these version parameters.

adzuci (approved):

devops/jobs/CreateSandbox.groovy
dianekaplan: Thanks, Adam!  I'm not sure- the instructions on the setup page (linked in description) say this change will give the option to toggle the learning MFE on/off in the CreateSandbox job configuration UI, and then I think it's the [changes in the Config repo](https://github.com/edx/configuration/pull/6388/files) that will allow a user to specify a branch version, but this is my first time through so I could be wrong.

PR #1378, opened 2021-05-18
feat: CD Prefect flows Jenkins job

jazibhumayun:
A Jenkins job will poll master branch of prefect-flows repository. Upon new merge commits it will detects file changes or addition of new flow and passes those files in downstream which then calls the respective job. Deployment script install requirements, do the required authentication and make deploy flow

### DIFF
--- a/dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
+++ b/dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
@@ -1,0 +1,141 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_authorization
+import static org.edx.jenkins.dsl.AnalyticsConstants.slack_publisher
+
+
+class PrefectFlowsDeployment{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("prefect-flows-deployment-poll"){
+            description(
+                'This job polls prefect-flows repository code changes to trigger deployment ' +
+                'It detects file changes for latest merge commit, creates a downstream.properties ' +
+                ' file to pass comma delimeted list of flows in FLOWS_TO_DEPLOY parameter to downstream job'
+            )
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('PREFECT_FLOWS_URL', allVars.get('PREFECT_FLOWS_URL'), 'URL for the prefect-flows repository.')
+                stringParam('PREFECT_FLOWS_BRANCH', allVars.get('PREFECT_FLOWS_BRANCH'), 'Branch of prefect-flows repository to use.')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+            }
+            scm {
+                // As it is a poll job it should only use git scm for prefect-flows code repository
+                // Do not add anyother git scm in this poll job
+                git {
+                    remote {
+                        url('$PREFECT_FLOWS_URL')
+                        branch('$PREFECT_FLOWS_BRANCH')
+                    }
+                    extensions {
+                        relativeTargetDirectory('prefect-flows')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+            }
+            publishers common_publishers(allVars)
+            triggers common_triggers(allVars, allVars)
+            wrappers common_wrappers(allVars)
+            wrappers {
+                colorizeOutput('xterm')
+                timestamps()
+                credentialsBinding {
+                    usernamePassword('GITHUB_USER', 'GITHUB_TOKEN', 'GITHUB_USER_PASS_COMBO');
+                }
+            }
+            steps {
+                shell(dslFactory.readFileFromWorkspace('dataeng/resources/prefect-flows-deployment-identify.sh'))
+            }
+            publishers {
+                downstreamParameterized {
+                    trigger('prefect-flows-deployment-intermediate') {
+                        condition('SUCCESS')
+                        parameters {
+                            // The contents of this file are generated as part of the shell script in build step.
+                            propertiesFile('${WORKSPACE}/downstream.properties')
+                        }
+                    }
+                }
+            }
+        }
+        dslFactory.job("prefect-flows-deployment-intermediate"){
+            description(
+                'It receives comma delimated list of FLOWS_TO_DEPLOY and use this parameter to trigger downstream ' +
+                'Environment variables created inside shell script can not be referred back in same DSL again ' +
+                'This job works as intermediate to get FLOWS_TO_DEPLOY and use it to call relevant downstream deployment jobs'
+            )
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('FLOWS_TO_DEPLOY', allVars.get('FLOWS_TO_DEPLOY'), 'Comma separated list of Flows that upstream job has identified and needs to redeploy')
+            }
+            wrappers common_wrappers(allVars)
+            publishers {
+                downstreamParameterized {
+                    trigger('$FLOWS_TO_DEPLOY') {
+                        condition('SUCCESS')
+                        parameters {
+                            // Added parameter to pass name flows to be deployed
+                            predefinedProp('FLOWS_TO_DEPLOY', '$FLOWS_TO_DEPLOY')
+                        }
+                    }
+                }
+            }
+        }
+        allVars.get('PREFECT_FLOWS').each { prefect_flows ->
+            dslFactory.job("prefect-flows-deployment-$prefect_flows"){
+                description(
+                    'Changes detected in upstream job gets trigger this prefect-flows-deployment job ' +
+                    'It installs requirements, autheticates prefect and deploy latest code changes'
+                )
+                // if name of this job is ever changed. Make to sure to update deletion of job name prefix in prefect-deployment-identify.sh
+                authorization common_authorization(allVars)
+                logRotator common_log_rotator(allVars)
+                parameters secure_scm_parameters(allVars)
+                parameters {
+                    stringParam('PREFECT_FLOWS_URL', allVars.get('PREFECT_FLOWS_URL'), 'URL for the prefect-flows repository.')
+                    stringParam('PREFECT_FLOWS_BRANCH', allVars.get('PREFECT_FLOWS_BRANCH'), 'Branch of prefect-flows repository to use.')
+                    stringParam('ECR_LOGIN', allVars.get('ECR_LOGIN'), 'ECR repository URI used to login to ECR')
+                    stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                }
+                environmentVariables {
+                    env('PREFECT_VAULT_KV_PATH', allVars.get('PREFECT_VAULT_KV_PATH'))
+                    env('PREFECT_VAULT_KV_VERSION', allVars.get('PREFECT_VAULT_KV_VERSION'))
+                }
+                multiscm secure_scm(allVars) << {
+                    git {
+                        remote {
+                            url('$PREFECT_FLOWS_URL')
+                            branch('$PREFECT_FLOWS_BRANCH')
+                        }
+                        extensions {
+                            relativeTargetDirectory('prefect-flows')
+                            pruneBranches()
+                            cleanAfterCheckout()
+                        }
+                    }
+                }
+                publishers common_publishers(allVars)
+                wrappers {
+                    colorizeOutput('xterm')
+                    timestamps()
+                    credentialsBinding {
+                        usernamePassword('ANALYTICS_VAULT_ROLE_ID', 'ANALYTICS_VAULT_SECRET_ID', 'analytics-vault');
+                        usernamePassword('GITHUB_USER', 'GITHUB_TOKEN', 'GITHUB_USER_PASS_COMBO');
+                    }
+                }
+                steps {
+                    shell(dslFactory.readFileFromWorkspace('dataeng/resources/prefect-flows-deployment.sh'))
+                }
+            }
+        }
+    }
+}
--- a/dataeng/jobs/createJobsNew.groovy
+++ b/dataeng/jobs/createJobsNew.groovy
@@ -7,6 +7,7 @@ import static analytics.ModelTransfers.job as ModelTransfersJob
 import static analytics.SnowflakeCollectMetrics.job as SnowflakeCollectMetricsJob
 import static analytics.SnowflakeExpirePasswords.job as SnowflakeExpirePasswordsJob
 import static analytics.SnowflakeSchemaBuilder.job as SnowflakeSchemaBuilderJob
+import static analytics.PrefectFlowsDeployment.job as PrefectFlowsDeploymentJob
 import static analytics.TerminateCluster.job as TerminateClusterJob
 import static analytics.UpdateUsers.job as UpdateUsersJob
 import static analytics.WarehouseTransforms.job as WarehouseTransformsJob
@@ -44,6 +45,7 @@ def taskMap = [
     SNOWFLAKE_COLLECT_METRICS_JOB: SnowflakeCollectMetricsJob,
     SNOWFLAKE_EXPIRE_PASSWORDS_JOB: SnowflakeExpirePasswordsJob,
     SNOWFLAKE_SCHEMA_BUILDER_JOB: SnowflakeSchemaBuilderJob,
+    PREFECT_FLOWS_DEPLOYMENT_JOB: PrefectFlowsDeploymentJob,
     TERMINATE_CLUSTER_JOB: TerminateClusterJob,
     UPDATE_USERS_JOB: UpdateUsersJob,
     WAREHOUSE_TRANSFORMS_CI_JOB: WarehouseTransformsCIJob,
@@ -149,6 +151,13 @@ listView('dbt') {
     jobs {
         name('snowflake-schema-builder')
         regex('dbt-.*|warehouse-transforms-.*')
+    }
+    columns DEFAULT_VIEW.call()
+}
+listView('prefect') {
+    description('Jobs that are used for prefect flows deployment')
+    jobs {
+        regex('prefect-flows-.*')
     }
     columns DEFAULT_VIEW.call()
 }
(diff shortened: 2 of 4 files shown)


dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
...
+####### uncomment the following code after deploying all existing flows first time #######
+# if [ "$BUILD_NUMBER" -eq 1 ]
+# then
+#     echo "Creating repository as its a first build"
+#     aws ecr create-repository --repository-name $FLOW --region us-east-1
+# else
+#     echo "Not creating repository as its not a first build"
+# fi
+####### uncomment the above code after deploying all existing flows first time #######
+
```
jazibhumayun, line 36: As per the comments above these lines are commented out for now but will be commented after all flows are run for the first time

dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
...
+source "${PLATFORM_VENV}/bin/activate"
+
+# Removing prefix 'prefect-flows-deployment-'
+# $FLOW will contain the name of flow
+FLOW=$(echo $JOB_NAME | cut -c 26-)
+
+cd $WORKSPACE/edx-prefectutils
+pip install -r requirements/base.txt
+pip install -r requirements/base.in
+
```
jazibhumayun, line 16: `install -r requirements/base.txt`
`pip install -r requirements/base.in`

I am installing these two as a requirements here but there might be a better way do it

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+                        relativeTargetDirectory('prefect-flows')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+
+            }
+            triggers {
+                scm('H/2 * * * *') // change it to 10 minutes after testing
+            }
```
jazibhumayun, line 41: For few days of launch let it poll 2 minutes to help in testing quickly then change it to 10 minutes or 15 minutes

dataeng/resources/prefect-flows-deployment-identify.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+set -ex
+
+
+# Setup to run dbt commands
+cd $WORKSPACE/prefect-flows
+
+COMMIT_ID=$(git rev-parse --short HEAD)
```
estute, line 8: ^ I don't think you use this variable in the script

dataeng/resources/prefect-flows-deployment-identify.sh
```diff
@@ -0,0 +1,59 @@
...
+do
+  # Taking the immediate parent name e.g.if flows/load_google_sheets_to_snowflake/load_google_sheets_to_snowflake.py is the modified file it will extract the immediate
+  # parent directory name which is load_google_sheets_to_snowflake
+  parentname="$(basename "$(dirname "$file")")"
+  temp="$(echo -n "${parentname} ")"
+  list=$list$temp
+done
+
+# If both .toml and .py file of a flow has changed those immediate parent directory entries will be add twice so removing the duplicates
+duplicatesRemoved=$(echo "$list" | awk '{for (i=1;i<=NF;i++) if (!a[$i]++) printf("%s%s",$i,FS)}{printf("\n")}')
```
estute, line 46: I think there is a simpler way of doing this. Could you do something like:
```
git diff $PREV_MERGE_COMMIT_ID --name-only | grep -E ".*\.(py|toml)" |cut -d'/' -f2 |sort -u
```

dataeng/resources/prefect-flows-deployment-identify.sh
```diff
@@ -0,0 +1,59 @@
...
+
+echo $duplicatesRemoved # Priting for debugging purpose
+
+# Loop will parse all files and append it with 'FLOW_NAME=' to pass this parameter in downstream job
+files=($duplicatesRemoved)
```
estute, line 51: In this case, it's not so much files, as flows, right?

dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+set -ex
+
+# Creating python3.8 virtual env
+PLATFORM_VENV="platform_venv"
```
estute, line 5: Is this copy/paste? We aren't using the edx-platform in this job

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
estute: Makes sense. In general, we put cron statements in to the JOB_CONFIGS, so maybe we should do the same here.

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+            wrappers {
+                colorizeOutput('xterm')
+            }
+            wrappers common_wrappers(allVars)
+            steps {
+                virtualenv {
+                    pythonName('PYTHON_3.7')
```
estute, line 48: Why are we using python3.7 here and 3.8 later in the PR?

estute: actually, it looks like the identify script doesn't use Python, so maybe you can use a `shell` section, instead of `virtualenv`.

dataeng/resources/prefect-flows-deployment.sh
jazibhumayun: So I modified the command like 
```
# The following statement will help us to determine if repository already exists otherwise it will create a new repository with the name of flow
aws ecr describe-repositories --repository-names $FLOW --region us-east-1 || aws ecr create-repository --repository-name $FLOW --region us-east-1

```
where $FLOW is the name used with .py (e.g. `load-paypal-cases-to-snowflake-flow`) file and used in make command. The above command works fine but I realized our existing ecr repository name are not same of .py file rather they have - hyphens (e.g. `load-paypal-cases-to-snowflake-flow`) not sure why use this naming convention so this mismatch in the names is causing issue. I have also seen some repository name breaking this naming convention and have some different words in their name (e.g. `load-course-structure-to-snowflake-prefect-flow`)

jazibhumayun: So when I am automating this this job script detects any toml or .py changes to either re-deploy (if its a change in existing flow) or create new repository and then deployment (if new flow added). My suggestion is to change the names of repository and be consistent with ecr repository names same as of .py file name, but I am looking for suggestions on this point

estute: I agree that standardization would be a good idea, but would renaming the repository be a problem in Prefect? Can it be done as a simple swap-in-place? Also, does repository name always equal flow name? I know that it does not have to be the same as the Python filename.

jazibhumayun: Yes that is correct it doesn't have to be same as python filename, let me try to explain why I am suggesting it should be same as python file name.
Case 1: PR merged is a  New Flow added
- Suppose a person writing prefect-flow and name Docker storage image as different from its python file name, as I can see such naming in many of the flows currently [flow1](https://github.com/edx/prefect-flows/blob/master/flows/load_affiliate_window_transactions_to_snowflake/load_affiliate_window_transactions_to_snowflake.py#L103), [flow2](https://github.com/edx/prefect-flows/blob/master/flows/load_course_structure_to_snowflake/load_course_structure_to_snowflake.py#L612) and either we use decide to hyphens or underscore in the names if flow writer has changed the name we won't be able to check if the repository exits or not so that is why I feel standardization is important here.
Case 2: PR merged is a Existing Flow added
In this case will check the existence of repository on the basis of some naming convention hence some are already failing the naming convention so we will end up creating a new repository whereas their code will be pointing to old out updated ecr

jazibhumayun: And if we have to standardize it think using the `_ underscore` to make is exactly like python file will simplify it otherwise it will be overhead to swap underscores with hyphens
@estute

dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
...
+virtualenv --python=python3.8 --clear "${PLATFORM_VENV}"
+source "${PLATFORM_VENV}/bin/activate"
+
+# Removing prefix 'prefect-flows-deployment-'
+# $FLOW will contain the name of flow
+FLOW=$(echo $JOB_NAME | cut -c 26-)
+
+cd $WORKSPACE/edx-prefectutils
```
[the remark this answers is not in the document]
jazibhumayun: Well not sure why but it gives me opaque-keys error. Although that is added in [base.in#L5](https://github.com/edx/edx-prefectutils/blob/master/requirements/base.in#L5) and [base.txt](https://github.com/edx/edx-prefectutils/blob/master/requirements/base.txt#L75)

dataeng/resources/prefect-flows-deployment.sh
jazibhumayun: I thought `base.in` should be installed, I am installing `base.in` in setup.py file as well, should it be changed to `base.txt` 
[setup.py#L92](https://github.com/edx/edx-prefectutils/blob/master/setup.py#L92)

dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
...
+# then
+#     echo "Creating repository as its a first build"
+#     aws ecr create-repository --repository-name $FLOW --region us-east-1
+# else
+#     echo "Not creating repository as its not a first build"
+# fi
+####### uncomment the above code after deploying all existing flows first time #######
+
+# Install prefect in jenkins python3.8 venv
+pip install prefect
```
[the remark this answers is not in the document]
jazibhumayun: Yup correct agreed, I realized we have `prefect[aws,google,snowflake,viz]>=0.14.5` in prefect-flows requirements https://github.com/edx/prefect-flows/blob/master/requirements.txt#L6 I think we don't need to install prefect in this script

dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -0,0 +1,44 @@
...
+set -ex
+
+# Creating python3.8 virtual env
+PLATFORM_VENV="platform_venv"
+virtualenv --python=python3.8 --clear "${PLATFORM_VENV}"
+source "${PLATFORM_VENV}/bin/activate"
+
+# Removing prefix 'prefect-flows-deployment-'
+# $FLOW will contain the name of flow
+FLOW=$(echo $JOB_NAME | cut -c 26-)
```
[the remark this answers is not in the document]
jazibhumayun: No, FLOW_NAME is actually the full name of a jenkins job e.g. [prefect-flows-deployment-load_enterprise_tables_from_s3_to_aurora](http://jenkins-new.analytics.edx.org/job/prefect-flows-deployment-load_enterprise_tables_from_s3_to_aurora/)
whereas in make command it requires us to specify only the python file name. I can not shorten $FLOW_NAME environment variable as it contains the name of jenkins job to trigger it. That is the reason I am cutting the starting words

dataeng/resources/prefect-flows-deployment.sh
jazibhumayun: yeah it was typo

dataeng/resources/prefect-flows-deployment-identify.sh
jazibhumayun: Yes, if one PR has 3 file changes it will create a comma separated list of those 3 flows

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+                            // Added parameter to pass name flows to be deployed
+                            predefinedProp('FLOW_NAME', '$FLOW_NAME')
+                        }
+                    }
+                }
+            }
+        }
+
+        allVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("prefect-flows-deployment-$environment"){
```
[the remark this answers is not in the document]
jazibhumayun: I have renamed it to `PREFECT_FLOWS` instead of `ENVIRONMENTS`, so yes this is a list of `PREFECT_FLOWS` in order to create downstream job per flow

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+
+            }
+            wrappers {
+                colorizeOutput('xterm')
+            }
+            wrappers common_wrappers(allVars)
```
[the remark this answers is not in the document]
jazibhumayun: Initially I also tried to solve this with any intermediate job but due to a limitation of environment variable being populated in shell script can not be used in same job `downstreamStream parameterized` parameter section so I had to create an intermediate job, I have also updated its description

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
jazibhumayun: Yeah no need to use python_venv simply using shell now

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("prefect-flows-deployment-poll"){
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('PREFECT_FLOWS_URL', allVars.get('PREFECT_FLOWS_URL'), 'URL for the prefect-flows repository.')
+                stringParam('PREFECT_FLOWS_BRANCH', allVars.get('PREFECT_FLOWS_BRANCH'), 'Branch of prefect-flows repository to use.')
+                stringParam('JENKINS_JOB_DSL_URL', allVars.get('JENKINS_JOB_DSL_URL'), 'URL for the jenkins-job-dsl repo.')
+                stringParam('JENKINS_JOB_DSL_BRANCH', allVars.get('JENKINS_JOB_DSL_BRANCH'), 'Branch of jenkins-job-dsl repo to use.')
```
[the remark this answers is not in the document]
jazibhumayun: No use, removed

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
```diff
@@ -0,0 +1,162 @@
...
+
+
+class PrefectFlowsDeployment{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("prefect-flows-deployment-poll"){
+            authorization common_authorization(allVars)
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            parameters {
+                stringParam('PREFECT_FLOWS_URL', allVars.get('PREFECT_FLOWS_URL'), 'URL for the prefect-flows repository.')
```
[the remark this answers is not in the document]
jazibhumayun: No, I think I can remove this param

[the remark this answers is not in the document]
jazibhumayun: We don't intend to use other than master when this will be productionize but as of starting weeks I think we should this parameter for debugging purpose

dataeng/resources/prefect-flows-deployment.sh
estute: Everything in `base.in` is contained in `base.txt`. The difference is that `base.txt` has the pinned versions for these requirements, as well as all dependencies. Running `pip install` for both is redundant.

dataeng/resources/prefect-flows-deployment-identify.sh
estute: The name of this variable is what is confusing to me.

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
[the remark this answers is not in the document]
estute: I'm sorry, I am still a little confused. Two main questions here:
- why does this job have an `scm` section? It doesn't touch any code. From your description it seems like it just passes a variable to another job.
- I still don't understand why we need this intermediate job. As far as I can tell, the initial polling job creates a property file with one item in it, a comma-separated list (i.e. `FLOW_NAME: flow1, flow2, ...`). Then it triggers one build instance of the intermediate job to fan out this `FLOW_NAME` variable into multiple jobs. I think you can simplify it by creating multiple property files, each with a simple configuration like `FLOW_NAME: flow1`. Then, you can run a downstream job for each property file. Check out this example: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/AnalyticsExporter.groovy#L245

From the DSL docs: ```// Looks for files that match the specified pattern in the current build, then for each of them trigger a build of the specified project(s) by passing that file as a file parameter.
forMatchingFiles(String filePattern, String parameterName, String noFilesFoundAction = 'SKIP')```

dataeng/resources/prefect-flows-deployment-identify.sh
jazibhumayun: Updated it with the way you commented

dataeng/resources/prefect-flows-deployment-identify.sh
jazibhumayun: updated it

dataeng/resources/prefect-flows-deployment.sh
jazibhumayun: Here I am renaming the the flow names:
[https://github.com/edx/prefect-flows/pull/76](https://github.com/edx/prefect-flows/pull/76)

dataeng/jobs/analytics/PrefectFlowsDeployment.groovy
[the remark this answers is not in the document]
jazibhumayun: As per our discussion in meeting and my search on injectenv which helps to inject environment var in dsl and use in Build and Post Build execute shell script but not available in 
```
downstreamParameterized {
                    trigger('$ENV_VAR') { 
```
trigger block of DSL, I feel we can merge with the current implementation and will see if it needs refactoring like Analytics exporter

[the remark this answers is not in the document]
estute: Agreed. Thanks for looking into this.

dataeng/resources/prefect-flows-deployment.sh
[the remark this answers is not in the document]
estute: I still don't understand why we need to do this. edx-prefectutils is a dependency of prefect-flows. If we install the dependencies of prefect-flows, it should include edx-prefectutils (and its dependencies). https://github.com/edx/prefect-flows/blob/master/requirements.txt#L7

dataeng/resources/prefect-flows-deployment.sh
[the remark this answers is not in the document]
estute: Maybe changing the names of the variables would help make it more clear. For example:
- FLOW_NAME -> FLOW_NAMES or FLOWS_TO_DEPLOY
- FLOW -> FLOW_NAME

dataeng/jobs/createJobsNew.groovy
```diff
@@ -152,3 +154,10 @@ listView('dbt') {
     }
     columns DEFAULT_VIEW.call()
 }
+listView('prefect') {
```
estute, line 157: ah, good call on creating a new view. Thanks!

estute (approved):

PR #1379, opened 2021-05-18
fix: Capture dbt artifacts from failed runs

bmedx:
Currently the shell script will exit immediately if a dbt command fails, causing the artifacts to be lost. These changes should allow us to upload the artifacts and then exit if there is a failure, as well as clean up some boilerplate code.

### DIFF
--- a/dataeng/resources/warehouse-transforms.sh
+++ b/dataeng/resources/warehouse-transforms.sh
@@ -8,19 +8,15 @@ pip install -r requirements.txt
 
 cd $WORKSPACE/warehouse-transforms/projects/$DBT_PROJECT
 
-# Turn off automatic failure of this script if the command returns non-0
-set +e
-
 # Fails the job if a dbt command fails and uploads the dbt artifacts to Snowflake if the job is configured for it
+# First argument is the dbt operation name, second is the result code from the dbt command
 function postCommandChecks {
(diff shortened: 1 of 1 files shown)


dataeng/resources/warehouse-transforms.sh
```diff
@@ -6,41 +6,54 @@ cd $WORKSPACE/warehouse-transforms
...
-dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
+# Turn off automatic failure of this script if the command returns non-0
+set +e
 
-if [ "$SKIP_SEED" != 'true' ]
-then
-  dbt seed --full-refresh --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
-  if [ "$PUSH_ARTIFACTS_TO_SNOWFLAKE" = 'true' ]
+# Fails the job if a dbt command fails and uploads the dbt artifacts to Snowflake if the job is configured for it
+function postCommandChecks {
```
pwnage101, line 15: Nit: in addition to the echo statement, a comment explaining the function arguments would be good.

pwnage101 (approved):

pwnage101 (approved):

PR #1383, opened 2021-05-21
fix: move pipeline acceptance job to old de jenkins

estute:
For some reason, when I tried to get this job working on the new jenkins, it was able to create a virtualenv using the `python2.7` alias, but it is running into the following issue on the older box: 
```
/var/lib/jenkins/workspace/edx-analytics-pipeline-acceptance-test-manual/build/venvs/analytics-exporter/bin/python2.7: Too many levels of symbolic links
```
I fixed this by referring to the python2.7 binary instead of an alias.


pwnage101:
Can you comment (in the commit message too) on why you're moving it between jenkinses?

pwnage101 (approved):

PR #1386, opened 2021-05-25
fix: env interpolation in run-app-permissions.sh

kdmccormick:
With single quotes, '${ENVIRONMENT}-${DEPLOYMENT}'
is passed in as a literal string. Use double quotes
so that it is interpolated to, eg, 'prod-edx'.

Part of https://openedx.atlassian.net/browse/TNL-8274

Fixes another bug introduced in #1376


doctoryes (approved):
It's so hard to get bash quoting correct the first time...

PR #1394, opened 2021-06-18
feat: Added cluster parameters to specify on-demand capacity.

HassanJaveed84:


jazibhumayun (approved):
LGTM

dataeng/resources/enrollment.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 if [ -z "$NUM_REDUCE_TASKS" ]; then
-    NUM_REDUCE_TASKS=$(( $NUM_TASK_CAPACITY*2 ))
+    NUM_CAPACITY=$(( $NUM_TASK_CAPACITY + $ON_DEMAND_CAPACITY ))
```
estute, line 4: In this case, either `NUM_TASK_CAPACITY` or `ON_DEMAND_CAPACITY` would have a non-zero value, but not both, right?

estute (approved):

dataeng/resources/enrollment.sh
HassanJaveed84: Correct.

PR #1399, opened 2021-07-09
feat: full refresh checkbox also uses larger warehouse

pwnage101:
Typically, full refreshes should use a large warehouse, so this commit
will codify that.

DESUPPORT-959

### DIFF
--- a/dataeng/resources/warehouse-transforms.sh
+++ b/dataeng/resources/warehouse-transforms.sh
@@ -26,6 +26,20 @@ function postCommandChecks {
   fi
 }
 
+# If the FULL_REFRESH_INCREMENTALS checkbox was enabled, switch to a bigger
+# warehouse and tell dbt to do a "full refresh".
+FULL_REFRESH_ARG=""
+if [ "$FULL_REFRESH_INCREMENTALS" = 'true' ]
+then
+    FULL_REFRESH_ARG="--full-refresh"
+    # Only switch to a larger warehouse if the target is prod, since there's no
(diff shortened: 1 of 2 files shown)


dataeng/resources/warehouse-transforms.sh
```diff
@@ -26,6 +26,16 @@ function postCommandChecks {
...
 
+# If the FULL_REFRESH_INCREMENTALS checkbox was enabled, switch to a bigger
+# warehouse and tell dbt to do a "full refresh".
+FULL_REFRESH_ARG=""
+PROFILES_DIR="$WORKSPACE/analytics-secure/warehouse-transforms/"  # The default profile with a smaller warehouse.
+if [ "$FULL_REFRESH_INCREMENTALS" = 'true' ]
+then
+    FULL_REFRESH_ARG="--full-refresh"
+    PROFILES_DIR="$WORKSPACE/analytics-secure/warehouse-transforms/profile-with-big-warehouse/"
+fi
```
brianhw, line 37: Instead of creating a new profile, why not just use a different target?   See https://openedx.atlassian.net/wiki/spaces/DE/pages/2345762996/Creating+and+managing+an+event-based+incremental+table+in+DBT#Initial-load-of-the-incremental-table for details.

pwnage101: Oh!  I must have created that directory locally a long time ago and forgot that it wasn't part of upstream.

brianhw (approved):
LGTM.  👍

PR #1404, opened 2021-07-27
fix: Run m2m for squash merges

jazibhumayun:
W-T Merges to Master job was running on Merge commits whereas some team members use Squash Merge to merge their PR which do not create a merge commit, this PR fixes that issue by using --first-parent flag. https://openedx.atlassian.net/browse/DENG-961

### DIFF
--- a/dataeng/resources/warehouse-transforms-ci-master-merges.sh
+++ b/dataeng/resources/warehouse-transforms-ci-master-merges.sh
@@ -9,29 +9,20 @@ make requirements
 cd $WORKSPACE/warehouse-transforms
 
 COMMIT_ID=$(git rev-parse --short HEAD)
-# Only run CI on merge commits, otherwise exit
-HEAD_COMMIT=$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


dataeng/resources/warehouse-transforms-ci-master-merges.sh
```diff
@@ -9,29 +9,20 @@ make requirements
...
-HEAD_COMMIT=$(git rev-parse HEAD)
-LAST_MERGE_COMMIT=$(git log --merges origin/master --format='%H' --max-count=1)
-if [ $HEAD_COMMIT == $LAST_MERGE_COMMIT ]
-then
-    echo "This is one of merge commit, Run CI"
-else
-    echo "Exiting because not a merge commit"
-    exit 0
-fi 
-
```
jazibhumayun, line 22: We don't need to check the above condition. If team members "Squash Commit" their pull request it will squash all the intermediate commits and pushes as one single commit, in other case if team members use "Merge Commit" even without squashing their intermediate commits the top most commit will me Merge commit, so we don't need to check this condition now.

brianhw (approved):
LGTM.   👍

PR #1405, opened 2021-07-30
feat: creating DBTAutomated job to run schema builder changes

jazibhumayun:
DBT Automated job will take care of running the schema builder changes and on-call person will not need to run dbt-manual. dbt-manual will still be available for DS&A and DE team to quickly apply changes e.g. fixes or updates https://openedx.atlassian.net/browse/DENG-954


dataeng/resources/dbt-automated.sh
```diff
@@ -0,0 +1,31 @@
...
+set -ex
+
+# Setup to run dbt commands
+cd $WORKSPACE/warehouse-transforms
+
+# Using --first-parent flag helps to avoids any intermediate commits that developers sometimes leave and merge without squashing. It helps us ignoring the noise in commit history.
+HEAD_COMMIT=$(git log --first-parent origin/master --format='%H' --max-count=1)
+
+if git log --format=%B -n 1 $HEAD_COMMIT | grep 'Schema Builder automated dbt update at' -q; then isSchemaBuilderPR="true"; else isSchemaBuilderPR="false"; fi
+
```
jazibhumayun, line 11: Keeps polling for warehouse-transform repository and trigger when new changes are merged, then this conditions checks if PR message or title contains `Schema Builder automated dbt update at` and this is how it detects if this is one of Schema Builder generated PR which is merged

jazibhumayun: Writing just for more clarity that if team member/on-call person use "Create a Merge commit" option to merge PR then `Schema Builder automated dbt update at` will be a commit message however if team member use "Squash commit" option to merge PR then `Schema Builder automated dbt update at` will be commit title.

`if git log --format=%B -n 1 $HEAD_COMMIT` tackles both the scenarios

dataeng/jobs/analytics/DBTAutomated.groovy
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                    extensions {
+                        relativeTargetDirectory('warehouse-transforms')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+            }
+            triggers {
+                scm('H/2 * * * *')
```
estute, line 45: I don't think we should run this job this frequently. In fact, doing so would result in 99% of the build history being `It is not a Schema Builder commit. Exiting from job`, and it would be hard to track down which run actually introduced the change.

Ideally, we could use GH webhooks to trigger these with a push model, rather than needing to constantly pull. Perhaps we should revisit that decision at some point.

estute (commented):
In general, looks good. However, with the exception of the bit about polling and the git log checking, the rest of the job is identical to the existing dbt-manual job. I was thinking it would make sense to reuse that code, rather than duplicate it. I don't want you to have to scrap all of this, so what do you think? I feel like there is some risk/complexity in having duplicate code/job definitions, but maybe I am overthinking it.

dataeng/jobs/analytics/DBTAutomated.groovy
jazibhumayun: Actually GH polling isn't that bad, It polls in the background and triggers only when it detects new commits so polling with a frequency of 2 minute is fine.

jazibhumayun: But I would like to have a discovery ticket for GH webhooks may be it can be possible to apply

jazibhumayun:
I think the job web UI should be different. I think having Manual and Automated two different jobs make more sense and doesn't create confusion when debugging the outputs. However the shell script they both run can be combined and have one script that can work for both of them @estute let me how you feel about it.

estute:
Oh, yes, I think we are agreeing. There should be two jobs for sure, but we can probably reuse code behind the scenes.

dataeng/resources/dbt-run.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup to run dbt commands
+cd $WORKSPACE/warehouse-transforms
+
+# Using --first-parent flag helps to avoids any intermediate commits that developers sometimes leave and merge without squashing. It helps us ignoring the noise in commit history.
```
estute, line 7: Interesting. `first-parent` seems useful.

dataeng/resources/dbt-run.sh
```diff
@@ -0,0 +1,31 @@
...
+HEAD_COMMIT=$(git log --first-parent origin/master --format='%H' --max-count=1)
+
+if git log --format=%B -n 1 $HEAD_COMMIT | grep 'Schema Builder automated dbt update at' -q; then isSchemaBuilderPR="true"; else isSchemaBuilderPR="false"; fi
```
estute, line 10: nit: can you split this onto multiple lines for readability?

estute, line 10: other nit: I think we should standardize on one format for variables. `IS_SCHEMA_BUILDER_PR`, for example.

dataeng/jobs/analytics/DBTRun.groovy
```diff
@@ -29,6 +28,7 @@ class DBTManual{
                 stringParam('DBT_RUN_OPTIONS', allVars.get('DBT_RUN_OPTIONS'), 'Additional options to dbt run/test, such as --models for model selection. Details here: https://docs.getdbt.com/docs/model-selection-syntax')
                 stringParam('DBT_RUN_EXCLUDE', allVars.get('DBT_RUN_EXCLUDE'), 'Models to exlude from this run, the default is known incremental models. Leave these if you are not explicitly updating them!')
                 stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+                stringParam('JOB_TYPE', allVars.get('JOB_TYPE'), 'Pass parameter value of manual')
```
estute, line 31: This should always be `manual`, right? If so, I don't think it makes sense to be a user modifiable parameter. Rather, it seems like an environment variable.

estute (approved):
Couple of nits, otherwise, this looks good and can be merged.

In an earlier comment/discussion, I had mentioned trying to make this more DRY, and I was thinking of this: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/WarehouseTransforms.groovy. One DSL `job` definition, and one bash script, that generates multiple jobs that are differentiated by the configs passed into them. I realize that these two jobs differ in structure a little, and to try to make it into one single definition might be taking the DRY principle too far. What are your thoughts?

jazibhumayun:
> Couple of nits, otherwise, this looks good and can be merged.
> 
> In an earlier comment/discussion, I had mentioned trying to make this more DRY, and I was thinking of this: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/WarehouseTransforms.groovy. One DSL `job` definition, and one bash script, that generates multiple jobs that are differentiated by the configs passed into them. I realize that these two jobs differ in structure a little, and to try to make it into one single definition might be taking the DRY principle too far. What are your thoughts?

Yes, I thought about to have one single job DSL but dbt-manual and dbt-automated are slightly different and dbt-automated is using scm poll and I don't think we can incorporate such differences in one single DSL like we have in  https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/analytics/WarehouseTransforms.groovy.

dataeng/jobs/analytics/DBTRun.groovy
jazibhumayun: right, done

dataeng/resources/dbt-run.sh
jazibhumayun: yes we should, done

jazibhumayun: yes, done that

estute (approved):
Changes look good. Make sure to communicate this out to the team as it is a change in behavior.

I have been on the fence about this one for a while and have two final thoughts (you don't need to make these changes, but I just wanted to start a converation):
- Would it make sense to trigger the automated job as part of the warehouse-transforms-ci-master-merges job? As in, if that job passes the tests, call this job as a downstream?
- Maybe in a follow up PR we could add a slack notification to say something like "The following commit has triggered models to be built" and provide a link.

jazibhumayun:
> Changes look good. Make sure to communicate this out to the team as it is a change in behavior.
> 
> I have been on the fence about this one for a while and have two final thoughts (you don't need to make these changes, but I just wanted to start a converation):
> 
> * Would it make sense to trigger the automated job as part of the warehouse-transforms-ci-master-merges job? As in, if that job passes the tests, call this job as a downstream?
> * Maybe in a follow up PR we could add a slack notification to say something like "The following commit has triggered models to be built" and provide a link.

Alright in general your suggested items seems ok to me. I have created this ticket to address your suggestions: https://openedx.atlassian.net/browse/DENG-979

jazibhumayun:
Made the announcement: https://edx-internal.slack.com/archives/CDATH3P6V/p1630412969044300

PR #1422, opened 2021-08-30
PSRE-974 | Fix create-data-czar job

arbabkhalil:
Fixing Artifact path and bash script


devops/resources/create-data-czar.sh
```diff
@@ -9,17 +9,18 @@ create_virtualenv --python=python3.8 --clear
 set -u
 
 cd "$WORKSPACE/configuration"
-pip install -r requirements.txt
+pip install -r util/jenkins/requirements.txt
```
adzuci, line 12: Do you know if we use this requirements file for all Jenkins jobs?

devops/resources/create-data-czar.sh
```diff
@@ -9,17 +9,18 @@ create_virtualenv --python=python3.8 --clear
...
 . util/jenkins/assume-role.sh
 
 assume-role ${ROLE_ARN}
 
 cd util/create_data_czar
 
 # Create Policy
-if CREATE_ORG == "true"  then;
+if [ ${CREATE_ORG} == "true"  ]; then
```
adzuci, line 20: This if statement makes me wonder if we should have a dry run flag that can be set to make the job print what it would do without actually doing it.  Would that be useful?

devops/jobs/CreateDataCzar.groovy
```diff
@@ -42,11 +42,11 @@ class CreateDataCzar{
                 stringParam('ORGANIZATION',
                         'Name of organization to create data czar. e.g alaskax')
                 booleanParam('CREATE_ORG', true,
-                        'true if create new organization or false if organization is already created.')
+                        'Check if create new organization or Uncheck if organization is already created.')
                 stringParam('USER_EMAIL',
                         'User Email address to generate Data Czar')
                 fileParam('user_gpg_key.gpg', 
-                        'Paste the GPG key for encryption of aws credentials')
+                        'Attach the GPG key file for encryption of aws credentials')
```
adzuci, line 49: Should this be a file or should we just paste in the public key?  Also could you add the word public to this description?

devops/jobs/CreateDataCzar.groovy
```diff
@@ -42,11 +42,11 @@ class CreateDataCzar{
                 stringParam('ORGANIZATION',
                         'Name of organization to create data czar. e.g alaskax')
                 booleanParam('CREATE_ORG', true,
-                        'true if create new organization or false if organization is already created.')
+                        'Check if create new organization or Uncheck if organization is already created.')
```
adzuci, line 45:
```suggestion
                        'Check this box if you are creating a new organization.')
```

adzuci (approved):
LGTM, also left some small optional change suggestions.

jdmulloy (approved):
Looks good to me, with @adzuci's suggestions.

devops/jobs/CreateDataCzar.groovy
arbabkhalil: We should attach the file for public key, this is same as previous workflow.
I am adding public in this message.

devops/resources/create-data-czar.sh
arbabkhalil: Not in all jenkins jobs, I have seem a few that are using main requirements.txt file. This was a suggestion from Nadeem in a review meeting we had on meet.

PR #1425, opened 2021-09-10
Feat: Add App-Permissions for other IDAs

NIXKnight:
This PR adds services other than `edxapp` into App-Permissions. See [ADR](https://edx.readthedocs.io/projects/edx-django-utils/en/latest/decisions/0005-user-and-group-management-commands.html) for more context.

### DIFF
--- a/devops/resources/run-app-permissions.sh
+++ b/devops/resources/run-app-permissions.sh
@@ -17,22 +17,53 @@ env
 assume-role ${ROLE_ARN}
 cd $WORKSPACE/configuration/playbooks
 
+# Function to set ASG and path variables
+function setASGAndPath () {
+	service_name="$1"
+	if [[ "${service_name}" == "lms" ]] || [[ "${service_name}" == "cms" ]] || [[ "${service_name}" == "edxapp" ]]; then
+		INVENTORY=$(./active_instances_in_asg.py --asg ${ENVIRONMENT}-${DEPLOYMENT}-worker)
+		SERVICE_ENV_PATH="/edx/app/edxapp/edxapp_env"
+		SERVICE_PYTHON_PATH="/edx/app/edxapp/venvs/edxapp/bin/python"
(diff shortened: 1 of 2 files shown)


kdmccormick:
@NIXKnight and @adzuci , thanks for tagging me. Do let me know if you have specific questions I can help with, however, I won't be able to provide general reviews on app-permissions-related PRs.

devops/resources/run-app-permissions.sh
```diff
@@ -17,22 +17,53 @@ env
...
-INVENTORY=$(./active_instances_in_asg.py --asg ${ENVIRONMENT}-${DEPLOYMENT}-worker)
 if [[ -n ${INVENTORY} ]]; then
     ansible-playbook -i ${INVENTORY} manage_edxapp_users_and_groups.yml \
+										 -e "env_path=${SERVICE_ENV_PATH}" -e "python_path=${SERVICE_PYTHON_PATH}" \
+										 -e "manage_path=${SERVICE_MANAGE_PATH}" -e "service=${service}" \
                      -e@${configfile} -e "group_environment=${ENVIRONMENT}-${DEPLOYMENT}" \
-                     --user ${USER} --tags manage-${JOB_TYPE}
+                     --user ${USER} --tags ${ANSIBLE_TAG}
 else
     echo "Skipping ${ENVIRONMENT} ${DEPLOYMENT}, no worker cluster available, get it next time"
```
adzuci, line 68: While we're updating this script we may want to expand on this log message, what does it mean by get it next time?

adzuci (approved):
I think getting the job types from extra vars makes sense.  I would be curious if @jdmulloy has any suggestions for how we could further simplify this logic, but it looks like it will work so I'm going to approve it.  It would be helpful to have a PR description and we may want to consider updating an ADR or adding one to explain why we chose to make some of these changes.

PR #1433, opened 2021-09-22
fix: fix setting end date in user retirement archiver script

estute:
Tested locally: 
```
jenkins-job-dsl (estute/deng-983-date) % END_DATE=2021-09-22
jenkins-job-dsl (estute/deng-983-date) % COOL_OFF_DAYS=10
jenkins-job-dsl (estute/deng-983-date) % END_DATE=$(date --date "$(date "+%B %d, %Y") -$COOL_OFF_DAYS days" +%y-%m-%d)
jenkins-job-dsl (estute/deng-983-date) % echo $END_DATE
21-09-12
```

### DIFF
--- a/devops/resources/user-retirement-archiver.sh
+++ b/devops/resources/user-retirement-archiver.sh
@@ -19,9 +19,9 @@ assume-role ${ROLE_ARN}
 cd $WORKSPACE/tubular
 pip install -r requirements.txt
 
-# In case this is being run without an explicit END_DATE, default to running with "now"
-if [[ "$END_DATE" == "" ]]; then
-    END_DATE=$(date +%Y-%m-%d)
+# In case this is being run without an explicit END_DATE, default to running with "now" - COOL_OFF_DAYS
+if [[ ! -v END_DATE ]]; then
+    END_DATE=$(date --iso --date "$(date --iso) - $COOL_OFF_DAYS days")
 fi
 
 # Call the script to read the retirement statuses from the LMS, send them to S3, and delete them from the LMS.


devops/resources/user-retirement-archiver.sh
```diff
@@ -19,9 +19,9 @@ assume-role ${ROLE_ARN}
 cd $WORKSPACE/tubular
 pip install -r requirements.txt
 
-# In case this is being run without an explicit END_DATE, default to running with "now"
-if [[ "$END_DATE" == "" ]]; then
-    END_DATE=$(date +%Y-%m-%d)
+# In case this is being run without an explicit END_DATE, default to running with "now" - COOL_OFF_DAYS
+if [[ ! -v END_DATE ]]; then
+    END_DATE=$(date --date "$(date "+%B %d, %Y") -$COOL_OFF_DAYS days" +%y-%m-%d)
```
brianhw, line 24: naive question -- where does COOL_OFF_DAYS come from?   And will it always be defined?

pwnage101, line 24: I think you should use ISO formatting instead of manual formatting:
```suggestion
    END_DATE=$(date --iso --date "$(date --iso) - $COOL_OFF_DAYS days")
```
Currently, the output produces a 2-digit year instead of 4-digit.

pwnage101: COOL_OFF_DAYS is a job parameter with a default value: https://github.com/edx/jenkins-job-dsl/blob/47b9a3d63670ee237faca055c56f05d909172752/platform/jobs/RetirementJobs.groovy#L251

Seems okay to assume it exists.

pwnage101 (changes requested):

devops/resources/user-retirement-archiver.sh
estute: Ah, good catch thanks

adzuci (approved):
LGTM, TIL iso things.

PR #1436, opened 2021-09-24
fix: remove obsolete cache uploader job

mraarif:
we recently decided to remove our `a11y` check from `edx-platform` CI and run the migrations check separately using GHA, as part of `a11y` check we ran `bokchoy` tests and we had this job setup some DB cache for `bokchoy` tests to run, we don't need this job now.

relevant JIRA issue: https://openedx.atlassian.net/browse/BOM-2719


jmbowman (approved):
Please add a description (and preferably ticket link) to the PR description.  Thanks!

mzulqarnain1 (approved):

PR #1438, opened 2021-09-28
feat: Tableau Server automated backup

jazibhumayun:
This job runs every once a week and create Tableau Server backup and place it on S3 bucket.

### DIFF
--- a/dataeng/resources/tableau-backup.sh
+++ b/dataeng/resources/tableau-backup.sh
@@ -14,8 +14,7 @@ DATE="$(date '+%Y.%m.%d')"
 BACKUPNAME="$DATE.11.00-tableau-data-backup.tsbak"
 
 # Check if tableau backup file exist, exit if the file does not exist
-fileStatus=$([ -f /home/tableau_backups/$BACKUPNAME ] && echo 'Found' || echo 'NotFound' )
-if [[ "$fileStatus" == "NotFound" ]]
+if [[ ! -f /home/tableau_backups/$BACKUPNAME ]]
 then
     echo "$BACKUPNAME Backup file does not exist. It seems to be some problem generating tableau backup. Refer to Administration of Tableau wiki docs to debug Tableau Regular Backup Automation"
     exit 1
--- a/dataeng/resources/tableau-backup.sh
+++ b/dataeng/resources/tableau-backup.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+set -ex
+
+
+# This script ssh into Tableau server. Default centos user does not have access on Tableau directories
+# so it runs commands as root. Backups are being created using tsm maintenance backup utility. This script
+# copies backup file into s3 bucket and then remove backup file from tableau server.
+
+ssh -tt $USER_NAME@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "S3_BUCKET=$S3_BUCKET" '
+
+# It gets the current date from system in 2021.10.05 format
+DATE="$(date '+%Y.%m.%d')"
+# It adds the date as prefix on backup name. Backup runs at 11:00 UTC so it also prefixed in the name.
+BACKUPNAME="$DATE.11.00-tableau-data-backup.tsbak"
+
+# Check if tableau backup file exist, exit if the file does not exist
+fileStatus=$([ -f /home/tableau_backups/$BACKUPNAME ] && echo 'Found' || echo 'NotFound' )
+if [[ "$fileStatus" == "NotFound" ]]
+then
+    echo "$BACKUPNAME Backup file does not exist. It seems to be some problem generating tableau backup. Refer to Administration of Tableau wiki docs to debug Tableau Regular Backup Automation"
+    exit 1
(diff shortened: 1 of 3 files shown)


dataeng/resources/tableau-backup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -ex
+
+
+ssh -tt centos@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "
```
estute, line 5: I would recommend parameterizing the username as well, as we probably will be moving off of CentOS

dataeng/resources/tableau-backup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
estute, line 1: Would you mind adding a comment into the top of this script describing what it does?

dataeng/resources/tableau-backup.sh
```diff
@@ -0,0 +1,11 @@
...
+set -ex
+
+
+ssh -tt centos@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "
+
+sudo -H -u root bash -c 'aws s3 cp /var/opt/tableau/tableau_server/data/tabsvc/files/backups/$(date '+%Y.%m.%d').11.00-tableau-data-backup.tsbak s3://edx-analytics-tableau/backups/$(date '+%Y.%m.%d').11.00-tableau-data-backup.tsbak'
+sudo -H -u root bash -c 'rm /var/opt/tableau/tableau_server/data/tabsvc/files/backups/$(date '+%Y.%m.%d').11.00-tableau-data-backup.tsbak'
```
estute, line 8: Is there a reason we need to run this as root? I usually think it should be avoided wherever possible. Maybe we can look into changing the ownership of the directories or files in play here.

estute, line 8: How much control over the names of backfup files do we have? I imagine we will be upgrading the version of tableau we use. One thing I find useful is to create a string for the name of the file you will be manipulating and use it as a variable. It makes things a bit more reabable.

dataeng/resources/tableau-backup.sh
jazibhumayun: Sure, absolutely

dataeng/resources/tableau-backup.sh
jazibhumayun: Updated the code. I am now creating an environment variable so we will have to change the name only place.

jazibhumayun: I have tried multiple ways but nothing else work. I feel running these two commands as root is less risky than than changing directory ownership. It will require as to create a new user group and add all the required users into that group and then change owership of tableau specific directory so that's why I was hesitant doing that and opted this way.

dataeng/resources/tableau-backup.sh
```diff
@@ -0,0 +1,23 @@
...
+# This script ssh into Tableau server. Default centos user does not have access on Tableau directories
+# so it runs commands as root. Backups are being created using tsm maintenance backup utility. This script
+# copies backup file into s3 bucket and then remove backup file from tableau server.
+
+ssh -tt $USER_NAME@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "S3_BUCKET=$S3_BUCKET" '
+
+# It gets the current date from system in 2021.10.05 format
+DATE="$(date '+%Y.%m.%d')"
+# It adds the date as prefix on backup name. Backup runs at 11:00 UTC so it also prefixed in the name.
+BACKUPNAME="$DATE.11.00-tableau-data-backup.tsbak"
```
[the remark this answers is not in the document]
jazibhumayun: sure done that

dataeng/resources/tableau-backup.sh
jazibhumayun: So I have now changed the default location for backup files to get stored. I created a new folder '/home/tableau_backups' and gave 'centos' and 'tableau' users access to that folder

jazibhumayun: In this way this is pretty restrictive for `centos` as it was before and we can avoid running these commands as root. As we will now be only iterating with  `/home/tableau_backups `

estute (commented):
One question, but aside from that, looks great!

dataeng/resources/tableau-backup.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Check if tableau backup file exist, exit if the file does not exist
+fileStatus=$([ -f /home/tableau_backups/$BACKUPNAME ] && echo 'Found' || echo 'NotFound' )
+if [[ "$fileStatus" == "NotFound" ]]
```
estute, line 18: Not a blocker, but you can actually just use the `find` command as the test in the `if` statement, like:
```
if [[ ! -f /home/tableau_backups/$BACKUPNAME ]] then
    echo "backup file not found"
    exit 1
fi
```

estute (approved):

PR #1445, opened 2021-10-13
feat: add new job to run user retirement cleanup script against snowf…

estute:
…lake


dataeng/resources/snowflake-user-retirement-status-cleanup.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -ex
+
+# Setup
+cd $WORKSPACE/analytics-tools/snowflake
+make requirements
+
+python retirement_cleanup.py \
+    --key_path $WORKSPACE/analytics-secure/$KEY_PATH \
+    --passphrase_path $WORKSPACE/analytics-secure/$PASSPHRASE_PATH \
```
adzuci, line 10: I'm guessing this is a common pattern, but I'm curious if `PASSPHRASE_PATH` could be more descriptive.

adzuci (approved):
LGTM, though I didn't confirm the arguments match what `retirement_cleanup.py` expects.

dataeng/jobs/analytics/SnowflakeUserRetirementStatusCleanup.groovy
```diff
@@ -0,0 +1,80 @@
...
+                        relativeTargetDirectory('analytics-tools')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+            }
+            triggers {
+                // Run on the seventh day of every month. The user-retirement-archiver job
+                // runs on the first of every month, so this should allow enough time to
+                // address any issues with the archival process.
```
adzuci, line 59: Wonder if we should link to this line in the other trigger and state that they should both be updated in tandem if they do need to be updated.

estute: I'll link it once it merges

brianhw (approved):
LGTM.  👍 

There is a note that the user-retirement job would point to this user-retirement-status cleanup.  But I'm assuming that this can run on whatever schedule, and would just clean up whatever is deleted or be a no-op.  I don't understand why it would need to be disabled.

estute:
@brianhw yeah, you are right. I suppose I just wanted to highlight the relationship between the two. Let me see if I can improve that

PR #1449, opened 2021-10-18
fix: Use new requirements file for sysadmin repo

jdmulloy:
This relates to https://openedx.atlassian.net/browse/DOS-2677


devops/resources/check-lifecycle-hooks.sh
```diff
@@ -12,7 +12,7 @@ cd $WORKSPACE/configuration
 . util/jenkins/assume-role.sh
 
 cd $WORKSPACE/sysadmin
-pip install -r requirements.txt
+pip install -r requirements/base.txt
```
adzuci, line 15: We may have also wanted to update the comment on https://github.com/edx-ops/sysadmin/blob/master/requirements/constraints.txt#L11

PR #1453, opened 2021-10-29
feat: allow cleanup-python-code-create-pr to read the PR description from a file

rgraber:
Update cleanup-python-code-create-pr to look for a file containing the desired PR description before defaulting to the current message. Allows for more customization in creating PRs from this job.


testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -54,7 +54,11 @@ do_one_repo () {
   echo "Running script to create PR..."
   cd "$WORKSPACE/testeng-ci"
   pip install -r requirements/base.txt
-  local message="$(cat <<EOF
+  if [ -s "$repo_dir/.git/cleanup-python-code-message" ]
+  then
+     local message=$(cat "$repo_dir/.git/cleanup-python-code-message")
+  else
+     local message="$(cat <<EOF
```
robrap, line 61: What does this line do exactly, and do we want one or the other, or potentially both?

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -54,7 +54,11 @@ do_one_repo () {
   echo "Running script to create PR..."
   cd "$WORKSPACE/testeng-ci"
   pip install -r requirements/base.txt
-  local message="$(cat <<EOF
+  if [ -s "$repo_dir/.git/cleanup-python-code-message" ]
+  then
+     local message=$(cat "$repo_dir/.git/cleanup-python-code-message")
```
robrap, line 59: Is there an explanation around the choice of dropping this file in `.git`?

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -54,7 +54,11 @@ do_one_repo () {
   echo "Running script to create PR..."
   cd "$WORKSPACE/testeng-ci"
   pip install -r requirements/base.txt
-  local message="$(cat <<EOF
+  if [ -s "$repo_dir/.git/cleanup-python-code-message" ]
```
robrap, line 57: Could you add a version of the PR description comment in the code? I had no idea what this was doing or used for without it.

rgraber: I don't understand. What should the comment be about?

testeng/resources/cleanup-python-code-create-pr.sh
rgraber: I think for now we want one or the other. There isn't really a use case for making this too customizable yet, and doing so would require non trivial changes elsewhere (to allow for more options on the jenkins job)

testeng/resources/cleanup-python-code-create-pr.sh
rgraber: Ah I think I get it now. Was having trouble with sentence parsing

testeng/resources/cleanup-python-code-create-pr.sh
rgraber: Seemed like a place unlikely to conflict with anything else

testeng/resources/cleanup-python-code-create-pr.sh
robrap: Thanks!

testeng/resources/cleanup-python-code-create-pr.sh
robrap: I still don’t know what this does.

testeng/resources/cleanup-python-code-create-pr.sh
robrap: Ok. Doesn’t feel right to me, but I don’t have a proposal, so I guess we’re going with it, unless you ask for other ideas. 🙂

robrap (approved):
Still curious about one part, but doesn’t need to block this. Thanks.

testeng/resources/cleanup-python-code-create-pr.sh
timmc-edx: It's also a place that won't conflict with the rest of the working directory.

PR #1456, opened 2021-11-01
feat: make PR/commit more configurable in cleanup-python-code

rgraber:


jinder1s (approved):

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -75,7 +75,7 @@ EOF
 )"
   fi
   python -m jenkins.pull_request_creator --repo-root="$repo_dir" \
-         --base-branch-name="cleanup-python-code" --commit-message="$message" \
```
timmc-edx, line 78: Is this a decision we can push back up to the caller?

jinder1s: +1 this. Seems like a good default, but could the script be used for something else besides a feat?

rgraber: We can always make another optional file

timmc-edx: Oh, I just meant that the caller would include the conventional-commit prefix in their commit message from the get-go.

rgraber: Still a little bit confused. Do you mean as part of the job config?

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -73,10 +67,25 @@ The following packages were installed:
 \`$PACKAGES\`
 EOF
 )"
+  # if present and non-empty, set the commit message from .git/cleanup-python-code-commit-message
+  # this will allow jobs using this script to make custom commit messages
+  if [ -s "$repo_dir/.git/cleanup-python-code-commit-message" ]
+  then
+    local commit_msg=$(cat "$repo_dir/.git/cleanup-python-code-commit-message")
+  else
+    local commit_msg=$description
```
timmc-edx, line 76:
```suggestion
    local commit_msg="$description"
```
I think that without quotes, this will convert all whitespace runs into single spaces.

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -73,10 +67,25 @@ The following packages were installed:
...
+  fi
+
+  # if present and non-empty, set the PR description from .git/cleanup-python-code-description
+  if [ -s "$repo_dir/.git/cleanup-python-code-description" ]
+  then
+    local pr_description=$(cat "$repo_dir/.git/cleanup-python-code-description")
+  else
+    local pr_description=$description
```
timmc-edx, line 84:
```suggestion
    local pr_description="$description"
```
same here

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -55,14 +55,8 @@ do_one_repo () {
...
-  # if present and non-empty, set the PR description from .git/cleanup-python-code-message
-  # this will allow jobs using this script to make custom PR descriptions
-  if [ -s "$repo_dir/.git/cleanup-python-code-message" ]
-  then
-     local message=$(cat "$repo_dir/.git/cleanup-python-code-message")
-  else
-     local message="$(cat <<EOF
-Python code cleanup by the cleanup-python-code Jenkins job.
+  local description="$(cat <<EOF
+feat: Python code cleanup by the cleanup-python-code Jenkins job.
```
timmc-edx, line 59: Some thoughts:

- It would be nice if we could have custom commit messages/PR descriptions that also automatically include the script and package lists.
- This wasn't what I had in mind before, but now that you mention it we probably *do( want to just go ahead and add "commit title" to the job parameters so that people have to actually describe it.
- I think it's rare that we'll want to have different commit and PR titles/bodies.

What if we did this?

- Add commit/PR title as a single-line field, with a link to the conventional-commits doc. (And make it required, if that's something we can enforce in the UI.)
- Add commit/PR body as an optional multi-line field
- Body is constructed of: Commit body (if provided), contents of `.git/cleanup-python-code-commit-message` (if present), and finally the autogenerated information here about what job created the PR and what scripts and packages were involved
- Commit is just commit title + that body (params already distinguish title vs. body of PR)

...and then I think we end up in a better situation for people who are looking at these PRs and commits out of context.

rgraber: Do we really want the commit to include the full body? That will be a lot of information

timmc-edx: I don't think it would be *too* too big. I'll often write large commit messages, but I just make sure to do it "pyramid style", with the most concise and important parts up at the top and the most incidental and expanded parts at the bottom -- people don't have to read the whole thing, but it's there if they need it. In our case, I don't think it would run to more than 10-20 lines, max (unless there's a huge script, I guess!)

testeng/jobs/cleanupPythonCode.groovy
```diff
@@ -18,6 +23,8 @@ job('cleanup-python-code') {
     pythonVersion = '${pythonVersion}'
     packagesToInstall = '${packages}'
     scriptsToRun = '${scripts}'
+    commitMessage='${commiteMessage}'
```
timmc-edx, line 26:
```suggestion
    commitMessage='${commitMessage}'
```
typo

testeng/jobs/cleanupPythonCode.groovy
```diff
@@ -10,6 +10,11 @@ job('cleanup-python-code') {
         stringParam('repoNames', null, 'Comma or space separated list of names of (public) target repositories in $org org')
         choiceParam('pythonVersion', ['3.8', '3.5', '2.7'], 'Version of python to use')
         stringParam('packages', '', 'Comma or space separated list of packages to install (optional)')
+        stringParam('commitMessage', null, 'Commit message (will also be used as PR title).
+            See https://github.com/edx/open-edx-proposals/blob/master/oeps/best-practices/oep-0051-bp-conventional-commits.rst
+            for more information on best practices for commit messages')
+        textParam('prBody', '', 'Additional information for the PR body. You can add dynamic information to the
+        PR body by writing to .git/cleanup-python-code-description in the bash script (optional)')
```
timmc-edx, line 17:
```suggestion
        stringParam('title', null, 'Commit message first line (will also be used as PR title).
            See https://github.com/edx/open-edx-proposals/blob/master/oeps/best-practices/oep-0051-bp-conventional-commits.rst
            for more information on best practices for commit messages')
        textParam('body', '', 'Additional information for the commit message and PR body.
        The script and packages will be included automatically.
        You can add dynamic information to the
        commit/PR body by writing to .git/cleanup-python-code-description in the bash script (optional).')
```
^ What do you think about breaking it up into title/body like this?

testeng/resources/cleanup-python-code-create-pr.sh
rgraber: The script we are running is already well over 10-20 lines.

testeng/jobs/cleanupPythonCode.groovy
rgraber: I'm still not convinced including the PR body in the commit message is a good idea. In the upcoming use case, between the output of the requires.txt files and the actual script, that will be a huge commit message.

timmc-edx: Hmm, no, you're right -- the dynamic output could be huge. :thinking:

For the script, though, can we do something like this?

```
curl -sS "https://raw.githubusercontent.com/edx/edx-cookiecutters/2299c8e9c570155a4ea28c506dab4f7d22b3a87a/scripts/update_setup_py.sh" > update_setup_py.sh
bash update_setup_py.sh
```

rgraber: oh cute. i'll try that.

rgraber: It worked! The PR description is still long enough that I'd be worried about putting it in a commit message though

timmc-edx: Yeah, agreed, it could be really large. But it could also be too large for a PR description, too.

One possibility is to use a details/summary block for the PR body, like so:

<details>
<summary>Possibly giant chunk of output</summary>

```
output
output
output
```
</details>

I guess that could be either done by this PR-creator job or by whatever script it runs.

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -55,31 +57,47 @@ do_one_repo () {
...
-     local message="$(cat <<EOF
+    local pr_description_from_file=''
+  fi
+
+  local description="$(cat <<EOF
+$base_pr_body
+
+<details>
+<summary> Additional information from script execution</summary>
+$pr_description_from_file
```
timmc-edx, line 73:
```suggestion
<summary>Additional information from script execution</summary>

$pr_description_from_file
```

nit: I'd suggest an additional line break here after the summary element just to avoid an annoying Markdown situation where if description from file starts with a code block, it needs an empty line before it. (Otherwise it doesn't render as a code block.)

Suggestion also removes stray space.

timmc-edx (approved):
Looks good! I'd still prefer having the scripts and packages included in the commit, perhaps with a request to put any large scripts into an external file that can be separately reviewed, but I don't consider this blocking.

PR #1461, opened 2021-11-10
chore:adding additional masters programs from prod to updatemasterssandbox.

jsongedx:


devops/jobs/UpdateMastersSandbox.groovy
```diff
@@ -79,6 +79,10 @@ class UpdateMastersSandbox {
         )
         textParam("program_uuids",
                   "ef82b8e2-49ed-45b0-982d-c836deef507b:special-masters",
+                  "2b4fb318-5bec-4484-9b6f-e279fa0f4693:asu-supply-chain",
+                  "30616bb3-bc95-41f9-84c6-bc1a40494b4b:iu-it-management",
+                  "df6bb0b1-b543-46d4-abfb-16dd81dcd936:purdue-civil-engineering",
+                  "721f524b-da8f-4923-83b4-a18de44fdc43:utaustin-data-science",
```
matthugs, line 85: The default value we want is the string `"ef82b8e2-49ed-45b0-982d-c836deef507b:special-masters,30616bb3-bc95-41f9-84c6-bc1a40494b4b:iu-it-management,df6bb0b1-b543-46d4-abfb-16dd81dcd936:purdue-civil-engineering,721f524b-da8f-4923-83b4-a18de44fdc43:utaustin-data-science"`, which is different from what you have here.

By contrast, outside of the quotation marks in this context in the code the commas are used to separate different parameters being passed to the textParam function/macro/whatever they'd be called in groovy
```suggestion
                  "2b4fb318-5bec-4484-9b6f-e279fa0f4693:asu-supply-chain,30616bb3-bc95-41f9-84c6-bc1a40494b4b:iu-it-management,df6bb0b1-b543-46d4-abfb-16dd81dcd936:purdue-civil-engineering,721f524b-da8f-4923-83b4-a18de44fdc43:utaustin-data-science",
```

This may actually be more difficult than I'm representing it as, though, since we might get complaints about a single line being too long if we apply exactly the suggestion I have above. Keeping lines short is a common thing folks get concerned about in code contexts. So we may need to learn how to concatenate several strings (i.e. stuff in `""`s) together in groovy, and/or how to break long strings up across multiple lines in groovy.

matthugs (changes requested):

matthugs (approved):

PR #1466, opened 2021-11-16
fix: schema builder hub token

jazibhumayun:
Exporting github token as an environment variable that will be used by Hub to created pull request


brianhw (approved):
Just adding in your explanation here:  

The jenkins job before was using both environment variables Username and Token instead of Username and Password.  It was probably running due to cached tokens. This PR moves it away from using username/password, and instead just using the github token."

LGTM.

PR #1472, opened 2021-12-01
fix: temporarily disable usages of build-name-setter

pwnage101:
Currently, jobs which use the build-name-setter plugin in Build Jenkins
are crashing.  This is a temporary mitigation to get the user retirement
pipeline back up and running while we try to fix the problem on our own
time.

DENG-1063/PSRE-1204


estute (approved):

brianhw (approved):
Looks like a reasonable hack -- and I don't presume anything is actually depending on the format of the build name.   But ... I'll definitely miss it -- it's helpful info.

adzuci (commented):
🎉

PR #1474, opened 2021-12-06
chore: update schema builder python venv to 3.8 version

jazibhumayun:
As part of dbt version upgrade, we are updating python version from 3.7 to 3.8 in schema builder repository. This PR is updating python virtual environment version to 3.8 to run schema builder script in jenkins.

### DIFF
--- a/dataeng/jobs/analytics/SnowflakeSchemaBuilder.groovy
+++ b/dataeng/jobs/analytics/SnowflakeSchemaBuilder.groovy
@@ -44,14 +44,8 @@ class SnowflakeSchemaBuilder {
             publishers common_publishers(allVars)
             steps {
 
-                virtualenv {
-                    pythonName('PYTHON_3.7')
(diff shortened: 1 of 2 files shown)


dataeng/jobs/analytics/SnowflakeSchemaBuilder.groovy
```diff
@@ -44,14 +44,8 @@ class SnowflakeSchemaBuilder {
             publishers common_publishers(allVars)
             steps {
 
-                virtualenv {
-                    pythonName('PYTHON_3.7')
```
jazibhumayun, line 48: Installed python3.8 version from shipping panda plugin is not working properly.
I saw some places where we are using the following syntax to provision python3.8 virtual env

brianhw (approved):
This seems fine to me -- we had lots of jobs in the past that create venvs in shell scripts.

PR #1475, opened 2021-12-06
chore: dbt v0.21.1 upgrade source freshness

jazibhumayun:
`dbt source snapshot-freshness` has been renamed to `dbt source freshness` from version dbt v0.21.0 and above


brianhw (approved):
LGTM.  👍

PR #1480, opened 2021-12-13
chore: load only required seeds using SEED_SELECTOR

brianhw:
In the past, all seeds were loaded on every run of a DBT workflow.
But since all runs use a MODEL_SELECTOR to run a subset of models,
only a subset of seeds are actually needed.  Excluding these helps
streamline the Amplitude-extraction workflow, which writes to
different ("amplitude_"-prefixed) schema names, and some seeds get
loaded into a schema that doesn't exist.

In support of DENG-1070.

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransforms.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransforms.groovy
@@ -23,6 +23,7 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('DBT_COMMAND', env_config.get('DBT_COMMAND', allVars.get('DBT_COMMAND', 'run')), 'dbt command to be executed.  Default is \'run\'.')
+                    stringParam('SEED_SELECTOR', env_config.get('SEED_SELECTOR', allVars.get('SEED_SELECTOR', '*')), 'Model selector that will be used when running the seed step.  Defaults to "*".')
                     stringParam('SKIP_TESTS', env_config.get('SKIP_TESTS', 'false'), 'Set to \'true\' to skip all tests. All other values will allow tests to run given the other configuration values.')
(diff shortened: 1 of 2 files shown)


pwnage101 (approved):
I'm a bit confused how this works without actually adding any tags to the seeds (to me it looks like the seeds are tagless), so I would have assumed this would prevent seeds from ever running.  However, it seems like you tested it and it works, so...

brianhw:
Update to:

In the past, all seeds were loaded on every run of a DBT workflow.
But since all runs use a MODEL_SELECTOR to run a subset of models,
only a subset of seeds are actually needed.  Excluding these helps
streamline the Amplitude-extraction workflow, which writes to
different ("amplitude_"-prefixed) schema names, and some seeds get
loaded into a schema that doesn't exist.  This will be accomplished
by using a non-default value of SEED_SELECTOR, while all other jobs
would continue to use '*' as the selector.

pwnage101 (approved):

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -23,6 +23,7 @@ class WarehouseTransforms{
                     stringParam('DBT_PROFILE', env_config.get('DBT_PROFILE', allVars.get('DBT_PROFILE')), 'dbt profile from analytics-secure to work on.')
                     stringParam('DBT_TARGET', env_config.get('DBT_TARGET', allVars.get('DBT_TARGET')), 'dbt target from analytics-secure to work on.')
                     stringParam('DBT_COMMAND', env_config.get('DBT_COMMAND', allVars.get('DBT_COMMAND', 'run')), 'dbt command to be executed.  Default is \'run\'.')
+                    stringParam('SEED_SELECTOR', env_config.get('SEED_SELECTOR', allVars.get('SEED_SELECTOR', '\\*')), 'Model selector that will be used when running the seed step.  Defaults to "\\*".')
```
pwnage101, line 26: It looks like you changed this already, but when it was still `'"*"'`, I wrote this:

Actually, are you sure you want double quotes inside the single quotes?  I assume that makes the text box in the web UI contain `"*"` by default, but that exposes the low-level machinery of the bash.

pwnage101: but, it looks like you're trying different escaping sequences.  If you find that still isn't working, I think just double quoting the variable in the bash script is the way to go.

brianhw:
Yup, double-quoting is what I needed.  Single-quotes would have worked, but Jenkins would not use single-quotes when the value was just an asterisk, but then use single-quotes on all other forms I tried.   But this works.

pwnage101 (approved):

PR #1483, opened 2021-12-20
chore: Enable ORA Grading MFE for sandboxes

jnlapierre:
chore: Enable ORA Grading MFE for sandboxes, configuration updates.

### DIFF
--- a/devops/jobs/CreateSandbox.groovy
+++ b/devops/jobs/CreateSandbox.groovy
@@ -192,8 +192,8 @@ class CreateSandbox {
                 booleanParam("learner_portal_enterprise",false,"Enable Enterprise Learner Portal MFE")
                 stringParam("learner_portal_enterprise_version","master","The repository version of the frontend-app-learner-portal-enterprise")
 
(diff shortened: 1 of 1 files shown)


devops/jobs/CreateSandbox.groovy
```diff
@@ -192,6 +192,9 @@ class CreateSandbox {
                 booleanParam("learner_portal_enterprise",false,"Enable Enterprise Learner Portal MFE")
                 stringParam("learner_portal_enterprise_version","master","The repository version of the frontend-app-learner-portal-enterprise")
 
+                booleanParam("ora-grading",false,"Enable ORA Grading MFE")
+                stringParam("ora-grading_version","master","The repository version of the frontend-app-ora-grading")
```
matthugs, line 196: This sets up `ora-grading_version` as an environment variable. Hyphens aren't super kosher in that context; I think many implementations would work, but stylistically we really should use only alphnumeric and underscores here.

matthugs (changes requested):

matthugs (approved):

adzuci (approved):

PR #1485, opened 2022-01-04
fix: Update Minos Job Timeout & Decrease Frequency

adzuci:
This change is to resolve the timeouts we're seeing in the minos cleanup jobs.


jdmulloy (commented):
I'm not sure it makes sense to increase timeout. It's ok for the job to fail and try again. I think it needs to run frequently to refresh something in AWS. I think AWS has a 1 hour timer where if you haven't pinged it, it will terminate an instance.

One of the minos jobs hits that API, I don't remember which.

adzuci:
Closing, since we opted to increase the OpsGenie heartbeat from 1hr to 2hr.

PR #1489, opened 2022-01-09
feat: move retirement jobs in analytics jenkins

jazibhumayun:
moving retirement jobs (user retirement and partner reporter) in analytics jenkins new

### DIFF
--- a/dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
+++ b/dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
@@ -1,0 +1,60 @@
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_authorization
+
+class RetirementJobEdxTriggers{
+    public static def job = { dslFactory, allVars ->
+        allVars.get('ENVIRONMENTS').each { environment, env_config ->
+            dslFactory.job("$environment"){
+                // Creates user_retirement_trigger_<environment> and retirement_partner_report_trigger_<environment> jobs
+                // This defines the job which triggers the collector and reporter job for a given environment.
+                description("Scheduled trigger of the " + env_config.get('DOWNSTREAM_JOB_NAME') + " job for the " + env_config.get('DOWNSTREAM_JOB_NAME') + " environment")
+                // Marking the jobs disabled, they will be enabled as part of DENG-1101
+                disabled(env_config.get('DISABLED'))
+
+                authorization common_authorization(env_config)
+                // Disallow this job to have simultaneous instances building at the same
+                // time.  This might help prevent race conditions related to triggering
+                // multiple retirement driver jobs against the same user.
+                concurrentBuild(false)
+                triggers common_triggers(allVars, env_config)
+
+                // keep jobs around for 30 days
+                // allVars contains the value for DAYS_TO_KEEP_BUILD which will be used inside AnalyticsConstants.common_log_rotator
+                logRotator common_log_rotator(allVars)
+
+                wrappers {
+                    buildName('#${BUILD_NUMBER}')
+                    timestamps()
+                    colorizeOutput('xterm')
+                }
+                wrappers common_wrappers(allVars)
+                parameters secure_scm_parameters(allVars)
+                steps {
+                    downstreamParameterized {
+                        trigger(env_config.get('DOWNSTREAM_JOB_NAME')) {
+                            // This section causes the build to block on completion of downstream builds.
+                            block {
+                                // Mark this build step as FAILURE if at least one of the downstream builds were marked FAILED.
+                                buildStepFailure('FAILURE')
+                                // Mark this entire build as FAILURE if at least one of the downstream builds were marked FAILED.
+                                failure('FAILURE')
+                                // Mark this entire build as UNSTABLE if at least one of the downstream builds were marked UNSTABLE.
+                                unstable('UNSTABLE')
+                            }
+                            parameters {
+                                predefinedProp('ENVIRONMENT', env_config.get('ENVIRONMENTS_DEPLOYMENT'))
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+    }
+}
--- a/dataeng/jobs/analytics/RetirementJobs.groovy
+++ b/dataeng/jobs/analytics/RetirementJobs.groovy
@@ -1,0 +1,610 @@
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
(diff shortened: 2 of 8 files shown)


dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
```diff
@@ -0,0 +1,145 @@
+/* UserRetirementJobEdxTriggers.groovy
+ *
+ * Defines jobs which trigger downstream user retirement jobs for each edX
+ * environment.
+ */
+
+package analytics
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.jenkins_build_team_security
```
pwnage101, line 9: are you using this anywhere?

dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+                // keep jobs around for 30 days
+                // allVars contains the value for DAYS_TO_KEEP_BUILD which will be used inside AnalyticsConstants.common_log_rotator
+                logRotator common_log_rotator(allVars)
+
+                wrappers {
+                    // DENG-1063/PSRE-1204: Temporarily avoid using the build-name-setter
+                    // plugin while it's currently broken due to plugin incompatibilities.
+                    //buildName('#${BUILD_NUMBER}')
```
pwnage101, line 117: This comment no longer applies because we're moving to a new jenkins instance, so I think we can re-enable buildName.

dataeng/jobs/analytics/RetirementJobs.groovy
```diff
@@ -0,0 +1,725 @@
...
+                    failBuild()
+                }
+                buildUserVars() /* gives us access to BUILD_USER_ID, among other things */
+                // DENG-1063/PSRE-1204: Temporarily avoid using the build-name-setter
+                // plugin while it's currently broken due to plugin incompatibilities.
+                //buildName('#${BUILD_NUMBER}, ${ENV,var="ENVIRONMENT"}')
+                timestamps()
+                colorizeOutput('xterm')
+                credentialsBinding {
+                    file('USER_RETIREMENT_SECURE_DEFAULT', 'user-retirement-secure-default.yml')
```
pwnage101, line 231: I think we can do away with this credential. It hasn't been used in years, and I think it was only used for development.  I would just make the ENVIRONMENT parameter a required parameter with no default.

dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
```diff
@@ -0,0 +1,145 @@
...
+
+
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+
+class RetirementJobEdxTriggers{
+    public static def job = { dslFactory, allVars ->
+        List jobConfigs = [
```
pwnage101, line 20: Our current established pattern is to define job configs in analytics-secure under the job-configs directory---I think this case is no different than other jobs.  I also noticed that our authorization/permissions for these jobs were defined inside the job DSL, but I think we should also move those to the yaml job configs as is done in our other jobs.

jazibhumayun: Yes sure, I have done that change

dataeng/jobs/analytics/RetirementJobEdxTriggers.groovy
jazibhumayun: Alright sure, I have removed these lines

dataeng/jobs/analytics/RetirementJobs.groovy
jazibhumayun: Sorry I couldn't understand the second part of comment (`I would just make the ENVIRONMENT parameter a required parameter with no default`)
I removed the line `file('USER_RETIREMENT_SECURE_DEFAULT', 'user-retirement-secure-default.yml')` so it gave me error on cp command as its being used in shell files of the job like here: [https://github.com/edx/jenkins-job-dsl/blob/master/platform/resources/user-retirement-collector.sh#L26](https://github.com/edx/jenkins-job-dsl/blob/master/platform/resources/user-retirement-collector.sh#L26)

jazibhumayun: or can I just simply remove this [cp command](https://github.com/edx/jenkins-job-dsl/blob/master/platform/resources/user-retirement-collector.sh#L26) if this is not used anywhere ?

dataeng/resources/retirement-partner-report-cleanup.sh
```diff
@@ -0,0 +1,26 @@
...
+source "${VENV}/bin/activate"
+
+# prepare credentials
+mkdir -p $WORKSPACE/user-retirement-secure
+cp $USER_RETIREMENT_SECURE_DEFAULT $WORKSPACE/user-retirement-secure/secure-default.yml
+
+# prepare tubular
+cd $WORKSPACE/tubular
+# snapshot the current latest versions of pip and setuptools.
+pip install 'pip==21.0.1' 'setuptools==53.0.0'
```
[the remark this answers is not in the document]
jazibhumayun: @brianhw  sure, can do that.
Should I create a file in  [https://github.com/openedx/tubular](https://github.com/openedx/tubular) repo. Because generally the requirement.txt files goes into code repository.
What are your on thoughts on if I place these requirements in [https://github.com/openedx/tubular/blob/master/requirements/base.in](https://github.com/openedx/tubular/blob/master/requirements/base.in) ?

dataeng/jobs/analytics/RetirementJobs.groovy
pwnage101: Yes, I would say remove the cp command, remove the empty file credential, and remove the default value for the ENVIRONMENT parameter because all of the upstream trigger jobs will provide a value for ENVIRONMENT.

I just remembered why we added all of that logic---it was for the open source community!  However, the open source instances that have contacted us about user retirement haven't been using these jenkins jobs---they have written their own automation to call the tubular scripts directly.  I think we can narrow down our supported methods of supplying configuration to just what we currently do (fetch from analytics-secure).

dataeng/resources/retirement-partner-report-cleanup.sh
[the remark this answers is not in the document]
pwnage101: It seems like pip and possibly setuptools is considered "unsafe" to go into requirements.txt files: https://github.com/edx/tubular/blob/1552c897ad6d6423bd50bd1053fb95ce11b7d303/requirements/pip-tools.txt#L16-L17

dataeng/resources/retirement-partner-report-cleanup.sh
```diff
@@ -0,0 +1,26 @@
...
+# Create and activate a virtualenv.  In case we ever change the concurrency
+# setting on the jenkins worker, it would be safest to keep the builds from
+# clobbering each other's virtualenvs.
+VENV="venv-${BUILD_NUMBER}"
+virtualenv --python=python3.8 --clear "${VENV}"
+source "${VENV}/bin/activate"
+
+# prepare credentials
+# mkdir -p $WORKSPACE/user-retirement-secure
+# cp $USER_RETIREMENT_SECURE_DEFAULT $WORKSPACE/user-retirement-secure/secure-default.yml
```
pwnage101, line 14: I think these lines can be deleted now

pwnage101 (approved):

PR #1493, opened 2022-01-13
fix: update github org for codejail

jdmulloy:


testingPipelines/jobs/generalPRPipeline.groovy
```diff
@@ -111,7 +111,7 @@ jobConfigs.each { jobConfig ->
                         }
                         remote {
                             credentials('jenkins-worker')
-                            github("edx/${jobConfig.repoName}", 'ssh', 'github.com')
+                            github("openedx/${jobConfig.repoName}", 'ssh', 'github.com')
```
feanil, line 114:
```suggestion
                            github("${jobConfig.org}/${jobConfig.repoName}", 'ssh', 'github.com')
```

feanil (approved):
Looks good but we have a variable for this that we can use instead.

PR #1499, opened 2022-01-20
fix: set bash env for analytics exporter jobs

estute:
The exporter recently failed to do a pip install but kept running. This led to all of the worker jobs have incorrectly set up environments and eventually sending all partner orgs with empty data packages.


brianhw (approved):
LGTM.  👍

PR #1503, opened 2022-02-01
Install latest pip version in cleanup job

aht007:
## DESC
Previously a pinned version of pip(20.0.2) was being installed which was causing issues with repos that had been updated to latest pip versions.This PR fixes that issue and unpins the installation of pip

### DIFF
--- a/testeng/resources/cleanup-python-code-create-pr.sh
+++ b/testeng/resources/cleanup-python-code-create-pr.sh
@@ -33,7 +33,7 @@ do_one_repo () {
   source "$venv_dir/bin/activate"
 
   echo "Upgrading pip..."
-  pip install "pip==20.0.2"
+  pip install pip
 
   local PKG_SPECS
   IFS=", " read -ra PKG_SPECS <<<"$PACKAGES"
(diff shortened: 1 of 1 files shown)


awais786 (approved):

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -33,7 +33,7 @@ do_one_repo () {
   source "$venv_dir/bin/activate"
 
   echo "Upgrading pip..."
-  pip install "pip==20.0.2"
+  pip install "pip"
```
mraarif, line 36:
```suggestion
  pip install pip
```

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -59,6 +59,9 @@ do_one_repo () {
   echo "Running script to create PR..."
   cd "$WORKSPACE/testeng-ci"
   pip install -r requirements/base.txt
+  
+  pip --version
+  pip freeze | grep pip-tools
```
mraarif, line 64: we don't need this anymore?

PR #1504, opened 2022-02-02
feat: add remove data czar job

arbabkhalil:


devops/resources/remove-data-czar.sh
```diff
@@ -0,0 +1,20 @@
...
+. "$venvpath/bin/activate"
+set -u
+
+cd "$WORKSPACE/configuration"
+pip install -r util/jenkins/requirements.txt
+. util/jenkins/assume-role.sh
+
+assume-role ${ROLE_ARN}
+
+cd util/create_data_czar
```
syed-awais-ali, line 17: I guess, for the better naming convention we can use the data_czar folder in the configuration repo. That folder should have `create_data_czar` and `delete_data_czar` scripts. Once you land that change you may need to change these paths/names

syed-awais-ali (changes requested):
A minor comment, looks good otherwise

PR #1505, opened 2022-02-03
build: Install node in jenkins worker

Jawayria:


Jawayria:
This branch is being used in jenkins cleanup job since we worked on Node 16 upgrade.

testeng/resources/cleanup-python-code-create-pr.sh
```diff
@@ -3,6 +3,18 @@ set -eu -o pipefail
...
+  source ~/.bashrc;
+  source ~/.nvm/nvm.sh;
+  }
+
+nvm install 16
+nvm use 16
+
+node --version
+
 PACKAGES="${PACKAGES:-}"
```
iamsobanjaved, line 18: You have placed your code after an explanatory comment for the `PACKAGES="${PACKAGES:-}"` line. So it would be best if you moved your addition up or down to not separate the line from its explanation.

UsamaSadiq:
Since we’ve moved from `Jenkins cleanup job` to `GitHub bulk repo update job`, this PR shouldn’t be needed anymore.

PR #1509, opened 2022-02-06
fix: fix the vault login race condition

pwnage101:
We used to take the retrieved vault token, then pass it into `vault login` which persists the token in a shared path under the home directory.  That creates a race condition in which multiple jenkins builds could clobber each other's "/home/jenkins/.vault-token.tmp" files.  Furthermore, jenkins jobs could also end up using each-other's temporary vault tokens if the timing lines up.

This new approach keeps each vault login session isolated within each build environment by not relying on a file to persist the token, but rather an environment variable.


brianhw (approved):
LGTM.  👍

PR #1512, opened 2022-02-09
chore: add update data czar job

arbabkhalil:


syed-awais-ali (commented):
Does this handle the case where we only need to update the keys of the data czar? Referring to the last section of the [doc](https://openedx.atlassian.net/wiki/spaces/SRE/pages/33358045/How+to+Handle+Data+Czar+Requests)

arbabkhalil:
> Does this handle the case where we only need to update the keys of the data czar? Referring to the last section of the [doc](https://openedx.atlassian.net/wiki/spaces/SRE/pages/33358045/How+to+Handle+Data+Czar+Requests)

yes it does as it first removes the user then create new credentials, providing the same email would fulfil that usecase.

PR #1523, opened 2022-03-07
refactor: Removed rest of requirements upgrader

timmc-edx:
This script was only used by the upgrader job, which was removed in PR #1464


rgraber (approved):
LGTM

PR #1533, opened 2022-03-21
fix: don't persist token locally

Hamza442:


dataeng/resources/prefect-flows-deployment.sh
```diff
@@ -24,7 +24,7 @@ aws ecr describe-repositories --repository-names $FLOW_NAME --region us-east-1 |
 vault write -field=token auth/approle/login \
   role_id=${ANALYTICS_VAULT_ROLE_ID} \
   secret_id=${ANALYTICS_VAULT_SECRET_ID} \
-| vault login -no-print token=-
+| vault login -no-store token=-
```
jazibhumayun, line 27: so its something we can not opt for `-no-print` with `-no-store` option ? https://www.vaultproject.io/docs/commands/login
It will be a trade off between no to store and avoid race condition

PR #1540, opened 2022-03-29
fix: order orgs for analytics-exporter jobs

estute:


dataeng/resources/setup-exporter.sh
```diff
@@ -45,3 +45,17 @@ exporter-properties \
...
+# Some orgs can take an exceptionally long time to run. Depending on the concurrency
+# settings for the analytics-exporter-master job and the location of the organization
+# alphabetically in the organizations directory, it's possible that these long-running
+# jobs will be started towards the end of the master run, which can extend the total
+# run-time quite a bit. Use PRIORITY_ORGS to pass a space separated list of orgs that
+# should run first. This is accomplished by prepending a number to the name of the orgs
+# in question.
+for ORG in ${PRIORITY_ORGS}; do
+    if [ -f organizations/$ORG ]; then
+        mv organizations/$ORG organizations/1_$ORG
```
AbdouSeck, line 59: @estute does the `edx-analytics-exporter` code sort the files in the organizations directory lexicographically? [This function](https://github.com/openedx/edx-analytics-exporter/blob/c2e3b80af2431aec5eb6e797d2d4f27a2eeb9fce/exporter/config.py#L20) seems to be the one loading the organizations, but it's not doing any sorting.

AbdouSeck (commented):
Please see inline comments.

dataeng/resources/setup-exporter.sh
pwnage101: from here, isn't it jenkins that reads the files in the organizations folder?  I think this is the code which reads the files written in that folder: https://github.com/edx/jenkins-job-dsl/blob/29ded3bea37a8c45305ddf37ddf9eb6883e01bb3/dataeng/jobs/analytics/AnalyticsExporter.groovy#L260

If they're currently run alphabetically, then this PR would probably work.

pwnage101 (approved):

dataeng/resources/setup-exporter.sh
estute: Yes, this is correct. The plugin we are using uses a glob pattern to list the files in the `organizations` directory, passing each file to a downstream job.

AbdouSeck: Great.

PR #1542, opened 2022-04-04
chore: adding insights snowflake aurora flow

jazibhumayun:


estute (approved):
Thanks for catching this!

PR #1543, opened 2022-04-05
feat: tableau config backup

jazibhumayun:
This jenkins job will ssh into Tableau instance and create config backup. Further a python script serialize the json object and compares baseline config with newly created backup config. If it finds differences it creates a pull request to update baseline config.

### DIFF
--- a/dataeng/jobs/analytics/TableauConfigBackup.groovy
+++ b/dataeng/jobs/analytics/TableauConfigBackup.groovy
@@ -1,0 +1,57 @@
+package analytics
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+class TableauConfigBackup{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("tableau-config-backup"){
+            description(
+                "This jenkins job will ssh into Tableau instance and create config backup. " +
+                "Further a python script serialize the json object and compares baseline config with newly created backup config. " +
+                "If it finds differences it creates a pull request to update baseline config. "
+            )
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            multiscm secure_scm(allVars)
+            parameters {
+                stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), , 'Branch of analytics tools repo to use.')
+                stringParam('TABLEAU_SERVER_DNS', allVars.get('TABLEAU_SERVER_DNS'), 'DNS address of Tableau Server.')
+                stringParam('USER_NAME', allVars.get('USER_NAME'), 'User name to be used for ssh into instance.')
+                stringParam('S3_BUCKET', allVars.get('S3_BUCKET'), 'S3 backup bucket name with path. This bucket will be used to store rotating config backup for 60 days.')
+                stringParam('NOTIFY', allVars.get('NOTIFY','$PAGER_NOTIFY'), 'Space separated list of emails to send notifications to.')
+            }
+            triggers common_triggers(allVars)
+            wrappers {
+                colorizeOutput('xterm')
+                timestamps()
+                credentialsBinding {
+                    string('GITHUB_TOKEN', 'GHPRB_BOT_TOKEN');
+                }
+            }
+            multiscm secure_scm(allVars) << {
+                git {
+                    remote {
+                        url('$ANALYTICS_TOOLS_URL')
+                        branch('$ANALYTICS_TOOLS_BRANCH')
+                        credentials('1')
+                    }
+                    extensions {
+                        relativeTargetDirectory('analytics-tools')
+                        pruneBranches()
+                        cleanAfterCheckout()
+                    }
+                }
+            }
+            wrappers common_wrappers(allVars)
+            publishers common_publishers(allVars)
+            steps {
+                shell(dslFactory.readFileFromWorkspace('dataeng/resources/tableau-config-backup.sh'))
+            }
+        }
+    }
+}
--- a/dataeng/resources/tableau-config-backup.sh
+++ b/dataeng/resources/tableau-config-backup.sh
@@ -1,0 +1,72 @@
+# This script is used to ssh into tableau instance, create config backup and store in s3. It also compares latest created backup with a
+# stored baseline backup to detect any changes and creates a PR for DE members to review and approve if changes seems correct.
+#!/usr/bin/env bash
+set -ex
+
+# It gets the current date from system in 2022.04.05 format
+DATE="$(date '+%Y.%m.%d')"
+
+# It adds the date as prefix on backup name. Backup runs at 09:00 UTC so it also prefixed in the name.
+TABLEAU_CONFIG="${DATE}-tableau-config-backup.json"
+
+ssh -tt $USER_NAME@$TABLEAU_SERVER_DNS -o StrictHostKeyChecking=no "S3_BUCKET=$S3_BUCKET TABLEAU_CONFIG=$TABLEAU_CONFIG" '
+
+sudo runuser -l tsm_admin -c "tsm settings export -f /home/tableau_backups/tableau-config-backup.json"
+
+# Check if tableau backup file exist, exit if the file does not exist
+if [[ ! -f /home/tableau_backups/tableau-config-backup.json ]]
+then
+    echo "$TABLEAU_CONFIG Backup file does not exist. It seems to be some problem generating tableau backup. Refer to Administration of Tableau wiki docs to debug Tableau Regular Backup Automation"
+    exit 1
+else
+    echo "Copying backup file into s3"
+    aws s3 cp /home/tableau_backups/tableau-config-backup.json s3://$S3_BUCKET/$TABLEAU_CONFIG
+
+    echo "Removing backup file from local disk"
+    sudo rm -f /home/tableau_backups/$TABLEAU_CONFIG
+
+fi
+
+'
+
+# Creating python 3.8 virtual environment to run python script.
+PYTHON38_VENV="py38_venv"
+virtualenv --python=python3.8 --clear "${PYTHON38_VENV}"
+source "${PYTHON38_VENV}/bin/activate"
+
+# Copying tableau-config-backup file from s3.
+aws s3 cp s3://$S3_BUCKET/$TABLEAU_CONFIG tableau-config.json
+
+pip install deepdiff click
+
+# Disabling the exit on failure flag
+set +e
+
+# Comparing current tableau-config.json file with stored backup and generates an updated config file if changes are detected
+python $WORKSPACE/analytics-tools/tableau/tableau_config_compare.py --input_config 'tableau-config.json' --baseline_config "$WORKSPACE/analytics-secure/tableau/tableau-config-backup-base.json" --output_file 'updated-tableau-config.json'
+
+if [ $? -eq 0 ]; then
+    echo "Tableau Config Backup is same"
+else
+    echo "Changes are detected in Tableau config so it is creating a pull request to update baseline backup file."
+    set -e
+    cp updated-tableau-config.json $WORKSPACE/analytics-secure/tableau/tableau-config-backup-base.json
+
+    cd $WORKSPACE/analytics-secure/
+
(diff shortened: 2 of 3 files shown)


dataeng/jobs/analytics/TableauConfigBackup.groovy
```diff
@@ -0,0 +1,52 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_log_rotator
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_wrappers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_publishers
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+class TableauConfigBackup{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("tableau-config-backup"){
```
estute, line 11: All new jobs should have a description. Can you add one please?

dataeng/jobs/analytics/TableauConfigBackup.groovy
```diff
@@ -0,0 +1,52 @@
...
+            logRotator common_log_rotator(allVars)
+            parameters secure_scm_parameters(allVars)
+            multiscm secure_scm(allVars)
+            parameters {
+                stringParam('ANALYTICS_TOOLS_URL', allVars.get('ANALYTICS_TOOLS_URL'), 'URL for the analytics tools repo.')
+                stringParam('ANALYTICS_TOOLS_BRANCH', allVars.get('ANALYTICS_TOOLS_BRANCH'), , 'Branch of analytics tools repo to use.')
+                stringParam('TABLEAU_SERVER_IP', allVars.get('TABLEAU_SERVER_IP'), 'IP address of Tableau Server.')
```
estute, line 18: I can't recall, did you say there was a reason why we can't use a hostname? Unless we can't I think it's preferable.

dataeng/jobs/analytics/TableauConfigBackup.groovy
```diff
@@ -0,0 +1,52 @@
...
+                stringParam('USER_NAME', allVars.get('USER_NAME'), 'SSH User name.')
+                stringParam('S3_BUCKET', allVars.get('S3_BUCKET'), 'S3 backup bucket name with path.')
```
estute, line 20: For the last few params, can you give a more helpful description of what the parameter is used for?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+
+ssh -tt $USER_NAME@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "S3_BUCKET=$S3_BUCKET" '
+
+# It gets the current date from system in 2022.04.05-10:35 format
+DATE="$(date '+%Y.%m.%d')"
+
+# It adds the date as prefix on backup name. Backup runs at 09:00 UTC so it also prefixed in the name.
+BACKUPNAME="${DATE}-tableau-config-backup.json"
+
+sudo runuser -l tsm_admin -c "tsm settings export -f /home/tableau_backups/tableau-config-backup.json"
```
estute, line 12: shouldn't you be exporting to $BACKUPNAME?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Check if tableau backup file exist, exit if the file does not exist
+if [[ ! -f /home/tableau_backups/tableau-config-backup.json ]]
```
estute, line 15: same issue with $BACKUPNAME

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
estute, line 1: Can you add a description at the top of this script indicating what it is used for and how to invoke it?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+    # Changes are detected in Tableau config so it will create a pull request to update baseline backup file.
+    echo "Tableau Config Backups are modified"
+
+    cp serialized-tableau-config.json $WORKSPACE/analytics-secure/tableau/tableau-config-backup-base.json
+
+    cd $WORKSPACE/analytics-secure/
+
+    # Create a new git branch
+    now=$(date +%Y_%m_%d_%H_%M)
+    branchname="tableau_config_$now"
```
estute, line 60: nit: can you standardize on CAPs for bash variables?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+
+'
+
+# Creating python 3.8 virtual environment to run python script.
+PYTHON38_VENV="py38_venv"
+virtualenv --python=python3.8 --clear "${PYTHON38_VENV}"
+source "${PYTHON38_VENV}/bin/activate"
+
+DATE="$(date '+%Y.%m.%d')"
+TABLEAU_CONFIG=${DATE}-tableau-config-backup.json
```
estute, line 36: This is defined twice. Maybe just define it once and pass it into the ssh script as a variable?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Copying tableau-config-backup file from s3.
+aws s3 cp s3://$S3_BUCKET/$TABLEAU_CONFIG tableau-config.json
+
+# Copying tableau-config-json-serializer from analytics-tolls repository to local repo.
+cp $WORKSPACE/analytics-tools/tableau/tableau-config-json-serializer.py tableau-config-json-serializer.py
```
estute, line 42: Why do you need to copy the script?

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# This script deletes the json object which shows non-relevant change.
+python tableau-config-json-serializer.py
+
+# Comparing both base line config file with the latest backup obtained from tableau.
+if cmp -s "$WORKSPACE/analytics-secure/tableau/tableau-config-backup-base.json" "serialized-tableau-config.json"; then
```
estute, line 48: Can you be certain the order of the serialized json file will be the same each time? You script serializes a json file into a Python dict, which after Python 3.6 enforces insertion order, but I'm not sure if that is the case here. I think it is, but you might want to check.

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "Tableau Config Backups are same"
+else
+    # Changes are detected in Tableau config so it will create a pull request to update baseline backup file.
+    echo "Tableau Config Backups are modified"
```
estute, line 52: I would recommend moving this comment into the logging/echo as we will be reading the Jenkins logs more often than the code.

dataeng/resources/tableau-config-backup.sh
```diff
@@ -0,0 +1,73 @@
...
+sudo runuser -l tsm_admin -c "tsm settings export -f /home/tableau_backups/tableau-config-backup.json"
+
+# Check if tableau backup file exist, exit if the file does not exist
+if [[ ! -f /home/tableau_backups/tableau-config-backup.json ]]
+then
+    echo "$BACKUPNAME Backup file does not exist. It seems to be some problem generating tableau backup. Refer to Administration of Tableau wiki docs to debug Tableau Regular Backup Automation"
+    exit 1
+else
+    echo "Copying backup file into s3"
+    aws s3 cp /home/tableau_backups/tableau-config-backup.json s3://$S3_BUCKET/$BACKUPNAME
```
estute, line 21: Rather than using s3 as an intermediary, what do you think about doing the following:

1. ssh into the box
2. if there is a backupfile present (at all) delete it
3. export the backup file with the tableau command
4. back on Jenkins, scp the file over

You are already using ssh, so you should be able to use the same credentials, and it removes the need to use a bucket. I don't know how useful this refactor would be. Just an idea.

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Yes, I tried doing it. Since this command has to run as `tsm_admin` (user which is used to run tsm commands). So exporting an environment variable in user  and accessing into another user required additional steps. That's why I thought to simply write a name of backup and append date while I am uploading it on s3. There is only one local copy of the backup so doesn't need to be appended with `date/time`

dataeng/jobs/analytics/TableauConfigBackup.groovy
jazibhumayun: Yes in case of jenkins and dbt-docs server, we can ssh into server using their DNS name. But in case of Tableau [tableau.analytics.edx.org](https://tableau.analytics.edx.org/) its Route53 DNS record points to public load balancer. This load balancer only allow traffic port on 80, 8080, 8850, 8060 which are the access points of Tableau and Tableau Server Manager. I am also in favor of not exposing port 22(ssh port) on public load balancer. @estute  

image

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Yes, order will remain same I don't think Tableau will change the order while creating backups I have checked by creating backups.

dataeng/resources/tableau-config-backup.sh
jazibhumayun: The reason I choose to copy the files on s3 was, it will keep the backup files for 60 days as we have life cycle configured for s3 bucket. So if we want to back track on the config files we will have config backups stored for last 60 days (1 backup each week = 8 backup files for last 8 weeks)

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Because I want to run it in the current directory.

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Infact I realized the lifecycle rule was not attached on `config-backups/`. Created PR now: [https://github.com/edx/terraform/pull/4489](https://github.com/edx/terraform/pull/4489)

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Done that

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Thanks, fixed it.

dataeng/resources/tableau-config-backup.sh
jazibhumayun: Done it. thanks

dataeng/resources/tableau-config-backup.sh
jazibhumayun: fixed it, thanks

dataeng/jobs/analytics/TableauConfigBackup.groovy
jazibhumayun: Fixed it.

dataeng/jobs/analytics/TableauConfigBackup.groovy
jazibhumayun: Done that, thanks for pointing

dataeng/jobs/analytics/TableauConfigBackup.groovy
```diff
@@ -0,0 +1,57 @@
...
+import static org.edx.jenkins.dsl.AnalyticsConstants.common_triggers
+import static org.edx.jenkins.dsl.AnalyticsConstants.secure_scm_parameters
+
+class TableauConfigBackup{
+    public static def job = { dslFactory, allVars ->
+        dslFactory.job("tableau-config-backup"){
+            description(
+                "This jenkins job will ssh into Tableau instance and create config backup." +
+                "Further a python script serialize the json object and compares baseline config with newly created backup config." +
+                "If it finds differences it creates a pull request to update baseline config."
```
pwnage101, line 15: I believe this needs spaces at the end of each line, like this:
```suggestion
                "This jenkins job will ssh into Tableau instance and create config backup. " +
                "Further a python script serialize the json object and compares baseline config with newly created backup config. " +
                "If it finds differences it creates a pull request to update baseline config. "
```

dataeng/jobs/analytics/TableauConfigBackup.groovy
pwnage101: Ideally the load balancer itself would also use a DNS record to resolve the IP address of the backend service (tableau).  Is that also not the case?

dataeng/resources/tableau-config-backup.sh
pwnage101: can't you just invoke it remotely?  like this:

```bash
python $WORKSPACE/analytics-tools/tableau/tableau-config-json-serializer.py --input_file ...
```

dataeng/resources/tableau-config-backup.sh
pwnage101: tableau may not change the order, but will python serialize ordered dicts deterministically?  Is that guaranteed in the python 3.8 documentation?

AbdouSeck (approved):

PR #1544, opened 2022-04-07
feat: adding token helper and vault-config

Hamza442:
The Following changes have been made to avoid race condition in analytics exporter job.
1 - A token helper with name token-helper.py has been added to resources folder in Jenkins job dsl repo, this token helper allow us to create a separate token for each worker job.
2 - A vault-config.sh file has been added this file will help to setup configuration for vault cli to use the token helper (token-helper.py)
3 - Since token-helper.py is added to resources folder of Jenkins-job-dsl repo. Need to clone this repo to get the token helper, need to provide exact path  for token helper to enable vault cli to locate it

### DIFF
--- a/dataeng/jobs/analytics/AnalyticsExporter.groovy
+++ b/dataeng/jobs/analytics/AnalyticsExporter.groovy
@@ -99,7 +99,7 @@ class AnalyticsExporter {
                 stringParam('ORG')
                 stringParam('PLATFORM_VENV')
                 stringParam('EXTRA_OPTIONS')
-            }
+                stringParam('JOB_DSL_BRANCH','master', 'Branch to use for the jenkins-job-dsl repository.')            }
             parameters secure_scm_parameters(allVars)
 
             environmentVariables {
@@ -124,7 +124,19 @@ class AnalyticsExporter {
 
             concurrentBuild()
 
-            multiscm secure_scm(allVars)
+            multiscm secure_scm(allVars) << {
+                git {
+                    remote {
+                        url('git@github.com:edx/jenkins-job-dsl.git')
+                        branch('$JOB_DSL_BRANCH')
+                        credentials('1')
+                    }
+                    extensions {
+                        pruneBranches()
+                        relativeTargetDirectory('jenkins-job-dsl')
+                    }
+                }
+            }
 
             wrappers {
                 timestamps()
@@ -177,7 +189,8 @@ class AnalyticsExporter {
                 stringParam('ORG_CONFIG', 'data-czar-keys/config.yaml', 'Path to the data-czar organization config file.')
                 stringParam('DATA_CZAR_KEYS_BRANCH', 'master', 'Branch to use for the data-czar-keys repository.')
                 stringParam('PRIORITY_ORGS', allVars.get('PRIORITY_ORGS'), 'Space separated list of organizations to process first.')
-                stringParam('JOB_DSL_BRANCH', 'origin/master', 'Branch from the jenkins job dsl repository to get vault token helper.')
+                
+
             }
             parameters secure_scm_parameters(allVars)
             environmentVariables {
@@ -219,7 +232,7 @@ class AnalyticsExporter {
                         relativeTargetDirectory('data-czar-keys')
                     }
                 }
-
+                
             }
 
             triggers{
--- a/dataeng/jobs/analytics/AnalyticsExporter.groovy
+++ b/dataeng/jobs/analytics/AnalyticsExporter.groovy
@@ -99,7 +99,8 @@ class AnalyticsExporter {
                 stringParam('ORG')
                 stringParam('PLATFORM_VENV')
                 stringParam('EXTRA_OPTIONS')
-                stringParam('JOB_DSL_BRANCH','master', 'Branch to use for the jenkins-job-dsl repository.')            }
+                stringParam('JOB_DSL_BRANCH','master', 'Branch to use for the jenkins-job-dsl repository.')            
+            }
             parameters secure_scm_parameters(allVars)
 
             environmentVariables {
@@ -129,7 +130,6 @@ class AnalyticsExporter {
                     remote {
                         url('git@github.com:edx/jenkins-job-dsl.git')
                         branch('$JOB_DSL_BRANCH')
-                        credentials('1')
                     }
                     extensions {
                         pruneBranches()
@@ -189,8 +189,6 @@ class AnalyticsExporter {
                 stringParam('ORG_CONFIG', 'data-czar-keys/config.yaml', 'Path to the data-czar organization config file.')
                 stringParam('DATA_CZAR_KEYS_BRANCH', 'master', 'Branch to use for the data-czar-keys repository.')
                 stringParam('PRIORITY_ORGS', allVars.get('PRIORITY_ORGS'), 'Space separated list of organizations to process first.')
-                
-
             }
             parameters secure_scm_parameters(allVars)
             environmentVariables {
--- a/dataeng/jobs/analytics/AnalyticsExporter.groovy
+++ b/dataeng/jobs/analytics/AnalyticsExporter.groovy
@@ -17,7 +17,6 @@ class AnalyticsExporter {
                 stringParam('NOTIFY', '', 'Space separated list of emails to notify in case of failure.')
                 stringParam('DATE_MODIFIER', '', 'Used to set the date of the CWSM dump.  Leave blank to use today\'s date.  Set to "-d 202x-0x-0x" if that is when the CWSM dump took place.  (Leave off quotes.)')
                 stringParam('TASKS', '', 'Space separated list of tasks to process. Leave this blank to use the task list specified in the config file.  Specify here only if you are running tests of a specific task.')
-                stringParam('JOB_DSL_BRANCH', 'origin/master', 'Branch from the jenkins job dsl repository to get vault token helper.')
             }
             parameters secure_scm_parameters(allVars)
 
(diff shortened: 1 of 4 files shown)


dataeng/jobs/analytics/AnalyticsExporter.groovy
```diff
@@ -169,6 +177,7 @@ class AnalyticsExporter {
                 stringParam('ORG_CONFIG', 'data-czar-keys/config.yaml', 'Path to the data-czar organization config file.')
                 stringParam('DATA_CZAR_KEYS_BRANCH', 'master', 'Branch to use for the data-czar-keys repository.')
                 stringParam('PRIORITY_ORGS', allVars.get('PRIORITY_ORGS'), 'Space separated list of organizations to process first.')
+                stringParam('JOB_DSL_BRANCH', 'origin/master', 'Branch from the jenkins job dsl repository to get vault token helper.')
```
estute, line 180: This is not used. When the job is run, it will not clone the repo.

dataeng/resources/vault-config.sh
```diff
@@ -0,0 +1,15 @@
...
+# make a directory to store vault token helper and config file
+mkdir ${WORKSPACE}/vault-config
+# get token helper from repo
+cp ${WORKSPACE}/jenkins-job-dsl/dataeng/resources/token-helper.py ${WORKSPACE}/vault-config/token-helper.py
+# remove the repo after getting token helper
+rm -rf ${WORKSPACE}/jenkins-job-dsl
+# create path for helper
+path_to_helper = ${WORKSPACE}"/vault-config/token-helper.py"
+# set up config for config file
+token_config = "$(echo token_helper = \"$path_to_helper\" )"
```
estute, line 13: I don't think you need to use command substitution here. I think it can be as simple as: `token_config="token_helper=$path_to_helper"`

dataeng/resources/vault-config.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+
+# make a directory to store vault token helper and config file
+mkdir ${WORKSPACE}/vault-config
+# get token helper from repo
+cp ${WORKSPACE}/jenkins-job-dsl/dataeng/resources/token-helper.py ${WORKSPACE}/vault-config/token-helper.py
+# remove the repo after getting token helper
+rm -rf ${WORKSPACE}/jenkins-job-dsl
```
estute, line 9: I don't think you need to delete any of the repos you clone

dataeng/resources/vault-config.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+
+# make a directory to store vault token helper and config file
+mkdir ${WORKSPACE}/vault-config
+# get token helper from repo
+cp ${WORKSPACE}/jenkins-job-dsl/dataeng/resources/token-helper.py ${WORKSPACE}/vault-config/token-helper.py
```
estute, line 7: Instead of copying it, you can just use the path to the file in the repo and save yourself a step

dataeng/resources/vault-config.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
estute, line 1: I would add some documentation in this script to explain why we are doing this, not just what we are doing. It was difficult to deduce what was causing our errors in the first place, and having docs to explain our reasoning will be useful for anyone who needs to use it in the future.

dataeng/resources/remote-config.sh
```diff
@@ -91,11 +91,18 @@ unassume_role
...
+# set path for config file which store token in each job workspace
+export VAULT_CONFIG_PATH= ${WORKSPACE}/vault-config/vault_config
+
+# write credentials to vault server to get the token
 vault write -field=token auth/approle/login \
     role_id=${ANALYTICS_VAULT_ROLE_ID} \
     secret_id=${ANALYTICS_VAULT_SECRET_ID} \
 | vault login -no-print token=-
 
+# set vault token
```
estute, line 103: Maybe add a line here explaining why we have to do this bit, because it isn't in the vault documentation. It's a hack to get around vault not working the way we expected.

dataeng/resources/token-helper.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python
+
+"""
+A vault token helper to change the default location for vault token.
+By default vault writes token to home direcotry of the server. This
+creates a race condition when multiple jobs try to access the token
+in parallel. In order to avoid it generating a separate token for 
+each job in its workspace.
```
estute, line 8: I would add a link to the vault documentation on token helpers here.

dataeng/resources/vault-config.sh
```diff
@@ -0,0 +1,15 @@
...
+
+
+# make a directory to store vault token helper and config file
+mkdir ${WORKSPACE}/vault-config
```
estute, line 5: Is there any risk that a job could re-use a workspace that already has a vault-config directory? What if a previous job failed before it performed a clean up? I can't remember if any of our jobs re-use work spaces.

dataeng/resources/token-helper.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python
```
estute, line 1: By specifying this shebang, you are explicitly pointing it to the system python binary, not the one specified in the virtualenv. This may cause errors.

dataeng/resources/vault-config.sh
Hamza442: Actually I need to enclose path_to_helper in double quotes that why a substitution is required. If I do  token_config="token_helper=$path_to_helper" output will be 
token_helper = /vault-config/token-helper.py but I need output as 
token_helper = "/vault-config/token-helper.py"

dataeng/resources/vault-config.sh
Hamza442: Just for clarification each build is a separate job ?

dataeng/resources/remote-config.sh
```diff
@@ -91,11 +91,29 @@ unassume_role
 #         usernamePassword('ANALYTICS_VAULT_ROLE_ID', 'ANALYTICS_VAULT_SECRET_ID', 'analytics-vault');
 #     }
 # }
+# set path for config file which store token in each job workspace
+export VAULT_CONFIG_PATH= ${WORKSPACE}/vault-config/vault_config
```
AbdouSeck, line 95: Just a heads-up that the space after the equal sign is not valid bash or sh syntax.

dataeng/resources/vault-config.sh
estute: In Jenkins terms, each time you run a job, that is called a "build". Each build gets it's own unique BUILD_ID and namespace (separate environment variables/parameters). Depending on how a job is set up, it can either have its own isolated workspace or share one. That's what I don't remember in this case.

dataeng/jobs/analytics/AnalyticsExporter.groovy
```diff
@@ -123,7 +124,19 @@ class AnalyticsExporter {
 
             concurrentBuild()
 
-            multiscm secure_scm(allVars)
+            multiscm secure_scm(allVars) << {
+                git {
+                    remote {
+                        url('git@github.com:edx/jenkins-job-dsl.git')
+                        branch('$JOB_DSL_BRANCH')
+                        credentials('1')
```
estute, line 132: No need for credentials, this is a public repo

dataeng/jobs/analytics/AnalyticsExporter.groovy
```diff
@@ -169,6 +189,8 @@ class AnalyticsExporter {
                 stringParam('ORG_CONFIG', 'data-czar-keys/config.yaml', 'Path to the data-czar organization config file.')
                 stringParam('DATA_CZAR_KEYS_BRANCH', 'master', 'Branch to use for the data-czar-keys repository.')
                 stringParam('PRIORITY_ORGS', allVars.get('PRIORITY_ORGS'), 'Space separated list of organizations to process first.')
+                
+
```
estute, line 193: nit: could you remove this whitespace?

dataeng/resources/token-helper.py
```diff
@@ -0,0 +1,61 @@
...
+        erase(PATH)
+
+    elif args == "store":
+        store(PATH)
+        
+    else:
+        print("Unknown method {}".format(args))
+        exit(1)
+
+
```
estute, line 61: nit: for readability, could you clean up some of the whitespace in this file?

dataeng/jobs/analytics/AnalyticsExporter.groovy
```diff
@@ -98,7 +99,7 @@ class AnalyticsExporter {
                 stringParam('ORG')
                 stringParam('PLATFORM_VENV')
                 stringParam('EXTRA_OPTIONS')
-            }
+                stringParam('JOB_DSL_BRANCH','master', 'Branch to use for the jenkins-job-dsl repository.')            }
```
estute, line 102: nit: can you fix the indent here?

dataeng/resources/vault-config.sh
Hamza442: I have checked from server there are multiple workspaces created for each worker such as 
analytics-exporter-worker@3
analytics-exporter-worker@4
analytics-exporter-worker@5
analytics-exporter-worker@6
after this segregation is there still any chance that workspace could be reused ??

estute (approved):
:+1:  This should get things working again, but pulling down the DSL repo during run time feels strange to me. I'd be interested to see if @pwnage101 has an opinion on how to manage this. The basic problem is: our usage of vault is fairly complex and requires an additional script and config file. However, it doesn't make sense to store these with the exporter code, as it is generic and can be applied to other jobs using vault.

PR #1545, opened 2022-04-12
chore: Extend answer distribution by another year.

HassanJaveed84:


estute (approved):
LGTM. Can we proactively add regexes for future years to avoid this next time?

PR #1550, opened 2022-04-18
chore: fix for vault race condition.

Hamza442:


dataeng/resources/remote-config.sh
```diff
@@ -92,12 +92,31 @@ unassume_role
 #     }
 # }
 
+# set up token path
+export VAULT_TOKEN_PATH=${WORKSPACE}/vault-config/vault-token
+# set path for config file which store token in each job workspace
+export VAULT_CONFIG_PATH=${WORKSPACE}/vault-config/vault_config
```
AbdouSeck, line 98: This config file is where the `token_helper` config is defined? How is the config file generated? I am not seeing it tracked in the git history.

AbdouSeck (changes requested):

dataeng/resources/remote-config.sh
Hamza442: This is generate from vault-config.sh file

Hamza442: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/resources/vault-config.sh

dataeng/resources/token-helper.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
AbdouSeck, line 1: This file needs to be executable. `chmod +x token-helper.py` should take care of that.

AbdouSeck (changes requested):

adzuci (approved):
LGTM

dataeng/resources/token-helper.py
Hamza442: Earlier we didn't made this executable still it was getting executed referring to this one http://jenkins.analytics.edx.org/job/analytics-exporter-worker/17946/console I am wondering do we need this in this case

PR #1559, opened 2022-05-06
feat: Add w-t support for cautious indirect test selection

pwnage101:
For running ampltiude enrichment models, I need to be able to run dbt
tests without invoking all the relationship tests which fail due to the
related models not existing in snowflake.  So far This has only been a
problem for testing amplitude enrichment models because it's the only
job which invokes a sub-set of the DAG inside an isolated schema, as
well as using the plus prefix `+` in front of the model selector.

This hasn't been a problem with CI tests because CI does not add the
plus prefix---it's more specific about exactly which models to run tests
for.

--indirect-selection=cautious is an effective solution because it
un-selects the failing relationship tests because they don't have ALL
related models selected in the original model selector.

DENG-1216


AbdouSeck (approved):
Looks good to me. Thanks for removing the unset `$exclude_param` variable from the `run` command.

PR #1560, opened 2022-05-09
feat: dbt monte carlo integration

Hamza442:
Clonning jenkins-job-dsl to use token-helper.py. vault-config.sh file is added so that each job creates a token in its own workspace.


dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -44,6 +45,16 @@ class WarehouseTransforms{
...
+                    git {
+                        remote {
+                            url('git@github.com:edx/jenkins-job-dsl.git')
+                            branch('$JOB_DSL_BRANCH')
+                        }
+                        extensions {
+                            pruneBranches()
+                            relativeTargetDirectory('jenkins-job-dsl')
+                        }
+                    }
```
HassanJaveed84, line 57: I don't think we need this as we're already reading rest of the scripts from the DSL when we seed/create the job.

dataeng/resources/warehouse-transforms.sh
```diff
@@ -65,6 +72,51 @@ fi
...
+# not working as expected and is not able to locate the new token location
+# Have to explicitly store token in token environment variable so that
+# vault cli can use the newly generated token for each job
+
+export VAULT_TOKEN="$(cat ${WORKSPACE}/vault-config/vault-token)"
+
+# set monte carlo api keys to integrate with monte carlo
+for KEY in ID TOKEN; do
+
+    export MCD_DEFAULT_API_${KEY}="$(vault kv get -version=1 -field=MCD_DEFAULT_API_${KEY} kv/monte-carlo-api-keys)"
```
HassanJaveed84, line 105: We already have the monte carlo API key in third_part_apis path in the vault, let's use that.

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -27,9 +27,11 @@ class WarehouseTransforms{
                     stringParam('SKIP_TESTS', env_config.get('SKIP_TESTS', 'false'), 'Set to \'true\' to skip all tests. All other values will allow tests to run given the other configuration values.')
                     stringParam('SKIP_SEED', env_config.get('SKIP_SEED', 'false'), 'Set to \'true\' to skip the dbt seed step. All other values will cause the seed step to be run.')
                     stringParam('TEST_SOURCES_FIRST', env_config.get('TEST_SOURCES_FIRST', 'true'), 'Set to \'true\' to perform source testing first (if SKIP_TESTS is false). All other values test sources post-run.')
+                    booleanParam('CAUTIOUS_INDIRECT_SELECTION', env_config.get('CAUTIOUS_INDIRECT_SELECTION', false), 'Check this box if you want a test to run if and only if ALL the models associated with that test have been selected (see dbt docs for --indirect-selection=cautious).')
```
HassanJaveed84, line 30: Curious why we're seeing these changes here, I thought they were already merged to master ?

PR #1566, opened 2022-05-16
feat: remove answer distribution job

Zacharis278:
@edx/masters-devs-cosmonauts 

Removes answer distribution job [MST-1412](https://2u-internal.atlassian.net/browse/MST-1412?atlOrigin=eyJpIjoiNTNlZGI5YTc0ZjUzNGFjNWE2NzhjNTRiMzM0NjdkYjUiLCJwIjoiaiJ9)

internal: https://github.com/edx/jenkins-job-dsl-internal/pull/656
remove config: https://github.com/edx-ops/analytics-secure/pull/667


AbdouSeck:
@Zacharis278 the job can be disabled without any changes to or deletion of the groovy code. [This line from the groovy script](https://github.com/edx/jenkins-job-dsl/blob/35946a560a077a65e23f796999791af81ad098e8/dataeng/jobs/analytics/AnswerDistribution.groovy#L15) expects a configuration value from `DISABLED` from analytics-secure. Happy to discuss this offline. But overall, it should be enough to disable the job.

Thanks,
Abdou

Zacharis278:
@AbdouSeck we've totally migrated off the output of these jobs at this point, they will no longer be used. For the previous ones we had disabled before removing but figured I'd save us a step this time. IMO these should eventually be removed for the sake of not keeping dead code around.

In any case, I had planned to open to ticket for this. I'll do that now.

AbdouSeck (approved):

PR #1570, opened 2022-05-23
chore: Increase timeout threshold on sandbox creation

marekwrobel:
Sandboxes created for prospectus may be taking more than 2h and sometimes they fail because they time out at 2h. 

This PR increases the threshold to 3.5h to give the sandbox more time to build successfully.


adzuci (approved):
LGTM, it would be ideal to know what takes so long and speed it up, but I'm guessing that's hard to do.

PR #1581, opened 2022-06-30
Add an dbt job to run the enterprise reporting models daily without t…

macdiesel:
…ests.


adzuci (commented):
Just curious, any reason this type of job couldn't be in Prefect?

PR #1586, opened 2022-07-06
chore: Fix string match for expiring GPG keys alert.

HassanJaveed84:


HassanJaveed84:
Sorry had missed your comment, I had tested this change before, we would always get `Keys expiring` counter in case there is a key expiration, so this should work.

PR #1591, opened 2022-08-17
feat: reworked tableau backups to actually create backup 

HassanJaveed84:


dataeng/resources/tableau-backup.sh
```diff
@@ -1,30 +1,39 @@
...
+# tsm maintenance backup utility. This script creates the data and config backup and copies them into s3
+# bucket and then remove backup files from tableau server.
+
+ssh $USER_NAME@$TABLEAU_SERVER_IP /bin/bash << EOF
+set -ex
+sudo su - tsm_admin
+set -ex
+
+# Data backup filename in the format of table_data_backup_2022_01_16_T23_48Z.tsbak
+data_backup_filename="tableau_data_backup_$(date -u +%Y_%m_%d_T%H_%MZ).tsbak"
```
pwnage101, line 14: Is it possible to stick with an ISO-compatible datetime format? https://en.wikipedia.org/wiki/ISO_8601

ISO 8601 allows the removal of non-alphanumeric characters, so this could be
```suggestion
data_backup_filename="tableau_data_backup_$(TZ=UTC; date -u +%Y%m%dT%H%MZ).tsbak"
```

which produces something that looks like this:

```
tableau_data_backup_20220817T1647Z.tsbak
```

You can actually confirm that this is valid parseable ISO format here: https://dencode.com/en/date/iso8601?v=20220817T1647Z&tz=UTC

dataeng/resources/tableau-backup.sh
```diff
@@ -1,30 +1,39 @@
...
+
+echo "Copying data backup file into s3"
+aws s3 cp /home/tableau_backups/\$data_backup_filename s3://$S3_BUCKET/\$data_backup_filename
+
+echo "Removing data backup file from local disk"
+rm /home/tableau_backups/\$data_backup_filename
 
-ssh -tt $USER_NAME@$TABLEAU_SERVER_IP -o StrictHostKeyChecking=no "S3_BUCKET=$S3_BUCKET" '
+# Config backup filename in the format of tableau_config_backup_2022_01_16_T23_48Z.json
+config_backup_filename="tableau_config_backup_$(date -u +%Y_%m_%d_T%H_%MZ).json"
```
pwnage101, line 26: ditto

dataeng/resources/tableau-backup.sh
```diff
@@ -1,30 +1,39 @@
...
-else
-    echo "Copying backup file into s3"
-    sudo aws s3 cp /home/tableau_backups/$BACKUPNAME s3://$S3_BUCKET/$BACKUPNAME
+echo "Copying config backup file into s3"
+aws s3 cp \$config_backup_filename s3://$S3_BUCKET/\$config_backup_filename
 
-    echo "Removing backup file from local disk"
-    sudo rm -f /home/tableau_backups/$BACKUPNAME
+echo "Removing config backup file from local disk"
+rm \$config_backup_filename
```
pwnage101, line 35: Can you change this to delete _all_ backups, in case any were left over from previous runs?  If they really are 40+ GB, then we should try to prevent accidental accumulation on disk.

dataeng/resources/tableau-backup.sh
```diff
@@ -1,30 +1,39 @@
...
+data_backup_filename="tableau_data_backup_$(date -u +%Y_%m_%d_T%H_%MZ).tsbak"
+
+echo "Creating data backup"
+tsm maintenance backup --file \$data_backup_filename --multithreaded
+
+echo "Copying data backup file into s3"
+aws s3 cp /home/tableau_backups/\$data_backup_filename s3://$S3_BUCKET/\$data_backup_filename
+
+echo "Removing data backup file from local disk"
+rm /home/tableau_backups/\$data_backup_filename
```
pwnage101, line 23: ditto

pwnage101 (approved):
LGTM!

PR #1601, opened 2022-08-31
feat: Added job to restore Tableau from backup.

HassanJaveed84:


dataeng/resources/tableau-restore.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+if [ -z $BACKUP_TIMESTAMP ]; then
+	aws s3 ls $S3_PATH | sort -r
+else
+    ssh $USER_NAME@$TABLEAU_SERVER_IP /bin/bash << EOF
```
AbdouSeck, line 6: Calling `ssh` like this can fail because of host key verification. In other words, if the jenkins machine does not have that IP listed in its `~/.ssh/known_hosts` file, this line will fail. Changing that line to `ssh -o StrictHostKeyChecking=no $USER_NAME@$TABLEAU_SERVER_IP /bin/bash` can make it more resilient to the IP not existing in `~/.ssh/known_hosts`.

AbdouSeck (changes requested):

dataeng/resources/tableau-restore.sh
AbdouSeck: Also why `TABLEAU_SERVER_IP`, and not `TABLEAU_SERVER_HOST`? It would make more sense to use route53 records, especially when ssh'ing.

HassanJaveed84: The rationale for IP instead of HOST is that the instance on which this backup job would run may or may not have a route53 record/hostname yet. I am okay to rename this to TABLEAU_SERVER_HOST if it still makes sense.

AbdouSeck: Oh I see. It just thought that it was only intended for IP addresses. I would advocate for a name with HOST since that could mean both IP and hostname.

AbdouSeck (approved):

PR #1603, opened 2022-08-31
fix: bind credentials correctly

Hamza442:
Currently in **Bindings** of warehouse-transform jobs **OPSGENIE_HEARTBEAT_CONFIG_KEY** is missing. For example: https://jenkins-new.analytics.edx.org/view/dbt/job/warehouse-transforms-google_analytics_sessions/configure in the mentioned job **Bindings** only **ANALYTICS_VAULT_ROLE_ID** and **ANALYTICS_VAULT_SECRET_ID** are present but **OPSGENIE_HEARTBEAT_CONFIG_KEY** is missing. Due to this reason when https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/resources/opsgenie-enable-heartbeat.sh gets executed the key is missing and due to missing key it doesn't add heartbeat to the job. 
Tested the change and now job is adding heart beat : https://jenkins-new.analytics.edx.org/view/dbt/job/warehouse-transforms-google_analytics_sessions/12936/console

**JIRA TICKET**: https://2u-internal.atlassian.net/browse/DENG-1318?atlOrigin=eyJpIjoiOTcxMTBiNTVjMjY3NDFhMmEyYTk2MTdjZjFkZTBjYTUiLCJwIjoiaiJ9

### DIFF
--- a/dataeng/jobs/analytics/WarehouseTransforms.groovy
+++ b/dataeng/jobs/analytics/WarehouseTransforms.groovy
@@ -68,11 +68,6 @@ class WarehouseTransforms{
                 wrappers {
                     credentialsBinding {
                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
-                    }
-                }
-                wrappers {
-                    timestamps()
-                    credentialsBinding {
                         usernamePassword('ANALYTICS_VAULT_ROLE_ID', 'ANALYTICS_VAULT_SECRET_ID', 'analytics-vault');
                     }
                 }


pwnage101 (approved):

dataeng/jobs/analytics/WarehouseTransforms.groovy
```diff
@@ -65,14 +65,12 @@ class WarehouseTransforms{
...
+                // Writing two different wrappers for credentialBinding won't write two different credentials.
+                // the last wrappers will overwrite will overwrite the previous ones
+                // To overcome this issue write all credentials in a single wrapper
                 wrappers {
                     credentialsBinding {
                         string('OPSGENIE_HEARTBEAT_CONFIG_KEY', 'opsgenie_heartbeat_config_key')
-                    }
-                }
-                wrappers {
-                    timestamps()
```
pwnage101, line 74: whoops, just noticed this line was removed.  Please add it back.

pwnage101 (changes requested):

dataeng/jobs/analytics/WarehouseTransforms.groovy
HassanJaveed84: I thought the same but then realized that timestamps() is already added as part of `common_wrappers`.

Hamza442: @pwnage101 @HassanJaveed84 So its fine if we don't add it back right ?

HassanJaveed84: Correct.

PR #1606, opened 2022-09-22
feat: add check for edx-exams service

alangsto:
Add optional check to add edx-exams to a sandbox. Checking this will not have any effect on the sandbox build right now, but eventually that value of this param will be used to actually add exams.


Zacharis278 (approved):
✅

jdmulloy (approved):

adzuci (approved):

PR #1613, opened 2022-09-29
chore: Disable Docker CI for old releases

jdmulloy:
Running docker builds for these very old openedx releases is overloading our tools jenkins server.


adzuci (approved):
LGTM

PR #1621, opened 2022-10-27
feat: add fluentd var

alangsto:
If fluentd is enabled, a container will be spun up to listen to port 5140, which is where we anticipate future logs from a containerized edx-platform to be sent to.


schenedx (approved):
👍

PR #1623, opened 2022-10-31
chore: use python3 for edx-analytics-configuration

Hamza442:


dataeng/resources/update-users.sh
```diff
@@ -0,0 +1,9 @@
+VENV_ROOT=$WORKSPACE/venvs
+mkdir -p $VENV_ROOT
+
+if [ ! -d "$VENV_ROOT/analytics-configuration" ]
+then
+    virtualenv --python=python3.8 $VENV_ROOT/analytics-configuration
+fi
+. $VENV_ROOT/analytics-configuration/bin/activate
+make -C analytics-configuration users.update
```
HassanJaveed84, line 9: nit: add new line.

dataeng/resources/update-users.sh
```diff
@@ -0,0 +1,9 @@
+VENV_ROOT=$WORKSPACE/venvs
+mkdir -p $VENV_ROOT
+
+if [ ! -d "$VENV_ROOT/analytics-configuration" ]
+then
+    virtualenv --python=python3.8 $VENV_ROOT/analytics-configuration
+fi
```
HassanJaveed84, line 7: Do we need this if here ?

Hamza442: Yes actually we don't need it here updated the PR

HassanJaveed84 (approved):

PR #1624, opened 2022-11-02
chore: use python3 for analytics-exporter

Hamza442:


dataeng/resources/email-optin-worker.sh
```diff
@@ -13,3 +13,6 @@ ${EXPORTER_VENV}/bin/exporter \
    --django-admin=${PLATFORM_VENV}/bin/django-admin.py \
    --django-pythonpath=${PLATFORM_VENV}/edx-platform \
    --gpg-keys=${GPG_KEYS_PATH}
+
+# remove virtualenv created by shiningpanda plugin to avoid virtualenv issue
+rm -rf ${EXPORTER_VENV}
```
HassanJaveed84, line 18: Is this not going to cause issues when the email optin or exporter job is executed for multiple orgs ?

Hamza442: Tested it with https://jenkins.analytics.edx.org/job/exporter-email-optin-master-copy-temp/4/console
Initially I thought the same but after testing it doesn't seems to cause any issue

mzakaria1 (approved):

PR #1627, opened 2022-11-18
Adds info to README-Hacking instructions

mattdrayer:
Adds a bit of helpful information to the README-Hacking instructions.


matthugs (approved):
it may make sense to note that this is `tools-edx-jenkins.edx.org`-specific advice, but I'm sure it's not the only point in this repo/doc which is

mattdrayer:
Yeah, I am assuming that the reader understands this because they already reviewed the first section of the overall README for this repo, which states _"This repository conatins Jenkins Job DSL code for generating Jenkins jobs on several edx.org Jenkins servers including tools-edx-jenkins and build.testeng.edx.org. It has a companion private repo for generating the Jobs on the edx.org Jenkins servers. Usage of this code without that repo may be difficult."_

PR #1641, opened 2023-01-20
fix: process both orgs yml files together

iamsobanjaved:
Instead of compiling dashboard twice and then appending, we should process yml files from both organizations at same time and compile dashboard together.

### DIFF
--- a/devops/resources/run-repo-health-on-org.sh
+++ b/devops/resources/run-repo-health-on-org.sh
@@ -142,15 +142,12 @@ done < "$input"
 # Go into data repo, recalculate aggregate data, and push a PR
 IFS=,
 failed_repo_names=`echo "${failed_repos[*]}"`
-# Delete existing dashboards to re compile
-find "${WORKSPACE}/repo-health-data/dashboards" -type f -iname "\dashboard*.csv" -delete
 
-for ORG_NAME in ${ORG_NAMES[@]}; do
-    echo "Pushing data for org $ORG_NAME"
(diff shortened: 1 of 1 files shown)


devops/resources/run-repo-health-on-org.sh
```diff
@@ -142,15 +142,12 @@ done < "$input"
...
-for ORG_NAME in ${ORG_NAMES[@]}; do
-    echo "Pushing data for org $ORG_NAME"
-    cd "${WORKSPACE}/repo-health-data/individual_repo_data/${ORG_NAME}"
-    ls
-    repo_health_dashboard --data-dir . --configuration "${WORKSPACE}/edx-repo-health/repo_health_dashboard/configuration.yaml" --output-csv "${WORKSPACE}/repo-health-data/dashboards/dashboard" --append
-done
+# Compiling dashboard with latest yml files from both orgs
+echo "Pushing data"
+cd "${WORKSPACE}/repo-health-data/individual_repo_data"
+ls
```
aht007, line 149: Are we using the ORG_NAMES variable anywhere else? If not we should probably delete that. Also do we still need ls command here? Previously it was printing out yaml files now it will only print out the directories that exists within `individual_repo_data` directory.

aht007 (approved):
Added some suggestions but they aren't something that would block this PR and hence approving.

devops/resources/run-repo-health-on-org.sh
iamsobanjaved: Yes, `ORG_NAMES` has other uses as well apart from this. And I have updated the `ls` command to output the names of files within each subdirectory.

PR #1645, opened 2023-01-25
chore: Remove shinningpanda Jenkins plugin

mzakaria1:


dataeng/resources/remote-config.sh
```diff
@@ -14,6 +14,11 @@ set -xe
 
 env | sort
 
+# Creating python3.7 virtual env
+PYTHON_VENV="python_venv"
+virtualenv --python=python3.7 --clear "${PYTHON_VENV}"
```
Hamza442, line 19: I would suggest to update it to python3.8 and also test the job with you branch and see if anything breaks. It looks good to me but just make sure you have tested it.

dnuzz:
Do you also need to remove the plugin from the `build.gradel` file?

HassanJaveed84 (approved):
Looks good, I would test out the exporter changes manually before merging this, just to make sure they work.

mzakaria1:
> Do you also need to remove the plugin from the `build.gradel` file?

We'll do that later sometime, but not in this PR

PR #1672, opened 2023-10-09
feat: trigger prefect flow after models successfully transferred to s3

muhammad-ammar:


dataeng/resources/trigger-prefect-flow.sh
```diff
@@ -0,0 +1,37 @@
...
+    ${PREFECT_VAULT_KV_PATH} \
+)
+
+# Get Authenticated with Prefect Cloud
+prefect auth login --key $PREFECT_CLOUD_AGENT_TOKEN
+
+set -x
+
+# Trigger prefect flow
+prefect run --id $FLOW_ID
```
muhammad-ammar, line 37: **TODO:** Need to decide where to store the value flow id?

HassanJaveed84 (approved):
LGTM. 
Were we able to test this ?

PR #1697, opened 2023-11-02
fix: add prospectus janitor job

ohnickmoy:
Add prospectus janitor job per [matts comments](https://github.com/edx/jenkins-job-dsl-internal/pull/813#pullrequestreview-1707187796)


adzuci (approved):

matthugs (approved):
We _could_ also avoid polluting the janitor.sh file with a bit more duplication, but there's not a lot of reason to I don't think

PR #1704, opened 2023-11-27
chore: call aws secretsmanager

katebygrace:
JIRA:CLOUDSEC-12

### DIFF
--- a/dataeng/resources/model-transfers.sh
+++ b/dataeng/resources/model-transfers.sh
@@ -28,7 +28,9 @@ fi
 
 ARGS="{mart: ${MART_NAME} }"
 
-dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
+source secrets-manager.sh analytics-secure/warehouse-transforms/profiles DBT_PASSWORD
+
+dbt deps --profiles-dir $WORKSPACE/dataeng/profiles --profile $DBT_PROFILE --target $DBT_TARGET
 
 # Call DBT to perform all transfers for this mart.
 dbt run-operation perform_s3_transfers --args "${ARGS}" --profile $DBT_PROFILE --target $DBT_TARGET --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/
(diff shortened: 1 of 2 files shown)


dataeng/resources/model-transfers.sh
```diff
@@ -28,7 +28,11 @@ fi
 
 ARGS="{mart: ${MART_NAME} }"
 
-dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
+DBT_TARGET="prod"
```
HassanJaveed84, line 31: Do we need to hardcode this to prod? 
DBT_TARGET is a jenkins job level parameter.
https://github.com/edx/jenkins-job-dsl/blob/a2d7d190f1576b6f8a6e76953957cc0f8fc8909a/dataeng/jobs/analytics/ModelTransfers.groovy#L22

katebygrace: nah, though its prod everywhere in analytics secure but i can leave it

dataeng/resources/model-transfers.sh
```diff
@@ -28,7 +28,11 @@ fi
...
+
+source secrets-manager.sh analytics-secure/warehouse-transforms/profiles DBT_PASSWORD
+
+dbt deps --profiles-dir . --profile $DBT_PROFILE --target $DBT_TARGET
```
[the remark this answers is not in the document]
katebygrace: oh interesting - should i add the files below there?

dataeng/resources/model-transfers.sh
```diff
@@ -28,7 +28,9 @@ fi
 
 ARGS="{mart: ${MART_NAME} }"
 
-dbt deps --profiles-dir $WORKSPACE/analytics-secure/warehouse-transforms/ --profile $DBT_PROFILE --target $DBT_TARGET
+source secrets-manager.sh analytics-secure/warehouse-transforms/profiles DBT_PASSWORD
+
+dbt deps --profiles-dir $WORKSPACE/dataeng/profiles --profile $DBT_PROFILE --target $DBT_TARGET
```
HassanJaveed84, line 33: So the workspace is a jenkins provided environment variable that points to the jenkins job workspace. 
jenkins-job-dsl repo. does not get cloned unless we specify it in the DSL.

PR #1709, opened 2023-12-04
feat: allow configurable ecommerce version in sandbox

christopappas:


devops/jobs/CreateSandbox.groovy
```diff
@@ -145,7 +145,7 @@ class CreateSandbox {
                 stringParam("forum_version","master","")
 
                 booleanParam("ecommerce",true,"")
-                stringParam("ecommerce_version","master","")
+                stringParam("ecommerce_version",ecommerceVersion,"")
```
pshiu, line 148: [optional] ⛳ Some golfing fun: 

```suggestion
                stringParam("ecommerce_version",extraVars.get("ECOMMERCE_VERSION", "master"),"")
```

pshiu (approved):

devops/jobs/CreateSandbox.groovy
christopappas: look at my first commit actually haha. i dispreferred the inline statement and thought it would be more readable to pull the `.get` outside of the params var

christopappas: (which follows the pattern of access_control above for example)

PR #1717, opened 2024-01-03
chore: remove sysadmin part2

katebygrace:
JIRA:CLOUDSEC-12


katebygrace:
re: rationale; sysadmin has credentials in it, so when we clone that onto jenkins we're back in plaintext password land; which we're moving off of, that said will give this a look over

matthugs:
re: sysadmin having credentials in it, great that makes sense & is what I expected you to say.

PR #1725, opened 2024-02-27
chore: migrate user retirement job from tubular to edx-platform

ohnickmoy:
First pass for https://github.com/openedx/axim-engineering/issues/881. Testing to move one user-retirement job from tubular repo to edx-platform repo


dianakhuang (approved):
seems reasonable enough. let's see how well it works.

PR #1728, opened 2024-02-28
chore: migrate analytics exporter

katebygrace:
there wasnt any sensitive info in these files so i think they can be safely copied, cloned, and called. hopefully got it correct

### DIFF
--- a/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
+++ b/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
@@ -133,6 +133,9 @@ class AnalyticsEmailOptin {
                 // This will create python 3.8 venv inside shell script instead of using shiningpanda
                 shell(dslFactory.readFileFromWorkspace('dataeng/resources/setup-platform-venv-py3.sh'))
                 shell(dslFactory.readFileFromWorkspace('dataeng/resources/setup-exporter-email-optin.sh'))
+                shell(dslFactory.readFileFromWorkspace('dataeng/analytics-exporter/email_optin.yaml'))
+                shell(dslFactory.readFileFromWorkspace('dataeng/analytics-exporter/default.yaml'))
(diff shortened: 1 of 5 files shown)


dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
```diff
@@ -133,6 +133,9 @@ class AnalyticsEmailOptin {
                 // This will create python 3.8 venv inside shell script instead of using shiningpanda
                 shell(dslFactory.readFileFromWorkspace('dataeng/resources/setup-platform-venv-py3.sh'))
                 shell(dslFactory.readFileFromWorkspace('dataeng/resources/setup-exporter-email-optin.sh'))
+                shell(dslFactory.readFileFromWorkspace('dataeng/analytics-exporter/email_optin.yaml'))
```
HassanJaveed84, line 136: I don't think this will work as I believe `shell` is for running a shell command only.

katebygrace: Okay, I was just trying to pull in the file, but I think it's already in there, am I correct?

dataeng/resources/setup-exporter-email-optin.sh
```diff
@@ -16,8 +16,11 @@ pip install -r github_requirements.txt
 pip install mysql-connector-python -e .
 popd
 
+aws secretsmanager get-secret-value --secret-id analytics-secure/analytics-exporter/task-auth.json --region "us-east-1" --output json | jq -r ".SecretString" >>  {WORKSPACE}/dataeng/analytics-exporter/task-auth.json
+
+
 # Configuration paths in analytics-secure
-SECURE_ROOT=${WORKSPACE}/analytics-secure/analytics-exporter
+SECURE_ROOT=${WORKSPACE}/dataeng/analytics-exporter
```
HassanJaveed84, line 23: I believe this would not work as this repo(jenkins-job-dsl), does not get cloned as part of this job.

PR #1756, opened 2024-04-11
chore: fix script + test

katebygrace:
This PR 
- Fixes the secrets-manager script so you can either call it with -w or --whole_file to get the whole secret file (for private keys) 
- Adds it into the refresh snowpipe job as a place to test, calling the actual private keys/passphrase from secrets maanager and not just providing the path.

### DIFF
--- a/dataeng/resources/snowflake-refresh-snowpipe.sh
+++ b/dataeng/resources/snowflake-refresh-snowpipe.sh
@@ -14,19 +14,20 @@ make requirements
 source $WORKSPACE/secrets-manager.sh
 # Fetch the secrets from AWS
 set +x
-get_secret_value analytics-secure/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS KEY_PATH
-get_secret_value analytics-secure/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS PASSPHRASE_PATH
-get_secret_value analytics-secure/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS USER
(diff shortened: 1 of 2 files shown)


dataeng/resources/snowflake-refresh-snowpipe.sh
```diff
@@ -14,19 +14,20 @@ make requirements
...
+secrets-manager.sh -w analytics-secure/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS snowflake/rsa_key_snowpipe_user.p8
+secrets-manager.sh -w analytics-secure/job-configs/SNOWFLAKE_REFRESH_SNOWPIPE_JOB_EXTRA_VARS snowflake/rsa_key_passphrase_snowpipe_user
+
 set -x
 
 python refresh_snowpipe.py \
-    --key_path $WORKSPACE/analytics-secure/$KEY_PATH \
-    --passphrase_path $WORKSPACE/analytics-secure/$PASSPHRASE_PATH \
-    --user $USER \
+    --user 'SNOWPIPE' \
```
HassanJaveed84, line 25: Do we need to do this ?

dataeng/resources/snowflake-refresh-snowpipe.sh
```diff
@@ -14,19 +14,20 @@ make requirements
...
     --schema $SCHEMA \
-    --account $ACCOUNT \
+    --account 'edx.us-east-1' \
```
HassanJaveed84, line 27: Same as the user, I think we can leave that as is.

dataeng/resources/snowflake-refresh-snowpipe.sh
katebygrace: nah, but we arent going to call the extra_vars unless we really need to, so i replaced all of the vars

PR #1767, opened 2024-05-20
chore: analytics exporter

katebygrace:
JIRA:CLOUDSEC-12


dataeng/resources/setup-exporter.sh
```diff
@@ -16,8 +16,12 @@ pip install -r github_requirements.txt
 pip install mysql-connector-python -e .
 popd
 
-# Configuration paths in analytics-secure
-SECURE_ROOT=${WORKSPACE}/analytics-secure/analytics-exporter
+cd analytics-config/analytics-exporter
+python3 secrets-manager.py -w -n analytics-secure/analytics-exporter/task-auth.json -v task-auth.json
```
HassanJaveed84, line 20: Looking at https://github.com/edx/jenkins-job-dsl/blob/6cc70f5c8dd1da7e182cc23767a4941014f8b52e/dataeng/jobs/analytics/AnalyticsExporter.groovy#L148, it looks like we also need this in addition to task-auth.json

HassanJaveed84: I noticed that the path here is `analytics-secure/analytics-exporter/task-auth.json`, shouldn't it be `analytics-config` ?

katebygrace: Yes to analytics-tools repo. It should be in there already. 

Task-auth has sensitive values, so i was going to pull it from secrets manager

PR #1781, opened 2024-09-05
feat!: remove demographics Snowflake cleanup job

justinhynes:
This job is no longer required as the Demographics IDA has been decommissioned and its data has been deleted.


dataeng/jobs/createJobs.groovy
```diff
@@ -9,7 +9,6 @@ import static analytics.JenkinsBackup.job as JenkinsBackupJob
 import static analytics.PipelineAcceptanceTestManual.job as PipelineAcceptanceTestManualJob
 import static analytics.PipelineAcceptanceTestMaster.job as PipelineAcceptanceTestMasterJob
```
deborahgu, line 10: what happened here? is this tabs to spaces or something?

justinhynes: No idea! That's how it's always looked for as long as I can remember: https://github.com/edx/jenkins-job-dsl/blob/master/dataeng/jobs/createJobs.groovy.

deborahgu (approved):

PR #1784, opened 2024-09-19
chore: use python 3.11 for schema builder

HassanJaveed84:
Switch to Python 3.11 for schema builder running on analytics Jenkins(OCM).


davidlwolf (approved):
please add a PR description - otherwise go for it

PR #1833, opened 2026-03-10
fix: deleting snowflake cleanup script

ktyagiapphelix2u:
**Summary**
Removing this job because it's no longer needed and has been failing since June 2024.

**Why Remove It?**
The job's purpose was to delete soft-deleted user retirement records from Snowflake that contained unredacted PII.

Now: We redact PII (redacted-XXX format) before it reaches Snowflake, so there's no unredacted PII to clean up. The cleanup job is obsolete.

**What Changed**
Deleted job definition and shell script
Removed from job creation pipeline
Deleted configuration file

**Impact**
Removes a failing job
No functional impact - PII is already redacted at the source
Soft-deleted records remain in Snowflake but contain only redacted values

**Ticket Link -**  https://2u-internal.atlassian.net/browse/BOMS-293

**Related PR**
https://github.com/edx/analytics-config/pull/48


robrap (approved):
Thanks @ktyagiapphelix2u. I'll take care of merging when I am ready, but this is all set.

PR #1844, opened 2026-05-29
chore: update PLATFORM_BRANCH parameter in AnalyticsEmailOptin job to point to the release-ulmo branch

rpipaliya-sonata:
This pull request updates the default branch used for the `PLATFORM_BRANCH` parameter in the `AnalyticsEmailOptin` job configuration. The new default is now `origin/release-ulmo` instead of a specific commit hash.

**JIRA**: https://2u-internal.atlassian.net/browse/COSMO2-911

### DIFF
--- a/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
+++ b/dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
@@ -75,7 +75,7 @@ class AnalyticsEmailOptin {
                 stringParam('ORGS','*', 'Space separated list of organizations to process. Can use wildcards. e.g.: idbx HarvardX')
                 stringParam('EXPORTER_BRANCH','origin/master',
                         'Branch from the edx-analytics-exporter repository. For tags use tags/[tag-name]. Should be environment/production.')
-                stringParam('PLATFORM_BRANCH', 'origin/release-ulmo', 'Branch from the edx-platform repository. For tags use tags/[tag-name]')
+                stringParam('PLATFORM_BRANCH', 'release-ulmo', 'Branch from the edx-platform repository. For tags use tags/[tag-name]')
                 stringParam('EXPORTER_CONFIG_FILENAME','email_optin.yaml', 'Name of configuration file in analytics-secure/analytics-exporter.')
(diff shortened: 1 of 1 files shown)


dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
```diff
@@ -75,7 +75,7 @@ class AnalyticsEmailOptin {
                 stringParam('ORGS','*', 'Space separated list of organizations to process. Can use wildcards. e.g.: idbx HarvardX')
                 stringParam('EXPORTER_BRANCH','origin/master',
                         'Branch from the edx-analytics-exporter repository. For tags use tags/[tag-name]. Should be environment/production.')
-                stringParam('PLATFORM_BRANCH', '83a18cdfc806709e08ed5b2cd02d25f6d247529b', 'Branch from the edx-platform repository. For tags use tags/[tag-name]')
+                stringParam('PLATFORM_BRANCH', 'origin/release-ulmo', 'Branch from the edx-platform repository. For tags use tags/[tag-name]')
```
timmc-edx: [suggestion] This might work as is (I see that `EXPORTER_BRANCH` already defaults to using a remote qualifier) but it assumes that we know what the remote will be called, so it's probably a bit safer to just use the branch name.

```suggestion
                stringParam('PLATFORM_BRANCH', 'release-ulmo', 'Branch from the edx-platform repository. For tags use tags/[tag-name]')
```

timmc-edx (approved):
Approved, although with one suggestion.

timmc-edx:
We'll also need some process for ensuring we later switch to `release-verawood`. I've made a note in https://2u-internal.atlassian.net/wiki/spaces/ENG/pages/2273771547/Named+Release+Upgrades+scratch+page to consider how to do this.

dataeng/jobs/analytics/AnalyticsEmailOptin.groovy
rpipaliya-sonata: Yes, We can just add branch name without remote qualifier.
Updated the code.

PR #1852, opened 2026-08-04
fix: Install node on user retirement health check script

rahulkanneri-2u:
### Description
Install Node.js locally during the user retirement health check job so the Playwright test setup can run npm commands even when the Jenkins host does not already have Node/npm on PATH.

### Changes
Add a workspace-local Node.js install using nodeenv.
Default the health check job to Node.js 24.
Activate the generated Node environment before running npm ci, Playwright browser install, and the retirement health test.
Print node and npm versions in the Jenkins log for easier debugging.

### Jira Ticket
[https://2u-internal.atlassian.net/browse/BOMS-528](https://2u-internal.atlassian.net/browse/BOMS-528)


dataeng/resources/user-retirement-health-check.sh
ttak-apphelix: can we also remove node and venv

rahulkanneri-2u: @ttak-apphelix this is done.

ttak-apphelix (approved):

